Issue #123, opened 2016-03-17
Support STIX 1.2 data format


polygloton:
# Support STIX 1.2 data format in our API

Use the java-stix lib and don't worry about swagger UI compatibility.

Here is my high-level TODO list:
- [x] Export a single Indicator
- [ ] Import a single Indicator
- [ ] Export judgements for an indicator
- [ ] Test importing real-world TAXI data

craigbro:
I think it should return the original representation of the indicator, without any judgements at `/cia/indicators/ID`  See my coments on 142 regarding how we should represent imported XML indicators as a direct JSON repreentation of XML.

I think we should have the export of an indicator in a form that is just the judgements rendered into a simple observable pattern at: `/cia/indiator/ID/judgements  and use the same filter map logic

polygloton:
See also #146.

Issue #142, opened 2016-03-26
Capture observable qualifier data


polygloton:
There are some fields on Cybox observable (which is polymorphic) that we don't currently capture.  Should we capture any of these?  Address is probably the most interesting.
## Address (ip, ipv6, device)
- [ ] is destination?
- [ ] is source?
- [ ] is spoofed?

See http://cybox.mitre.org/language/version2.1/xsddocs/objects/Address_Object.html
## Domain
- [ ] Type (FQDN, TLD)

See http://cybox.mitre.org/language/version2.1/xsddocs/objects/Domain_Name_Object.html
## User
- [ ] password required?
- [ ] a bunch of account detail fields the we probably don't care about

See http://cybox.mitre.org/language/version2.1/xsddocs/objects/User_Account_Object.html
## File (sha256, sha1, md5)
- [ ] A bunch of fields for files that we probably don't care about

See http://cybox.mitre.org/language/version2.1/xsddocs/objects/File_Object.html

craigbro:
Perhaps we can add a "qualifier" field to the observable object.  it does not playa  role as a key, but after a verdict hit, it can be used to check if it applies to the observable based on where you got it from?

PR #146, opened 2016-03-29
First pass at exporting an Indicator in STIX 1.2 XML format

polygloton:
There is more work to be done, but I think this is ready to be considered for merge.

Depends on or closes #145.

This work is for issue #123, though that issue is not done.


craigbro:
Check out my comment on #142 where I discuss how we should represent XML imported Indicators.

src/ctia/http/routes/stix12.clj
```diff
@@ -0,0 +1,23 @@
+(ns ctia.http.routes.stix12
+  (:require [compojure.api.sweet
+             :refer [defapi context GET]]
+            [ctia.http.handlers.stix12.indicator :as indicator]
+            [ring.util.http-response
+             :refer [ok not-found]]
+            [schema.core :as s]))
+
+(defapi stix12-routes
```
polygloton: Seems to work ok.  I'll change it if it becomes a problem.

saintx:
This looks like a good start, as long as we make a record of any remaining issues that we uncovered along the way  :+1:

PR #151, opened 2016-03-31
Fixing bugs in properties code

### LINKED ISSUE #116: Does the properties implementation break passed-in properties?
The properties file code reads in a properties file and stores the properties in an atom as a map.  Then we check the map for configuration in the code.  I think this means that properties specified on the command line won't work.  Need to investigate this and probably fix it.

polygloton:
Properties that are read from a properties file may be overridden with system properties of the same name (note that only overrides are kept; properties not defined in the file that are defined as system properties are ignored).  Also fixed a bug in transform (using rmerge instead of assoc-in).

Closes #116

Another improvement is that this version closes the properties file correctly after reading it.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -2,14 +2,17 @@
   :description "FIXME: write description"
   :dependencies [[org.clojure/clojure "1.7.0"]
                  [clj-time "0.9.0"] ; required due to bug in lein-ring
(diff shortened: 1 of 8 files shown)


polygloton:
Replaced with #157

Issue #140, opened 2016-03-24
Remove Joda Time from Schema


saintx:
Joda Time is deprecated and users are recommended to switch to the java.time (JSR-310) utility in Java SE 8+  However, we don't want to limit our app to only running in Java 8 environments, so we're going to move toward using java.util.Date.

It also makes our schemas not portable to cljs, and it breaks serialization in yml and other formats.  It also prevents me from using enduro for file-based atom durability.

Edit: I originally suggested we use https://docs.oracle.com/javase/8/docs/api/java/time/OffsetTime.html
More info about JSR-310 date and time improvements:
http://www.oracle.com/technetwork/articles/java/jf14-date-time-2125367.html  The team discussed this option and decided against it.

Edit: Use java.util.Date instead of java.time or joda DateTime.  For the schema use schema.core/Inst instead of defining a specific type since schema will use the desired type depending on environment.

polygloton:
Remaining to-dos:

 - [ ] Refactor to avoid joda as intermediary in coercion
 - [ ] Update the datastructures.md documentation (it references joda)

Issue #72, opened 2016-02-24
Implement durable atoms in memory store


polygloton:
https://github.com/alandipert/enduro
https://github.com/polygloton/durable-atom

polygloton:
Need to remember to test the atom store in both modes: in-memory and on-disk

PR #155, opened 2016-04-01
Using java.util.Date instead of JodaTime

polygloton:
New namespace ctia.lib.time, which is currently a facade in front of cli-time.  We can refactor the time namespace in a different PR.  This is for #140, but I don't think it is ready to be closed.  We can get this PR merged to make forward progress on #72.


polygloton:
I parked with @saintx on this today.

gbuisson:
Very nice, sounds good to me!

PR #157, opened 2016-04-01
Multiple improvements to properties

### LINKED ISSUE #116: Does the properties implementation break passed-in properties?

polygloton:
Fixes some bugs, improves the code, and adds schemas.

Closes #116.


src/ctia/init.clj
```diff
@@ -19,14 +19,14 @@
             [ctia.stores.memory.ttp :as mt]))
 
 (defn init-auth-service! []
-  (let [auth-service-name (get-in @properties/properties [:auth :service])]
-    (case auth-service-name
-      "allow-all" (reset! auth/auth-service (allow-all/->AuthService))
-      "threatgrid" (reset! auth/auth-service (threatgrid/make-auth-service
+  (let [auth-service-type (get-in @properties/properties [:auth :service :type])]
```
saintx: smurfy

src/ctia/properties.clj
```diff
@@ -1,51 +1,83 @@
...
 
-(def property-files ["ctia.properties"
-                     "ctia-default.properties"])
+(def property-files
+  "Property file names, in order of preference"
+  ["ctia.properties"
+   "ctia-default.properties"])
 
-(defonce properties (atom {}))
+(defonce properties
```
saintx: I'd rename this to "data", probably.

src/ctia/properties.clj
```diff
@@ -1,51 +1,83 @@
...
+            [clojure.string :as str]
+            [ctia.lib.map :as map]
+            [schema.coerce :as c]
+            [schema.core :as s]
+            [clj-time.coerce :as coerce])
+  (:import java.util.Properties))
 
-(def property-files ["ctia.properties"
-                     "ctia-default.properties"])
+(def property-files
```
saintx: I'd probably rename this to just "files"

saintx:
Other than minor naming issues, this looks pretty clean.

project.clj
```diff
@@ -2,14 +2,17 @@
   :description "FIXME: write description"
   :dependencies [[org.clojure/clojure "1.7.0"]
                  [clj-time "0.9.0"] ; required due to bug in lein-ring
-                 [metosin/schema-tools "0.7.0"]
+                 [prismatic/schema "1.1.0"]
```
gbuisson: we may run into issues if we use a version of schema which is not supported by compojure-api

gbuisson:
I like the coercion part, we just need to keep in mind that the schema dep version change may alter stability.

src/ctia/properties.clj
polygloton: I wasn't thrilled with properties/properties either, but I feel like "data" is too generic.  I think its okay in this case because almost every ns only cares about this atom (except for ctia.init which has to initialize this).  The other ns'es can just do a :refer [properties] and deref @properties as desired.  I _think_ that is clean.

polygloton:
I addressed the feedback.  Tried to remove surfy names.  Also I remembered that I don't need the explicit schema dependency so I just removed it.

PR #160, opened 2016-04-01
Remove clj-time coercion.

saintx:
This removes the coercion to joda time and then back to date time
or sql-time, in favor of direct coercion mechanisms.  All new Date
objects use UTC implicitly.  This resolves issue #140


polygloton:
Looks good.

Issue #161, opened 2016-04-01
Race condition in events_test.clj


saintx:
I intermittently get the following test failure from master:

lein test :only ctia.events-test/test-recents

FAIL in (test-recents) (events_test.clj:76)
expected: (= ["1" "2" "3"] (map (comp :id :model) (recent-events)))
  actual: (not (= ["1" "2" "3"] ()))

I think there's a race condition on that line.

polygloton:
Should be fixed, but since it is an intermittent failure, we should wait a few days and get confirmation that we aren't seeing this anymore.

polygloton:
Seems to be resolved.

PR #162, opened 2016-04-01
fixes issue-#161

quoll:


saintx:
Works for me :+1:

polygloton:
This seems correct.  Since "recent events" is a lose contract, there doesn't need to be any flushing and there isn't anything else to wait on that I can think of, so sleeping makes sense.

PR #170, opened 2016-04-04
Fix the ES issue on travis

### LINKED ISSUE #168: CI tests fail for the ES store
Getting error:

org.elasticsearch.client.transport.NoNodeAvailableException: None of the configured nodes are available: []

yogsototh:
Apparently the test suite doesn't uses `test/resources/ctia-test.properties` at least for the `es-store` tests.

This is a simple hot fix for the travis check to success. But we should take a look on how we could force the test suite to use the test properties file.

Closes #168

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ jdk:
 #     only:
 #         - master
 before_script:
-    - cp test/resources/ctia-test.properties.ci test/resources/ctia-test.properties
+    - cp test/resources/ctia-test.properties.ci resources/ctia-default.properties
     # Wait ES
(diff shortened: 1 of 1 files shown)


polygloton:
My preference is to try to get it working before resorting to this fix.  The handler tests have a fixture that reloads the properties.  It seems like the ES store code is remembering old property values.  We need the test fixtures to initialize (and configure) the ES store.  If that is going to be a major pain, then this fix would be okay.

This seems relate to the issue that @saintx noticed, where it seems like we try to connect to ES on a require statement (eg before we initialize the ES store).

Issue #159, opened 2016-04-01
Refactoring handlers to accomodate hooks, and events and stores


craigbro:
A discussion with the crew.

For CRUD operations, we would like the following "flow" within handlers.
 
## Creation and Update

  1. Realize the object from the inputs, using the existing "realize" functions.
    1. The realization function produced our initial, default StoredObject
    2. Call our "Realization" hooks, which may return a modified StoredObject
        - these are the before-create/update/delete hooks
  2. Wrap the StoredObject in the appropriate Event.
  3. Call the Event hooks, which cannot transform, and must all run succesfully (kafka feed...)
       - these are the after-create/update/delete-hooks, except those functions dont take events, they take the object itself.  This is a deviation
  4. Call the stores, passing in the StoredObject, to perform the operation, it must return succesfully
  5. Pass the Event (not just the StoredObject) on the core.async channel

## Deletion
  1. Replicate the realize- flow with a destroy- flow
  2. Product the Event, and it should contain the model, which is an optional field in the schema
  3. Call the Event hooks, which cannot transform, and must all run succesfully (kafka feed...)
  4. Call the stores delete/destroy, it must be successful
  5. Pass the Event on the core.async channel

## Realization Hooks
For every Stored Object type, we have a vector of Hook objects, modeled on GUNDAM hook.  Hooks are all "required" in that they must all return succesfully.  They must be run in a specific order, this allows the specification of hook order as a means of control flow configuration. Hooks can return a modified realized object.  If they return nil the current realized object is used.  They take the same parameter set as the realize functions.

## Commentary
Craig: Note that we isolate business logic in the realize and detroy steps, from the implementation logic in the Event hooks.

Yann: The hooks seem similiar to ring middlware, but middleware is defined on a per handler basis, and is dealing with non-realized input parameters (the request map).

yogsototh:
I believe I can split this task in two sub tasks:

- [ ] Refactor the code to call a list of function before modifying the store.
- [ ] Provide the ability to add multiple external jar, and to declare an order of execution.

polygloton:
## More TODO's copied out of #195

- [ ] Implement the `find-hooks` and call `init-hooks` during init
- [ ] Change the hooks data structure to provide documentation

And maybe:

- [ ] Make a demo with a java jar in resources + properties used for ordering or maybe a leon-module?

Also, I merged the PR195 in order to get the changes in ASAP, but there were code review comments.  Need to address them (or ignore them) as appropriate.

[...]

PR #173, opened 2016-04-06
Initial event log ES producer

### LINKED ISSUE #150: Event log ES Producer
a producer that write entity events to an ES instance

gbuisson:
Closes #150 

- New Producer protocol
- ES Producer Record
- Added new tests for each producer
- also fixes ES connection spawned on non integration tests

to be used after issue #159 is completed.


polygloton:
Looks good.

PR #187, opened 2016-04-08
Configure HTTP & nREPL servers via properties

### LINKED ISSUE #186: HTTP Port should be configured
It should be a property

polygloton:
Removes the lein-ring plugin and explicitly starts the jetty server.  We lose some automatic features, but gain control.  (Please let me know if you are missing a feature from lein-ring and I can add it).

Server is started with "lein run".

Closes #186 

I'm adding this so that we can use the configured port in URL IDs.

### DIFF
--- a/resources/ctia-default.properties
+++ b/resources/ctia-default.properties
@@ -1,4 +1,9 @@
 auth.service.type=allow-all
+ctia.http.port=3000
+ctia.http.min-threads=10
+ctia.http.max-threads=100
+ctia.nrepl.port=3001
+ctia.nrepl.enabled=true
 ctia.store.default=memory
 ctia.store.sql.db.classname=org.h2.Driver
 ctia.store.sql.db.subprotocol=h2
--- a/test/ctia/test_helpers/core.clj
+++ b/test/ctia/test_helpers/core.clj
@@ -2,6 +2,7 @@
   (:refer-clojure :exclude [get])
   (:require [ctia.auth :as auth]
             [ctia.auth.allow-all :as aa]
+            [ctia.http.server :as http-server]
             [ctia.properties :as props]
             [ctia.store :as store]
             [ctia.stores.atom.store :as as]
@@ -11,7 +12,6 @@
             [clojure.data :as cd]
             [clojure.edn :as edn]
             [clojure.test :as ct]
-            [ring.adapter.jetty :as jetty]
             [schema.core :as schema]))
 
 (defmethod ct/assert-expr 'deep= [msg form]
@@ -79,19 +79,10 @@
 
 (def fixture-atom-store (fixture-store atom-stores))
 
-(def http-port 3000)
-
-(defn fixture-server [app & {:keys [port]
-                             :or {port http-port}}]
-  (fn [f]
-    (let [server (jetty/run-jetty app
-                                  {:host "localhost"
-                                   :port port
-                                   :join? false
-                                   :max-threads 10
-                                   :min-threads 9})]
-      (f)
-      (.stop server))))
+(defn fixture-http-server [f]
+  (http-server/start! :join? false)
+  (f)
+  (http-server/stop!))
 
 (defn set-capabilities! [login role caps]
   (store/create-identity @store/identity-store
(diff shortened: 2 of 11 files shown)


yogsototh:
We might want to add `ring.middleware.wrap-reload` since if I'm not wrong, the lein-ring plugin did it for us.

May be we could use [`environ`](https://github.com/weavejester/environ) to put the wrap-reload middleware only for `dev` profile.

polygloton:
It looks like lein-ring has its own reload code.  I'll see if wrap-reload sufficiently replaces it.  We can probably just turn it on/off with a parameter.  Or we could have an environment parameter.

resources/ctia-default.properties
```diff
@@ -1,4 +1,9 @@
 auth.service.type=allow-all
+ctia.http.port=3000
+ctia.http.min-threads=10
+ctia.http.max-threads=100
+ctia.nrepl.port=3001
+ctia.nrepl.enabled=true
 ctia.store.default=memory
```
saintx: I like how configurable this makes the nrepl.

test/ctia/test_helpers/core.clj
```diff
@@ -2,6 +2,7 @@
   (:refer-clojure :exclude [get])
   (:require [ctia.auth :as auth]
             [ctia.auth.allow-all :as aa]
+            [ctia.http.server :as http-server]
             [ctia.properties :as props]
```
saintx: I'd make this line [ctia.properties :refer [properties]]

test/ctia/test_helpers/core.clj
```diff
@@ -101,7 +92,7 @@
 
 (defn url
   ([path]
-   (url path http-port))
+   (url path (get-in @props/properties [:ctia :http :port])))
```
saintx: ...and I'd switch this to ```(url path (get-in @properties [:ctia :http :port])))```

saintx:
This change would break the current functionality of the Dockerfile, which calls ```lein ring uberjar```, so we'd need to update that as well.

yogsototh:
I tried to add `ctia.http.dev-reload=true` in `resources/ctia-default.properties` and it doesn't appear to work. I tried to add some `println`. But I don't see why it doesn't work. The code seems correct to me.

A simple remark.
Why didn't you put the `wrap-reload` middleware in the `defapi`? I believe you could write

~~~clojure
(defn wrap-if [handler pred wrapper] (if pred (apply wrapper handler) handler))

(defapi ...
  (middleware [
      (wrap-if (dev-reload?) wrap-reload)
     ]
    {:swagger ... }
    (GET ...)
~~~

I don't believe this is better now. But, in the future if we want to add more than one middleware, we might prefer to have them defined in `defapi`.

test/ctia/test_helpers/core.clj
polygloton: I think it has to call init!, which is why it is there.  You can see me doing machinations to not have to write properties/properties.  :-P

polygloton:
@yogsototh I tested it and this is what I found.  If you make a change to the handler, it doesn't reload the handler (I assume that is because some reference to a handler instance is held), but if you change something else (I changed a realize fn in my test), then it seems to work.  I don't know if this was any different with lein-ring because I do mostly REPL or test driven dev.

polygloton:
@saintx I will fix the Docker file.

polygloton:
@yogsototh I agree that middleware should go into the API, usually.  In this case, there is runtime logic to decide if we want to apply the middleware (we don't want it running in prod).  The hander is currently a def, which is too late to check a property.  Also, the server already has all of the properties in its scope, so I just put it there.  You do make a good point though.  I could change the handler to be a defn instead of a def, and it could take the bool parameter for the middleware.  Please let me know if you think that is better and I'll change it.

yogsototh:
No you're right. The `wrap-reload` is a special case and the code is very clear that way. I don't believe we should make the handle a defn for now. So for me everything looks good.

gbuisson:
looks good

saintx:
I was able to build the docker container and launch it locally.  Looks like it's binding to ports correctly, and the app resolves.  Ship it!

PR #189, opened 2016-04-08
Refactoring to enable hooks

### LINKED ISSUE #159: Refactoring handlers to accomodate hooks, and events and stores

yogsototh:
This is a first preview on how I could refactor the code to add hooks.

I must modify all create function which before took a `NewX` and returned a `StoredX` via `realize-x`.
There is an unused macro, that would prevent a bit of copy/paste.

I prefer to ask if I'm starting correctly. Obviously this PR isn't ready to be merged.

Close #159

### DIFF
--- a/src/ctia/flows/crud.clj
+++ b/src/ctia/flows/crud.clj
@@ -1,0 +1,33 @@
+(ns ctia.flows.crud
+  "This namespace handle all necessary flows for creating, updating and deleting objects.
+
+  (Cf. #159)."
+  (:import java.util.UUID)
+  (:require [ctia.flows.hooks :refer :all]))
+
+
+(defn camel-case [txt]
+  (->> (clojure.string/split txt #"-")
+      (map clojure.string/capitalize)
+      clojure.string/join))
+
+(defn make-id [type-name _]
+  (str type-name "-" (UUID/randomUUID)))
+
+;; NOT USED MACRO BUT COULD BE USEFUL IF WE START TO BECOME LAZY
+;; OR IF THE NUMBER OF HANDLER START TO BECOME TOO BIG
+(defmacro realize
+  "Launch the create flow
+
+  1. realize the object provided a type, a login and a raw-object
+     For this the function realize-X is called where X the type type-name
+  2. Apply all hooks to the realized objects
+  3. return the realized object modified by all hooks
+
+  Saving the realized object should occurs after this call.
+  "
+  [hook-name type-name login object]
+  `(let [id# (make-id ~(name type-name) ~object)
+         realized# (~(symbol (str "realize-" (name type-name))) ~object id# ~login)
+         hooked# (apply-hooks ~type-name realized# ~hook-name)]
+     hooked#))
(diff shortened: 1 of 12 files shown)


src/ctia/flows/crud.clj
```diff
@@ -0,0 +1,33 @@
...
+
+  1. realize the object provided a type, a login and a raw-object
+     For this the function realize-X is called where X the type type-name
+  2. Apply all hooks to the realized objects
+  3. return the realized object modified by all hooks
+
+  Saving the realized object should occurs after this call.
+  "
+  [hook-name type-name login object]
+  `(let [id# (make-id ~(name type-name) ~object)
```
craigbro: I think we should only be setting ID when one is not provided.

src/ctia/flows/crud.clj
```diff
@@ -0,0 +1,33 @@
...
+
+  (Cf. #159)."
+  (:import java.util.UUID)
+  (:require [ctia.flows.hooks :refer :all]))
+
+
+(defn camel-case [txt]
+  (->> (clojure.string/split txt #"-")
+      (map clojure.string/capitalize)
+      clojure.string/join))
```
polygloton: Should move this to ctia.lib.string

src/ctia/stores/atom/actor.clj
```diff
@@ -1,11 +1,19 @@
...
   (:require [ctia.schemas.actor :refer [NewActor StoredActor realize-actor]]
-            [ctia.stores.atom.common :as mc]))
+            [ctia.stores.atom.common :as mc]
+            [schema.core :as s]))
 
 (def swap-actor (mc/make-swap-fn realize-actor))
 
-(mc/def-create-handler handle-create-actor
-  StoredActor NewActor swap-actor (mc/random-id "actor"))
+(mc/def-create-handler-from-realized handle-create-actor StoredActor)
```
craigbro: Let's convert this to a fn generator, instead of a macro, call validate, see if we can find a way to make the validate conditional on test mode.  If not, validate at all times on input, better safe than sorry, and this function will have other entry points that realize -- consider this is also the entry point for import/replication etc...

src/ctia/schemas/common.clj
```diff
@@ -180,3 +183,35 @@
     :else (http-response/bad-request!
            {:error "Mismatching :dispostion and dispositon_name for judgement"
             :judgement judgement})))
+
+(defn storedSchema
```
polygloton: Rename to stored-schema

src/ctia/flows/hooks.clj
```diff
@@ -0,0 +1,83 @@
+(ns ctia.flows.hooks
+  "Handle hooks (Cf. #159).")
+
+(def default-hooks
+  {:before-create []
+   :before-send-to-event-chan []})
```
craigbro: Let's rename this to after-create ?

Also, we need documentation, preferably embedded, but also in the header.

yogsototh:
Closing the PR for now to remove hipchat pollution. It is still work in progress.

Issue #24, opened 2016-02-15
A core.async listener which writes domain events to a log file.


craigbro:
A listener which takes events from our core.async channel for domain events, and logs them to a log file.

quoll:
My first inclination was to log this data at INFO level. Is that right? Do we make this configurable instead?

quoll:
I think I'm ready to PR this, but tests pulled from central are failing. It looks like I might just need to have some services running to make these work?

quoll:
This is done, but there is a new patch that builds on issue-23, and I'm waiting until that is complete before finishing this

PR #190, opened 2016-04-08
fixes issue #24

quoll:


src/ctia/logging.clj
```diff
@@ -0,0 +1,15 @@
...
+            [ctia.events :as e]
+            [schema.core :as s])
+  (:import [clojure.core.async.impl.protocols Channel]))
+
+(s/defn log-channel :- Channel
+  "Logging an event channel indefinitely. Returns the channel for a go loop that never ends."
+  [{m :mult :as ev} :- e/EventChannel]
+  (let [events (chan)]
+    (tap m events)
+    (go (loop [ev (<! events)]
```
craigbro: Please ensure that you shutdown gracefully.

http://stackoverflow.com/questions/20485188/gracefully-exit-a-clojure-core-async-go-loop-on-kill

Try running this and then killing your process.  if it hangs for a bit, remedy as above.  If it dies immediately, you are golden.

saintx:
#24 A core.async listener which writes domain events to a log file (for context)

src/ctia/logging.clj
polygloton: I recommend this pattern.  It terminates when the channel closes and doesn't have the double <!.
```clojure
(go-loop []
  (when-let [ev (<! events]
    ;; do things
    (recur)))

File: test/ctia/logger_test.clj [hunk 2]
```diff
@@ -0,0 +1,36 @@
+(ns ctia.logger-test
+  (:require [ctia.logging :refer :all]
+            [ctia.events :as c :refer :all]
+            [clojure.test :as t :refer :all]
```
polygloton: In general, try to avoid :refer :all style requirements.  _Maybe_ it is okay for clojure.test, but I would argue against doing a :refer :all for the system under test.  The (init!) function below is a common function name, so I can't tell where it is coming from.

test/ctia/logger_test.clj
```diff
@@ -0,0 +1,36 @@
...
+  
+(deftest test-logged
+  (init!)
+  (let [{b :chan-buf c :chan m :mult :as ev} @central-channel
+        sb (StringBuilder.)
+        ll log/log*
+        patched-log (fn [logger level throwable message]
+                      (.append sb message)
+                      (.append sb "\n"))]
+    (alter-var-root (var log/log*) (constantly patched-log))
```
polygloton: The alter-var-root is a smell to me.  If you can't use an atom, or can you use (with-redefs ...)?

quoll: It is indeed a code smell. I nearly wrote a macro to duplicate the functionality of with-redefs, since I thought it would be handy to have such a thing. Thank you for this!

The correct way to do it would be to register a LogWriter that picks up the logs and presents them back. Unfortunately, SLF4J wrapped and hid this functionality (which I learnt from the sources on Apache after spending some time thinking I could use a writer like I just said). Monkey patching the log* is a hack to get around this.

src/ctia/logging.clj
quoll: Had not seen non-terminating loops shut down like this in the past. Thanks.

Also like the when-let inside the loop. The double <! was bothering me.

polygloton:
Looks good

PR #195, opened 2016-04-12
Refactoring to handle hooks

yogsototh:
So this is a very big PR.

- As Craig asked, I replaced most macros in the atom store by functions.
- centralized the hooks workflows in `flow/crud.clj`
- removed the `login` parameter in the Store Protocols login is used in `realize-X` functions.
- You can find a demo for using hooks in `test/ctia/flows/hooks-test.clj`

I still see some other things to do:

- [ ] Implement the `find-hooks` and call `init-hooks` during init
- [ ] Change the hooks data structure to provide documentation
- [x] Refactor the `realize-X` functions (I only did it for `actor`)
- [x] Provide previous object to hooks for updates

And maybe:

- [ ] Make a demo with a java jar in resources + properties used for ordering or maybe a lein-module?

WIP on #159


resources/ctia-default.properties
```diff
@@ -4,7 +4,7 @@ ctia.store.sql.db.classname=org.h2.Driver
 ctia.store.sql.db.subprotocol=h2
 ctia.store.sql.db.subname=/tmp/ctia-h2-db;DATABASE_TO_UPPER=false
 ctia.store.sql.db.delimiters=
-ctia.store.es.uri=http://localhost:9200
+ctia.store.es.uri=http://192.168.99.100:9200
 ctia.store.es.indexname=ctia
-ctia.producer.es.uri=http://127.0.0.1:9200
+ctia.producer.es.uri=http://192.168.99.100:9200
```
polygloton: Is the IP address change intentional?  Maybe you should set the IP in your local ctia.properties.  If it is intentional, I recommend that you make this change in a separate PR.

craigbro: Stephen Sloan <notifications@github.com> writes:

> In resources/ctia-default.properties:
>
>>  ctia.store.es.indexname=ctia
>> -ctia.producer.es.uri=http://127.0.0.1:9200
>> +ctia.producer.es.uri=http://192.168.99.100:9200

Yah, that should not have been hanged in the PR.  The default shoudl
stay localhost.  update your own ctia.properties file, which is
gitignore'ed

>
> Is the IP address change intentional? Maybe you should set the IP in
> your local ctia.properties. If it is intentional, I recommend that you
> make this change in a separate PR.
>
> —
> You are receiving this because you were assigned.
> Reply to this email directly or view it on GitHub
>
>

-- 
Craig Brozefsky
Principal Engineer, AMP Threat Grid
Cisco Security Business Group
+1-773-469-8349

yogsototh: Good catch! I certainly made a `git add .`

src/ctia/stores/atom/common.clj
```diff
@@ -7,59 +7,50 @@
   (fn [_new-entity_]
     (str prefix "-" (UUID/randomUUID))))
 
-(defmacro def-read-handler [name Model]
-  `(s/defn ~name :- (s/maybe ~Model)
-     [state# :- (s/atom {s/Str ~Model})
-      id# :- s/Str]
-     (get (deref state#) id#)))
+(defn read-handler [Model]
+  (s/fn :- (s/maybe Model)
```
polygloton: The implementation of s/fn causes a performance hit.  It would be better to pull it out into a s/defn..

yogsototh: You're right, but I don't see how I could use a `defn` without going back to using macros. How would you do this?

AlexBaranosky: You could also not use schema/fn, and just use clojure.core/fn if concerned about the performance hit in this one spot. Though how much does performance actually matter here?

polygloton: Good points.  I don't want to lose the schema protection; it is very helpful when testing.  Probably s/fn is acceptable.

polygloton:
I like the refactoring done here!  Looking good.

src/ctia/flows/hooks.clj
```diff
@@ -0,0 +1,89 @@
...
+  (init [this])
+  (handle [this
+           type-name
+           stored-object
+           prev-object])
+  (destroy [this]))
+
+(defn find-hooks
+  "Should return a list of `Hook` in the correct order"
+  [hook-name]
```
polygloton: I think "hook-name" here (and below) should be "hook-type".

src/ctia/flows/crud.clj
```diff
@@ -0,0 +1,75 @@
...
+(defn create-flow
+  "This function centralize the create workflow.
+  It is helpful to easily add new hooks name
+
+  To be noted:
+    - `:before-create` hooks can modify the object stored.
+    - `:after-create` hooks can't modify the object returned but
+      a hook modifying an object will provide the modified object
+      to the next `:after-create` hook and so on."
+  [realize-fn
```
polygloton: Please consider using maps and destructuring for these fns.  Its hard to keep track of the args when reading the handler code.

PR #196, opened 2016-04-12
ES Producer Index Slicing

### LINKED ISSUE #176: Allow configuration of time sliced indexes for ES event producer
I shoul be able to tell the event producer to use time-sliced (by day, week, month, year) indexes for storing Events in ES.

gbuisson:
closes #176 
Initial support for time sliced indexes on ES producer
2 different strategies are supported

- `aliased-index` creates an index by time range, for each Index created, the same alias.
- `filtered-alias`creates a single Index with a date range filtered alias for each time range.

granularity can be configured and set to either minute hour day week month year.


src/ctia/properties.clj
```diff
@@ -50,7 +50,11 @@
    (s/optional-key "ctia.producer.es.host") s/Str
    (s/optional-key "ctia.producer.es.port") s/Int
    (s/optional-key "ctia.producer.es.clustername") s/Str
-   (s/optional-key "ctia.producer.es.indexname") s/Str})
+   (s/optional-key "ctia.producer.es.indexname") s/Str
+   (s/optional-key "ctia.producer.es.slicing.strategy")
+   (s/enum :filtered-alias :aliased-index)
+   (s/optional-key "ctia.producer.es.slicing.granularity")
+   (s/enum :minute :hour :day :week :month :year)})
```
polygloton: Having keys and values on separate lines like this is hard to parse unless you add spaces in between.

gbuisson:
This is a good case for optionalize I think

src/ctia/lib/es/index.clj
```diff
@@ -1,6 +1,8 @@
 (ns ctia.lib.es.index
   (:require
+   [clojure.core.memoize :as memo]
    [schema.core :as s]
+   [clj-time.core :as t]
```
polygloton: We should use ctia.lib.time as a facade to hide the time library implementation.

src/ctia/lib/es/slice.clj
```diff
@@ -0,0 +1,104 @@
...
+   granularity :- s/Keyword]
+
+  (let [parsed (coerce/from-date event-time)
+        year (t/year parsed)
+        month (t/month parsed)
+        day (t/day parsed)
+        hour (t/hour parsed)
+        minute (t/minute parsed)]
+
+    (-> (condp = granularity
```
polygloton: Can you use case here instead of (condp =)?  Its faster and idiomatic.

src/ctia/lib/es/index.clj
```diff
@@ -76,3 +92,63 @@
...
+   name
+   routing
+   filter))
+
+(def cached-create-filtered-alias!
+  (memo/ttl create-filtered-alias!
+            :ttl/threshold
+            alias-create-cache-ttl-ms))
+
+(def cached-create-aliased-index!
```
polygloton: I suggest calling this memo-create-aliased-index!.  Its a subtle nit though...

src/ctia/lib/es/index.clj
```diff
@@ -9,12 +11,19 @@
    [ctia.events.producers.es.mapping :refer [producer-mappings]]
```
polygloton: I think ideally a lib namespace should not use other CTIA namespaces like this.  A lib should be low-level and should not depend on higher level things.  Maybe this isn't a lib, or maybe parts of it are not.  Maybe this lib should just take parameters instead of being coupled.

polygloton:
I like how configurable this is.  I also like how you used schemas in places that helped when reading the code.

src/ctia/lib/es/index.clj
polygloton: I think you should use a FIFO memo instead of a TTL memo.  Dates will change as time progresses, so older entries should fall out.  I think the way that TTL memo is implemented, it only purges entries when they are hit.  Since we won't try to index in old slices (once a new time interval starts), the TTL cache could grow in size over time.  FIFO on the other hand fits our pattern and won't grow beyond a specified size.

resources/ctia-default.properties
```diff
@@ -12,6 +12,8 @@ ctia.store.sql.db.subname=/tmp/ctia-h2-db;DATABASE_TO_UPPER=false
 ctia.store.sql.db.delimiters=
 ctia.store.es.uri=http://localhost:9200
 ctia.store.es.indexname=ctia
+ctia.producer.es.uri=http://localhost:9200
 ctia.producer.type=es
-ctia.producer.es.uri=http://127.0.0.1:9200
 ctia.producer.es.indexname=ctia_events
+ctia.producer.es.slicing.strategy=aliased-index
```
polygloton: I think the name of this property is confusing.  The user cares about if the CTIA instance is using time slicing, and not how its implemented.

resources/ctia-default.properties
```diff
@@ -12,6 +12,8 @@ ctia.store.sql.db.subname=/tmp/ctia-h2-db;DATABASE_TO_UPPER=false
 ctia.store.sql.db.delimiters=
 ctia.store.es.uri=http://localhost:9200
 ctia.store.es.indexname=ctia
+ctia.producer.es.uri=http://localhost:9200
 ctia.producer.type=es
```
polygloton: Should this be on by default?

polygloton:
I think we should get #236 resolved along with this one.

PR #202, opened 2016-04-12
Standardized the properties used by ctia.init

### LINKED ISSUE #200: Store properties don't match
In the default properties file it is ctia.store.default and in init.clj its looking for [:store :service :default].  The default is actually provided in-code as :memory.

Alas, I think it should be ctia.store.service, with the default value defined in the property file only.

polygloton:
Cleaning up some inconsistencies

Closes #200

### DIFF
--- a/resources/ctia-default.properties
+++ b/resources/ctia-default.properties
@@ -1,10 +1,10 @@
-auth.service.type=allow-all
+ctia.auth.type=allow-all
 ctia.http.port=3000
 ctia.http.min-threads=10
 ctia.http.max-threads=100
 ctia.nrepl.port=3001
 ctia.nrepl.enabled=true
-ctia.store.default=memory
+ctia.store.type=memory
 ctia.store.sql.db.classname=org.h2.Driver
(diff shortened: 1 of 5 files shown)


saintx:
:+1:

resources/ctia-default.properties
```diff
@@ -1,10 +1,10 @@
-auth.service.type=allow-all
+ctia.auth.type=allow-all
 ctia.http.port=3000
 ctia.http.min-threads=10
 ctia.http.max-threads=100
 ctia.nrepl.port=3001
 ctia.nrepl.enabled=true
-ctia.store.default=memory
```
gbuisson: I think the purpose of having `default` here was to have the ability to define a store per entity, later on you would write something like `ctia.store.actor=memory` and so on.

polygloton: @gbuisson I like your approach.  I would like to get this change in, because compared to what we have now, it is an improvement, but I would like to make a new PR that implements your idea.

gbuisson: I fully agree

PR #203, opened 2016-04-13
replaced s/eq by s/enum to fix swagger UI bug

### LINKED ISSUE #199: Schema Error in Swagger UI posting default New Actor
When I use the "click to set as parameter value" feature for the Actor model, it sets an example Actor parameter in the "New Actor" field.  but when I click "Try it out", I get the following error:

    {
      "errors": {
        "type": "(not (= \"actor\" \"string\"))"
      }
    }

yogsototh:
Closes #199

### DIFF
--- a/src/ctia/schemas/actor.clj
+++ b/src/ctia/schemas/actor.clj
@@ -28,7 +28,7 @@
     }))
 
 (s/defschema Type
-  (s/eq "actor"))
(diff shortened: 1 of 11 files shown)


polygloton:
@saintx Can you confirm that this PR fixes the problem?

saintx:
Confirmed, this seems to fix the issue.

PR #206, opened 2016-04-13
Can set ENV variables to override properties

### LINKED ISSUE #192: Allow environment variables to set properties
Merged with property files, just like how system properties work.  Might need to replace "." with "_" and down-case variable names.

polygloton:
Also added some doc strings

Closes #192


gbuisson:
sounds good !

yogsototh:
looks good!

quoll:
Looks good. I like the mapping of the keys to their environment varname equivalents

PR #207, opened 2016-04-13
Use the CTIA entry point in tests

### LINKED ISSUE #185: Test ctia.init/init! path
It doesn't currently get tested, so we don't find bugs in that code.

### LINKED ISSUE #201: Correct CI setup
We are doing this in the .travis.yml file:

```yml
 before_script:
    - cp test/resources/ctia-test.properties.ci resources/ctia-default.properties
```

This was added because the test fixtures where we set properties was not causing ES to be reconfigured, even though there is a fixtures for the ES store.  ES store fixtures should initialize ES for each test using the currently set properties.  The tests should not depend on the actual properties file at all.  Need to correct (or disprove) the problem and remove the above line from the .travis.yml file.

polygloton:
- Better test coverage
- De-complect the auth service and the fake whoami service
- Deletes duplicate property files (leaving just one)

Closes #185 and closes #201.
Depends on or closes #206.


polygloton:
I had to get increase the PermGen space for CI to get tests to pass for Oracle JDK 7.  This suggests a reference is being held between test fixtures (memory leak).  Isn't an issue for OpenJDK7 or Oracle JDK 8 (doesn't use PermGen).  I don't know where the leak is, but Jetty is the likely culprit.  I think it is fine for now (the issue is specific to testing and CI is working), but we need to keep it in mind.

gbuisson:
looks good to me, however I wonder if we could skip the PermGen increase and specific lein profile

polygloton:
I wasn't able to get CI to work without the permgen increase.

yogsototh:
looks good to me. I believe having a specific lein profile for ci is acceptable.

PR #208, opened 2016-04-13
Hooks with java and jar file examples

### LINKED ISSUE #159: Refactoring handlers to accomodate hooks, and events and stores

yogsototh:
This PR provide two examples on how to use Java and/or Jar file to be used as hooks.

The method is not plug and play yet, but this is a good first step toward coding the `find-hooks` function.

Closes #159


yogsototh:
Actually, the autoload only uses jar that provide classes which implement `ctia.Hook` java interface.

In my code, I transform such class into a Record extending the `Hook` Protocol.
The goal was to prevent javaist to use clojure data structure and let them use `java.util.HashMap<String,Object>`.

So if we want to add new hooks coded in Clojure in external jar, we could use a less friendly `reify` from the `ctia.Hook` java interface and I could get rid completely of the `Hook` Protocol.
But I believe we should privilege  the Clojure Protocol.

I didn't provided any example, yet, but I'm currently trying to provide jar generated from Clojure.
I will have to modify my `autoload-hooks` function slightly.

gbuisson:
impressive! 👍

AlexBaranosky:
You guys aren't big PR commit squashers?

When you want to rollback a pull request you may hate life having to rollback 12 commits.

Also from the perspective of the master branch, it can be nice to see each piece of logical functionality within its own commit.

yogsototh:
Yes you're right, I'll squash most commits. Do you believe I should squash the complete PR into one commit?

AlexBaranosky:
@yogsototh I (quickly) read through the commits and it seems like there is really only one "chunk of meaningful functionality" or "feature" being added here. For that reason I'd squash it all to one commit.

If there were two meaningful separate chunks of functionality, that might want to be reverted independently, then I'd do two commits, etc.

polygloton:
I think you can remove the nested lein project files (like CHANGELOG, LICENSE, doc/, etc.).  If you don't strictly plan on using it, just delete it.  README might be useful for the example though.  The top level license, etc. will apply to all code in the repo.

yogsototh:
I've removed all file which I believe aren't used. Most java file are given as example on how to generate the jar files. May be we could but these examples in another repository, but I also use these `jar` files to make my tests. So it might be a good idea to keep them in this repository.

polygloton:
Looks good

PR #229, opened 2016-04-15
Implements pubsub for redis

### LINKED ISSUE #23: A core.async listener which forwards domain events to redis pub/sub
A listener which takes events from our core.async channel for domain events, and logs them to a redis pubsub channel.

quoll:
Rather than using multiple function callbacks being called from redis,
this subscribes a single function, and then uses a core.async multi channel
to distribute published events to all registered subscribers.

Closes #23


yogsototh:
On my localhost I stopped the docker-machine to be sure Redis on the docker was closed.
I launched a local Redis and then I did:
~~~
~/d/c/ctia ❯❯❯ export CTIA_STORE_REDIS_URI="redis://127.0.0.1:6379"
~/d/c/ctia ❯❯❯ lein test :only ctia.publish-test
~~~

Which failed with an Operation Timeout.

Apparently the environment variable doesn't really affect the redis connection settings.

test/ctia/test_helpers/redis.clj
```diff
@@ -0,0 +1,8 @@
+(ns ctia.test-helpers.redis
+  (:require [ctia.test-helpers.core :as h]))
+
+(defn fixture-properties:redis-store [f]
+  ;; May be overridden with ENV variables
+  (h/with-properties ["ctia.store.redis.enabled" true
+                      "ctia.store.redis.uri" "redis://192.168.99.100:6379"]
```
craigbro: This is likely the cause of your problem, your are hardcoding the value in your text fixture.

PR #232, opened 2016-04-18
Use of optionalize

### LINKED ISSUE #231: Use optionalize to enhance schema readability

yogsototh:
I believe that using `st/optional-keys` make the code more readable when there are a lot of `s/optional-key` in a schema. What do you think about it?

Closes #231

### DIFF
--- a/src/ctia/schemas/coa.clj
+++ b/src/ctia/schemas/coa.clj
@@ -10,56 +10,28 @@
 (s/defschema COA
   (merge
    c/GenericStixIdentifiers
-   {:valid_time c/ValidTime
-    (s/optional-key :stage)
-
-    (describe v/COAStage
-              "specifies what stage in the cyber threat management lifecycle this Course Of Action is relevant to ")
-
-    (s/optional-key :coa_type)
-    (describe v/COAType
-              "type of this CourseOfAction")
-
-    (s/optional-key :objective)
-    (describe [s/Str]
-              "characterizes the objective of this Course Of Action") ;; Squashed / simplified
-
-    (s/optional-key :impact)
-    (describe s/Str
-              "characterizes the estimated impact of applying this Course Of Action")
-
-    (s/optional-key :cost)
-    (describe v/HighMedLow
-              "characterizes the estimated cost for applying this Course Of Action")
-
-    (s/optional-key :efficacy)
-    (describe v/HighMedLow
-              "effectiveness of this Course Of Action in achieving its targeted Objective")
-
-    (s/optional-key :source)
-    (describe s/Str
-              "Source of this Course Of Action")
-
-    (s/optional-key :related_COAs)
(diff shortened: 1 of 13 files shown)


polygloton:
I think this is nice, though it may surprise some people when they've never seen it before.  I suggest that everyone on the dev team take a look at this.

gbuisson:
I think it's good as it improves readability of keys with descriptions.

src/ctia/schemas/coa.clj
```diff
@@ -10,56 +10,28 @@
...
+     :objective (describe [s/Str] "characterizes the objective of this Course Of Action") ;; Squashed / simplified
+     :impact (describe s/Str "characterizes the estimated impact of applying this Course Of Action")
+     :cost (describe v/HighMedLow "characterizes the estimated cost for applying this Course Of Action")
+     :efficacy (describe v/HighMedLow "effectiveness of this Course Of Action in achieving its targeted Objective")
+     :source (describe s/Str "Source of this Course Of Action")
+     :related_COAs (describe rel/RelatedCOAs "identifies or characterizes relationships to one or more related courses of action")
+     ;; Not provided: handling
+     ;; Not provided: parameter_observables ;; Technical params using the CybOX language
+     ;; Not provided: structured_COA ;; actionable structured representation for automation
+     })))
```
polygloton: Breaking out the descriptions onto multiple lines was for better readability (manually parsing and finding the value schema).

quoll:
I like this, though I think it should also be extended to ctia.properties/PropertiesSchema

src/ctia/schemas/coa.clj
quoll: I did wonder about that myself, but I also found that this new code was relatively readable

polygloton: It hurts my brain haha

yogsototh: Yes I wasn't sur about that. It depends on where you want to focus.
If you prefer to see access the list of keys, having on line per key seems better.
If you prefer to focus on the content of the description, then using many lines is better.

And for some descriptions we should even use `str` to break down one description on multiple lines.

polygloton: Keys and value schemas are equally important to me.  Schemas look great without the descriptions...

```clojure
{:foo FooSchema}
```

Since we need the descriptions, I added the whitespace as a compromise to improve readability.  Maybe something else could improve on the white space, but it was a decent first attempt.  Readability should be our #2 goal (#1 being "it works").

polygloton:
I like the readability proposal

yogsototh:
@polygloton do you like the readability proposal more than multilines or do you prefer that I revert to a multiline like before?

polygloton:
Both are better than the one-line version.  I think the use of (str ...) is correct, so your version is best.

PR #238, opened 2016-04-19
CRUD Event hook

### LINKED ISSUE #236: CRUD operations create events
This doesn't seem to be happening yet.

yogsototh:
Close #236

Connects to #159

Actually I don't believe I had the necessary functionalities to send events.

I tried to make a first attempt at creating such a hook.
But it is unclear to me, how I should send events?

There is a `produce` function which appear to do what we need.
But I believe there are also many `send-*` functions to send event in a core.async channel.
`produce` appear to be more general as we could use it to send to different logs storage.
Apparently @gbuisson made an interface for ES.

Actually the code isn't finished yet but there shouldn't be more work to finish it.
We should simply add all the events hooks in the init process to make them the default.
And also fix the problem of retrieving the model from an object type.


yogsototh:
This PR push event using `producer` for each CRUD operations.

I would appreciate a review concerning the schemas.
I made some object to event functions. I could provide directly the Model which I believe wasn't done before.

The update function try to match the event schema, but I have a concern about the list of triplets which I believe would be a list of operation to pass from the previous object to the current. My function only allow "deleted", "modified" and "added" operations which concerning hash-map would have the propriety to be equivalent to a set of operation (the order is not important).

The third element of the triplet has type `{s/Str s/Str}` I changed it to `{s/Any s/Any}`.
And I use it to put the different sub object.

Also to complete #159, we should add the event to the core/async channel.
There are two different method to achieve this.

1. Create a new hook
2. Create a new producer (as the hook I added call producer, which itself could manage many different producers, ES and later if I understood correctly Kafka, Redis, core/async, etc...)

Which would be the right choice in your opinion?

src/ctia/events/obj_to_event.clj
```diff
@@ -0,0 +1,61 @@
...
+            [schema.core :as s]
+            [ctia.lib.time :as t]
+            [ctia.schemas.actor :refer [StoredActor]]
+            [ctia.schemas.campaign :refer [StoredCampaign]]
+
+            [ctia.events.schemas :as vs]))
+
+
+(s/defn to-create-event :- vs/CreateEvent
+  [model :-  {s/Any s/Any}
```
polygloton: I think that the '{s/Any s/Any}' schemas are a smell.  It that tells me that you needn't bother with schemas at all in this namespace.

yogsototh: Unfortunately I couldn't find a schema to tell the input is itself a Schema.
As all our schema are maps I used it. But I also could remove the signature to the `model` input.

But I like the ability to tell that `to-create-event` will output a `CreateEvent` instead of anything.
So I'm not sure I'll be at ease to remove all schemas.

quoll: I think this came from me 2 months ago.
It's indicative of a map, but I did not know if the keys were keywords or strings (I haven't seen a model yet, but I'm now presuming that the keys will be keywords). If models have a defined set of keys, then it would obviously be better to use that schema. Otherwise, the key type should be updated

polygloton: @yogsototh I think you make a good point about using schema here because you want to specify the return type; that is fine.

Here is a what I think about when to use schemas.  It has nothing to do with this code.

We should consider _when_ to use schemas.  (I don't mean to suggest we remove all schemas).  We shouldn't autoamtically use schemas where they don't provide value.  In my opinion, schemas provide value when (1) they are defined on the entry point to an abstraction or, (2) when the readability of the code is noticeably improved (communicating assumptions and providing a contract) or, (2) they are added to improve common test failures by failing faster (higher up the stack).

polygloton:
I think we need to move the configuration out of project.clj in this PR.  Imagine that a customer will download an uberjar, and will use "java -jar ..." to run the CTIA.  It will be too late for them to configure project.clj.  For defaults, I suggest we hard-code them.  Code is the "gold standard".  Jar's provided by the customer will have to be added on the command line with "-cp".  How will we know to load the classes?  It will have to specified in a property.  Maybe multiple classes are provided in one property with ":" as a delimiter, or maybe properties support lists somehow (I don't know).

craigbro:
Let's NOT attempt to solve the loading of hooks in this PR.

craigbro:
Get simple hooks working, before adding stuff like read only hooks.

A hook implementor has ALL the tools they need to control wether they run in a thread or otherwise manage their resoures -- let's not complect our code trying to guess what they might want in the future

craigbro:
Implement the core.async gateway as a simple hook.  Why do we want to add yet another layer of abstraction with a "producer" abtractions?  What does it buy us?

yogsototh:
So now, when `ctia.producer.es.*` is configured, all CRUD operations sends events to ES.

There is one detail that bother me, I would like some clarification.
There is a `:model` key in which I provide the complete schema inside each event.
For me it looks like a lot of loss of resources, furthermore it makes the code a bit more difficult to write.
So does I understood the `:model` key correctly?

polygloton:
@yogsototh To answer your question about :model, it should contain the instance of the entity.  A create event, for example, should contain the thing that was created.  It should be possible for someone to replay all of the events and end up at the same state that our store is in.

Perhaps :model was the incorrect name for that field.  Please feel free to change it for better clarity.  When we created the event schemas, it was still abstract.  You can always make it more concrete (or more accurate).

src/ctia/flows/hooks/event_hook.clj
```diff
@@ -0,0 +1,23 @@
+(ns ctia.flows.hooks.event-hook
+  (:require
+   [ctia.flows.hook-protocol :refer [Hook]]
+   [ctia.events.producers.es.producer :as esp]))
+
+(defrecord ESEventProducerRecord [conn]
+  Hook
+  (init [_] (do
+              (reset! conn (esp/init-producer-conn))
+              (comment println "ESEventProducerRecord Initialized: " (pr-str @conn))))
```
polygloton: Probably fine, but the comment here means that nil is returned

yogsototh: Ah yes, I didn't made it really clear. But `init` and `destroy` functions should contains list of function for which the return value as not any importance. These are just function to give a chance to have some stateful effect, typically, asking for resources.

I should write better comment for the `Hook` protocol.

src/ctia/events/producers/es/mapping.clj
```diff
@@ -17,7 +17,7 @@
     :properties
     {:owner string
      :timestamp ts
-     :model {:type "object"}
+     :object {:type "object"}
```
polygloton: My preference would be to call it entity instead of object, because it is more specific.  On the other hand, I don't think it matters much.

craigbro:
:+1:

polygloton:
Looks good

PR #240, opened 2016-04-21
Added Feedback on everything

### LINKED ISSUE #222: Feedback on everything
Support setting feedback on all entities.

Since we all have URL IDs, we should be able to have feedback fields and just an entity reference (URL ID).  (See #212).

yogsototh:
Close #222


yogsototh:
I was completely wrong apparently.

PR #241, opened 2016-04-21
Make Sightings Awesome

### LINKED ISSUE #167: Remove sightings from indicator
@craigbro brought this up in chat

yogsototh:
Close #167


yogsototh:
The tests fail for ES.

Apparently #251 should fix the issue.

src/ctia/stores/atom/common.clj
```diff
@@ -42,15 +42,28 @@
...
     (into []
           (filter (fn [model]
                     (every? (fn [[k v]]
-                              (if (sequential? k)
-                                (= v (get-in model k ::not-found))
-                                (= v (get model k ::not-found))))
+                              (let [found-v (if (sequential? k)
+                                              (get-in model k ::not-found)
+                                              (get model k ::not-found))]
+                                (if (sequential? v)
```
quoll: I suggest using `set?` here, since the API is for "set or not set".
Using `sequential?` would match to a vector (is that an allowable value type?), but the following `contains?` would ensure it would never match.

test/ctia/http/routes/observable_test.clj
```diff
@@ -245,36 +244,37 @@
...
+      (let [{status :status		
+             sightings :parsed-body		
+             :as response}		
+            (get "ctia/ip/10.0.0.1/sightings"		
+                 :headers {"api_key" "45c1f5e3f05d0"})]		
+        (is (= 200 status))		
+        (let [res (->> sightings
+                       (map #(dissoc % :created :modified :indicator))
+                       set)]
+          (is (deep=
```
quoll: This test is currently failing

src/ctia/stores/atom/common.clj
yogsototh: @gbuisson saw this and, in fact at first I've had used `set?` instead of `sequential?` but generally you want it to work for sequential too. So I've just pushed a commit where I transform any sequential in a set. So now, the issue with `contains?` should be solved.

yogsototh: Finally using `(set v)` didn't worked for a reason hard to grasp. I surrender and use `set?` instead.
For which everything works correctly. Thanks @quoll

src/ctia/stores/atom/common.clj
```diff
@@ -42,15 +42,28 @@
...
                     (every? (fn [[k v]]
-                              (if (sequential? k)
-                                (= v (get-in model k ::not-found))
-                                (= v (get model k ::not-found))))
+                              (let [found-v (if (sequential? k)
+                                              (get-in model k ::not-found)
+                                              (get model k ::not-found))]
+                                (if (set? v)
+                                  (contains? v found-v)
+                                  (= v found-v))))
```
polygloton: I like this :)

yogsototh:
To do:

- [ ] get all the sighthings for an indicator (route /indicator/:id/sightings)
- [ ] have a `sightings-by-indicator` function for all the stores
- [ ] get all the sightings for an observable (route /observable/:observable_type/:observable_value/sightings)
- [ ] have a `sighting-by-observable` function for all the stores

yogsototh:
Actually the route `/observable/:observable_type/:observable_value/sightings` is made made like this:

~~~clojure
(some->> {:type observable_type
               :value observable_value}
              (list-judgements-by-observable @judgement-store)
              (list-indicators-by-judgements @indicator-store)
              (list-sightings-by-indicators @sighting-store))
~~~

For atom, it is certainly the best way to do it (I don't believe we should create index ourselves for Memory Storage)

But is it acceptable for other Store?
Certainly not for SQL but for now it is not active for this query.
Further more, as each resource type could use its own store, I don't think we could do any better unless we have some assurance every resources (judgement, indicator and sighting) share the the same store.

I believe this is beyond the scope of this issue. Am I correct?

So for now, I'll simply declare the function but I won't change its inner code.

polygloton:
We assume that, in the future, we will have a split store configuration where some entities are backed by SQL, while others are backed by ES.

craigbro:
(s/defschema Sighting
  "See http://stixproject.github.io/data-model/1.2/indicator/SightingType/"
  (st/merge
   {:id c/ID
    :timestamp c/Time
    :description s/Str
    ;; wether this Sighting is intended to be shared, replicated, copied...
    ;; TLPValue is an enum "red", "yellow", "green", "white"  default green.
    :tlp c/TLPValue}
   (st/optional-keys
    {:source s/Str
     ;; if we have a source, we should have a source URI for more details
     :source_uri c/URI
     ;; The openC2 Actuator name that best fits the device that is
     ;; creating this sighting, is another "vocabulary"
     ;; see http://openc2.org/docs/OpenC2%20%20Language%20Descrip%20Doc%20Draft%20%28Rev%200%206f%29%2003012016.pdf
     ;; and Section Actuator Vocabulary, page 24
     :source_device s/Sensor  ;; eg. "network.firewall"
     ;; link to some random object
     :reference c/URI
     :confidence v/HighMedLow
     ;; The object(s) of interest. We need either an observable or an
     ;; indicator, as a Sighting is useless without one of them.  We
     ;; should enforce this in our own predicate.
     :observables [c/Observable]
     ;; the indicators we think we are seeing
     :indicators [rel/RelatedIndicator]
     ;; provide any context we can about where the observable came
     ;; from. ObservedRelations should be the current ctia.relations
     ;; namespace, moved into the ctia.schema.common namespace
     :relations [ObservedRelations]
     })))

src/ctia/http/routes/indicator.clj
```diff
@@ -91,6 +86,21 @@
...
+      :path-params [id :- s/Str]
+      :summary "Gets all Sightings associated with the Indicator"
+      (if-let [indicator (read-indicator @indicator-store id)]
+        (if-let [sightings (list-sightings-by-indicators @sighting-store [indicator])]
+          ;; (do
+          ;;   (println (str \u001b "[33m" (with-out-str (clojure.pprint/pprint indicator)) \u001b "[0m"))
+          ;;   (println (str \u001b "[34m" (with-out-str (clojure.pprint/pprint sightings)) \u001b "[0m"))
+          ;;   (try (s/validate [StoredSighting] sightings)
+          ;;        (catch Exception e
+          ;;          (println (str \u001b "[31m" (pr-str e) \u001b "[0m")))))
```
gbuisson: perhaps remove that dead code ?

test/ctia/http/routes/indicator_test.clj
```diff
@@ -256,3 +198,51 @@
                   (map :parsed-body)
                   (map #(dissoc % :id :created :modified :owner))
                   set)))))))
+
+(def api-key "45c1f5e3f05d0")
+(defn redprintln [& s]
+  (print "\u001b[31m")
+  (apply println s)
+  (print "\u001b[0m"))
```
gbuisson: if we keep this fn, we should put it on a helper.

test/ctia/http/routes/observable_test.clj
```diff
@@ -13,268 +21,121 @@
 
 (use-fixtures :each whoami-helpers/fixture-reset-state)
 
+(def api-key "45c1f5e3f05d0")
+(defn redprintln [& s]
+  (print "\u001b[31m")
+  (apply println s)
+  (print "\u001b[0m"))
```
gbuisson: see my comment above ^^^^^

test/ctia/http/routes/observable_test.clj
```diff
@@ -13,268 +21,121 @@
...
+      (when (get-in resp [:parsed-body :message])
+        (redprintln (get-in resp [:parsed-body :message])))
+      (when (get-in resp [:parsed-body :errors])
+        (redprintln (get-in resp [:parsed-body :errors])))
+      (is (= 200 (:status resp)))
+      (when (= 200 (:status resp))
+        (is (= (set expected-entity)
+               (set (:parsed-body resp))))
+        (:parsed-body resp)))))
+
```
gbuisson: all of these could be promoted to test helper fns imo

gbuisson:
Besides those small remarks I think it looks good !

test/ctia/http/routes/indicator_test.clj
polygloton: But we should be logging and not printing :)

yogsototh: This is only for testing.
But yes `log/error` would definitively feel better here.

Issue #223, opened 2016-04-15
Document Debutant use cases


polygloton:
In _doc/use_cases.md_, fill out all of the use cases with curl examples for the Debutant milestone.  Link to the use cases in _doc/roadmap.md_.

PR #243, opened 2016-04-21
Additional use cases

polygloton:
WIP for #223

### DIFF
--- a/doc/use_cases.md
+++ b/doc/use_cases.md
@@ -61,24 +61,28 @@ Extract the IDs from the 'related_campaigns.campaign_id' fields and
 For each indicator object returned by:
 
     curl http://ctiahost/ctia/ip/192.168.1.1/indicators
-    
+
 Extract the IDs from the 'suggested_COA.COA__id' fields and
 
     curl http://ctiahost/ctia/coa/ID
 
 ### 2.4 As an incident responder, I would like to record an incident and it's sightings
 
-Determine the ID of the incident's indicator(s).
+Determine the ID of the incident's indicator(s).  For example, if you know the title:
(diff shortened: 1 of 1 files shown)


gbuisson:
looks good !

doc/use_cases.md
```diff
@@ -46,22 +52,38 @@ For each jugement object returned by:
...
 
 ### 2.4 As an incident responder, I would like to record an incident and it's sightings
 
+Determine the ID of the incident's indicator(s).  For example, if you know the title:
+
+	curl http://ctiahost/ctia/indicator/title/document-direct-ip-traffic
+
+Or search for the indicator:
+
+    curl http://ctiahost/ctia/indicator/_search?title=ip # URL TBD
```
yogsototh: I have a simple question as it is related to my current WIP on searching indicator.

I know that we should use `GET` but, if we want to remove the size of the request limit and also have a better swagger integration and typing coercion. We might prefer to provide complex search expression via `POST` instead.

Imagine a complex request containing many different dimension all with a large cardinality.
Is this something we want to support?

polygloton: GET requests can have a body too

yogsototh: By proposing to use POST for a read-only method, I somehow break the REST principe. By using the body in a GET request we would apparently don't follow the HTTP1.1 recommandation to ignore the body on GET queries (while I don't think this is perfectly clear).

So I don't really like to use POST because it breaks a nice REST specification and also because it won't be natural for users.

Do you think that starting to use GET with standard query parameters and then change it only if we encounter limitations would be the right choice?

PR #251, opened 2016-04-22
Ensure ES Strict mapping, fix mapping mistakes/omissions, initial CRUD generative tests

### LINKED ISSUE #250: Fix ES Mappings
On mappings, setting the :dynamic property to false DOES NOT disable dynamic mapping.

The property has to be set to "strict" for it to work, thus we've been mislead on the tests and I uncovered some mapping mistakes.

Fix this.

gbuisson:
closes #250


polygloton:
Looks good.  Need a new issue to use defspec (will require some macro work) for the generative tests that also asserts that the result value matches.

PR #254, opened 2016-04-22
Jar based deployment, and some ES 2.3 mapping compat. changes

craigbro:


saintx:
Confirmed that this works against 1.7 and 2.3 ES.  Well done!

    saintx@bastion:~/dev/ctia$ curl 'localhost:9200/_cat/indices?v'
    health status index               pri rep docs.count docs.deleted store.size pri.store.size
    yellow open   .kibana               1   1          1            0      3.1kb          3.1kb
    yellow open   ctia                  5   1       1128            0      1.3mb          1.3mb
    yellow open   logstash-2016.04.22   5   1         36            0    200.5kb        200.5kb

Issue #255, opened 2016-04-22
TODOs for redis pubsub


polygloton:
These are follow-up tasks related to #23 

- [x] Use core.async/thread for blocking IO instead of green threads.  (For example ctia.lib.async/register-listener could take a 'type that allows :compute or :blocking, much like core.async/pipeline does).
- [x] Refactor ctia.lib.async to not know what an Event is.  Might be able to just change the terminology from Event to Entity.
- [ ] Handle exceptions in core.async go/thread blocks.  Don't want exceptions to be swallowed or allow it to kill threads.
- [x] Resolve: What is a redis store?
  - A redis store should not be a thing
- [x] Use hooks to publish redis events
- [x] No need to support subscribing to events (except for testing of course)

polygloton:
Here is some code that I wrote in a past project to abstract the green/real thread concern:

https://github.com/polygloton/pipeline/blob/master/src/pipeline/process/worker.clj#L90

polygloton:
I'm not worrying about error handling at this time since we are not currently using the events channel.

PR #261, opened 2016-04-23
Fixed Redis on CI

### LINKED ISSUE #256: Resolve redis issues on travis
For #23, we had trouble getting tests to pass on CI, even though they passed locally.  There is a test marked :disabled, which was failing because Redis on CI didn't publish to the subscribed channel.

polygloton:
- Adding environment tests for redis that may trigger when redis doesn't work
- Will fix redis in CI
- Will do some refactoring as I go

Fixes #256

### DIFF
--- a/src/ctia/lib/redis.clj
+++ b/src/ctia/lib/redis.clj
@@ -1,0 +1,31 @@
+(ns ctia.lib.redis
+  (:require [taoensso.carmine :as c]))
+
+;; Connections
+
+(defn server-connection
+  "Build the server config"
+  [[host port]]
+  (assert (and host port) "Redis has been de-configured")
+  {:pool {}
+   :spec {:host host
+          :port port
+          :timeout-ms 30000}})
+
+;; Pub/Sub
+
+(defn unsubscribe [pubsub-listener]
+  (c/with-open-listener pubsub-listener
+    (c/unsubscribe)))
+
+(defn subscribe [host-port event-channel-name listener-fn]
+  (c/with-new-pubsub-listener (:spec (server-connection host-port))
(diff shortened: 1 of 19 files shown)


polygloton:
Here is what I did to fix the issue.  I added a test that worked out pub/sub directly with carmine, doing some refactoring along the way to make it easier.  Though I confused myself at first by misinterpreting the results, it proved that redis was working correctly on CI.  Then I cleaned up the way that state is managed, especially in how it was being fixtures in the tests.  Things still failed until I realized that we were overriding the REDIS_ENABLED var in the travis config.  I believe that the reason tests were failing was due to state management and that flag causing havoc.  Why did it pass locally but not on CI?  I'm not entirely sure, but I suspect there was a difference in the order or tests being run.  In the end it doesn't matter; the state management code is better in this PR and tests are passing.

src/ctia/lib/redis.clj
```diff
@@ -0,0 +1,31 @@
...
+  (c/with-open-listener pubsub-listener
+    (c/unsubscribe)))
+
+(defn subscribe [host-port event-channel-name listener-fn]
+  (c/with-new-pubsub-listener (:spec (server-connection host-port))
+    {event-channel-name listener-fn}
+    (c/subscribe event-channel-name)))
+
+(defn publish [host-port event-chanel-name event]
+  (c/wcar (server-connection host-port)
```
polygloton: Rebuilding the server-connection map each time we publish is inefficient, but I will resolve that in issue #255

polygloton:
I also changed the way that ENV vars are defined in travis so that there is one per line.  This should make it much easier to verify.

PR #264, opened 2016-04-27
Improvements to CTIA startup and redis store

polygloton:
- Each store is configured with a distinct property
 - ES and Redis stores are the default (was memory store)
 - New "Events" store, backed by redis
   - Consolidated redis code into the redis store (and the lib)
   - Redis state is held in a defrecord instance
 - Moved ES state out of init ns

Related to #255


containers/demo/ctia/config/ctia.properties
```diff
@@ -1,14 +1,5 @@
...
-ctia.http.min-threads=10
-ctia.http.max-threads=100
-ctia.nrepl.port=3001
-ctia.nrepl.enabled=true
-ctia.store.type=es
-ctia.store.sql.db.classname=org.h2.Driver
-ctia.store.sql.db.subprotocol=h2
-ctia.store.sql.db.subname=/tmp/ctia-h2-db;DATABASE_TO_UPPER=false
-ctia.store.sql.db.delimiters=
+ctia.store.redis.uri=redis://redis:6379
```
quoll: I take it that these are names that refer to the redis and the elastic-services hosts?

saintx: @quoll Yes, one of the things that docker-compose does is provide name service, allowing your containers to reference one another without having to know ahead of time what their IP address will be.

saintx:
Looks good.

craigbro:
This is why I used ctia.store.default.type, or osmething similiarly named, awhile back.
Also, why ctia.store.events???
they are not stored, they are pubsubbed there
it's not a store!
does it implement the store protocol? nope 8^)

polygloton:
We need a way to support multiple different stores (like ES and SQL), and I think this does it elegantly.  Its not like anyone has to type out all of those properties unless they want to override the default.

craigbro:
I am fine with the explicit listing of properties, I love explicit 8^)

The only real blocking concern I have is the ctia.store.events ... property, and treating the redis pub/sub hooks as a store.

src/ctia/lib/map.clj
```diff
@@ -5,3 +5,12 @@
...
     (last vals)))
+
+(defn dissoc-in
+  [m [k & ks :as keys]]
+  (if ks
+    (if-let [nextmap (get m k)]
+      (let [newmap (dissoc-in nextmap ks)]
+        (assoc m k newmap))
+      m)
+    (dissoc m k)))
```
quoll: This is very manual. I'd rather see it based on update-in. e.g.:

(defn dissoc-in
  [m keys]
  (let [ks (butlast keys)
        lk (last keys)]
    (update-in m ks dissoc lk)))

polygloton: Cool

yogsototh: I've just needed a `dissoc-in` today, I simply used `(update-in m [ks] :dissoc last-key)`

PR #265, opened 2016-04-27
Refactored properties so that each store has its own implementation type

polygloton:
Also makes ES the default store.

Related to #255
Has bits of #264, which will be closed rather than merged


test/ctia/test_helpers/core.clj
```diff
@@ -59,49 +60,58 @@
...
-                     fixture-ctia]))
+(defn fixture-ctia
+  ([test] (fixture-ctia test true))
+  ([test enable-http?]
+   ;; Start CTIA
+   ;; This starts the server on a random port (if enabled)
+   (with-properties ["ctia.http.enabled" enable-http?
+                     "ctia.http.port" (if enable-http?
+                                        (available-port)
+                                        3000)]
```
quoll: Since this is open source, isn't it possible that this test could be run on all sorts of environments? Many of those systems may have certain ports, or port ranges locked down. I can see a benefit in making sure that the test code isn't making presumptions, but is this how it should be done?

polygloton: The available-port fn let's java pick a port.  The 3000 value is just a place holder for when http is disabled.

PR #266, opened 2016-04-28
Convert redis store into redis hook

### LINKED ISSUE #159: Refactoring handlers to accomodate hooks, and events and stores

polygloton:
Relates to #255
Closes #159 

- Implements redis event producer as a hook
- Wires-up ES event producer as hook
- Refactors flow hook code
- Removes dead code
- Needs issue #271 as post merge clean up

@quoll and I paired on this PR.  :)


.travis.yml
```diff
@@ -30,4 +30,4 @@ env:
       - CTIA_STORE_ES_INDEXNAME=elasticsearch
       - CTIA_PRODUCER_ES_URI=http://127.0.0.1:9200
```
gbuisson: do you think we should discard this and use CTIA_HOOK_ES instead ?

gbuisson:
this is a lot of work! sounds good to me !

yogsototh:
I really like it! 👍

.travis.yml
polygloton: Will do.

src/ctia/flows/autoload.clj
```diff
@@ -46,7 +46,11 @@
...
-                                                (finally
-                                                  nil))))))
+  ([]
+   (load-hooks!
+    (:hooks-classes (p/read))))
+  ([profiles :- [s/Keyword]]
+   (load-hooks!
+    (:hooks-classes (try (p/read "project.clj" profiles)
+                         (finally
+                           nil))))))
```
quoll: I realize that this was already here, but is it really wise to be reading configuration information from the build description file? I'm happy to do this in testing, but in production it is causing the conflation of the role of the build file.

polygloton: You are correct, it must go.  I'm actually working on a refactor for the autoloader and will submit it in a different PR, maybe tomorrow.

PR #267, opened 2016-04-29
Initial List Paging Support

### LINKED ISSUE #221: Implement list pagination
Support pagination when more then one result is returned.

gbuisson:
Closes #221 

- [x] Atom Store Pagination & sort
- [x] SQL Store Pagination & sort
- [x] ES Store Pagination & sort
- [x] Pagination response Headers (total hits, next and previous URIs)
- [x] Use pagination on all list routes

I need someone to review allowed sort field options for each route.


test/ctia/http/routes/pagination_test.clj
```diff
@@ -0,0 +1,142 @@
...
+                                                           :value "1.2.3.4"}
+                                              :disposition 5
+                                              :disposition_name "Unknown"})))
+          judgement-post-responses (map #(post "ctia/judgement"
+                                               :body %
+                                               :headers headers) new-judgements)
+          stored-judgements (->> judgement-post-responses
+                                 (map :body))]
+
+      (is (= doc-count (count stored-judgements)))))
```
yogsototh: Wouldn't it be better to add a test for each entity, just to verify, we didn't forgot to add pagination for some routes?

polygloton: Maybe this could be added to the various entity route tests

src/ctia/http/routes/common.clj
```diff
@@ -0,0 +1,38 @@
...
+                            (map->header-value v)
+                            (str v))}) headers)))
+
+(defn paginated-ok
+  "returns a 200 with the supplied response
+   and its metas as headers"
+  [response]
+  {:status ok
+   :body response
+   :headers (map->paging-headers (meta response))})
```
polygloton: Would it be easy to pass in the paging-headers explicitly?  I prefer explicit over meta if it isn't a huge pain to change.

src/ctia/http/routes/judgement.clj
```diff
@@ -46,12 +48,13 @@
...
       :tags ["Feedback"]
       :return (s/maybe [StoredFeedback])
+      :query [params FeedbacksByJudgementQueryParams]
       :path-params [judgement-id :- s/Str]
       :header-params [api_key :- (s/maybe s/Str)]
       :capabilities #{:read-feedback :admin}
       :summary "Gets all Feedback for this Judgement."
-      (if-let [d (list-feedback @feedback-store {:judgement judgement-id})]
-        (ok d)
+      (if-let [d (list-feedback @feedback-store {:judgement judgement-id} params)]
```
polygloton: I like that you are passing the params to the store.  That means that the store can optionally optimize queries.

src/ctia/http/routes/common.clj
AlexBaranosky: 9 out of ten times I've used metadata I've regretted it. Why not just pass pagination information a it's own argument?

gbuisson: Yeah It's not so good after all, I'll change it and stop using metas.

gbuisson:
Based on feedback I'll:
- [X] replace metas with maps
- [ ]  move paging tests on each entity, make them more in depth

gbuisson:
Currently we follow judgement --> indicators relation to list` indicators` by an `observable`,
may we change that and use the `observable` field of indicator?
same goes for sightings (it is being changed by Yann's PR).

When these 2 list workflows will be changed, then I could move pagination tests to their related route entity test, for now I had to keep indicators and sightings pagination tests on observable tests.

polygloton:
@craigbro Will have to answer this question.  ^^^

gbuisson:
I need to fixup pagination tests to comply with the work of @yogsototh then it will be ready.

src/ctia/http/routes/common.clj
```diff
@@ -0,0 +1,42 @@
...
+  {(s/optional-key :sort_by) (describe (s/enum :id) "Sort results on a field")
+   (s/optional-key :sort_order) (describe (s/enum :asc :desc) "Sort direction")
+   (s/optional-key :offset) (describe Long "Pagination Offset")
+   (s/optional-key :limit) (describe Long "Pagination Limit")})
+
+(defn map->paging-header-value [m]
+  (str/join "&" (map (fn [[k v]]
+                       (str (name k) "=" v)) m)))
+
+(defn map->paging-headers
```
craigbro: Why as headers and not in the body itself?  I would rather have PaginatedJudgements and other appropriate wrapper objects.  This is not a blocker for the merge, IMO.

src/ctia/store.clj
```diff
@@ -5,81 +5,79 @@
   (create-actor [this new-actor])
   (update-actor [this id actor])
   (delete-actor [this id])
-  (list-actors [this filtermap]))
+  (list-actors [this filtermap params]))
```
craigbro: Is it possible to strongly define the contents of the params map, using a schema, or perhaps a destructuring keyword spec?  I would at least want the expected contents of params to be documented in docstrings in the store protocol definitions.

PR #275, opened 2016-05-04
Resolved identified todo's from redis work

### LINKED ISSUE #255: TODOs for redis pubsub

polygloton:
- Ns "lib.async" no longer refers to events
- May use a real thread when registering a listener

Closes #255

### DIFF
--- a/src/ctia/lib/async.clj
+++ b/src/ctia/lib/async.clj
@@ -2,7 +2,7 @@
   (:require [ctia.lib.time :as time]
             [ctia.schemas.common :as c]
             [ctia.schemas.verdict :as v]
-            [clojure.core.async :as a :refer [go-loop alt! chan tap]]
+            [clojure.core.async :as a :refer [go-loop alt! alt!! chan tap thread]]
             [schema.core :as s :refer [=>]])
   (:import [clojure.core.async.impl.protocols Channel]
            [clojure.core.async.impl.buffers FixedBuffer]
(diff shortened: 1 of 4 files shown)


src/ctia/lib/async.clj
```diff
@@ -63,28 +77,40 @@
...
+               :listener-fn listener-fn
+               :events events
+               :end-chan end-chan)
+
+       :compute
+       (listen :mode :compute
+               :pred pred
+               :listener-fn listener-fn
+               :events events
+               :end-chan end-chan))
```
yogsototh: I might forgot something here, but why a case?
Why not:

~~~ clojure
(listen :mode mode
    :pred pred
    :listener-fn listener-fn
    :events events
    :end-chan end-chan)
~~~

yogsototh:
Except from my remark sounds good to me.

src/ctia/lib/async.clj
polygloton: Since listen is a macro, it needs to be passed the mode at compile time.  It subtle.  :)

gbuisson:
sounds good!

PR #284, opened 2016-05-12
Issue 19 indicator routes

### LINKED ISSUE #19: All API endpoints for Judgements, Verdicts and Indicators working
We want all endpoints for judgements, verdicts and indicators to be functional.  This does NOT include all TTP, Campaign and Actor endpoints.

yogsototh:
Closes #19

### DIFF
--- a/src/ctia/http/routes/indicator.clj
+++ b/src/ctia/http/routes/indicator.clj
@@ -49,11 +49,15 @@
           (not-found))
         (not-found)))
     (GET "/:id/coas" []
-      :tags ["COA"]
       :return [StoredCOA]
-      :path-params [id :- Long]
+      :path-params [id :- s/Str]
+      :query [params SightingsByIndicatorQueryParams]
       :summary "Gets all COAs associated with the Indicator"
-      (not-found))
+      (if-let [indicator (read-indicator @indicator-store id)]
+        (if-let [coas (list-coas-by-indicators @coa-store [indicator] params)]
+          (paginated-ok coas)
+          (not-found))
+        (not-found)))
     (GET "/:id/ttps" []
       :tags ["TTP"]
       :return [StoredTTP]
--- a/src/ctia/http/routes/indicator.clj
+++ b/src/ctia/http/routes/indicator.clj
@@ -40,9 +40,14 @@
       (not-found))
     (GET "/:id/campaigns" []
       :return [StoredCampaign]
-      :path-params [id :- Long]
+      :path-params [id :- s/Str]
+      :query [params SightingsByIndicatorQueryParams]
       :summary "Gets all Campaigns associated with the Indicator"
-      (not-found))
+      (if-let [indicator (read-indicator @indicator-store id)]
+        (if-let [campaigns (list-campaigns-by-indicators @campaign-store [indicator] params)]
+          (paginated-ok campaigns)
+          (not-found))
+        (not-found)))
     (GET "/:id/coas" []
       :tags ["COA"]
       :return [StoredCOA]
--- a/src/ctia/http/routes/indicator.clj
+++ b/src/ctia/http/routes/indicator.clj
@@ -35,9 +35,14 @@
     :tags ["Indicator"]
     (GET "/:id/judgements" []
       :return [StoredJudgement]
-      :path-params [id :- Long]
+      :path-params [id :- s/Str]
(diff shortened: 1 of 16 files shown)


src/ctia/http/routes/indicator.clj
```diff
@@ -35,26 +35,44 @@
     :tags ["Indicator"]
     (GET "/:id/judgements" []
       :return [StoredJudgement]
-      :path-params [id :- Long]
+      :path-params [id :- s/Str]
+      :query [params SightingsByIndicatorQueryParams]
       :summary "Gets all Judgements associated with the Indicator"
-      (not-found))
+      (if-let [indicator (read-indicator @indicator-store id)]
+        (if-let [judgements (list-judgements-by-indicators @judgement-store [indicator] params)]
```
polygloton: I'm thinking about this style of list method.  In #288, there is a new ID protocol and a CtiaID record.  What do you think about passing in a set of CtiaID instances instead of the list of records?  The problem is that these routes receive the short-ID (as in :id here), but the references need to become URLs.  The ID protocol has methods for getting either the (.short_id ...) or the (.long_id ...), so the list methods can use whatever it needs.  Also, I am thinking that I would like to avoid parsing the IDs in the store method implementations.  Instead, we can do it once here and then pass the CtiaID records down.

Still... I'm not sure.  Let me know what you think.

yogsototh: While writing these I would have preferred to have a `list-X-by-indicators-ids` instead.
So yes I am with you here.

Do you have something like this in mind?:

~~~
(let [indicator (read-indicator @indicator-store id)
      judgement-ctia-ids (map #(-> % :judgement_id make-ctia-id) (:judgements indicator))]
    (if-let [jugements (list-judgements-by-indicator-ids @judgement-stor judgement-ctia-ids params)]
...
~~~

where for now I don't really know how to write `make-ctia-id`

polygloton: I'm trying to get it working locally.  I added this to ctia/http/routes/indicator.clj:

```clojure
(defn ->id [indicator-id]
  (id/->id :indicator
           indicator-id
           (get-in @properties [:ctia :http :show])))
```

Its a function that turns "short IDs" into CtiaID instances.

src/ctia/http/routes/indicator.clj
```diff
@@ -35,26 +35,44 @@
...
       :return [StoredCOA]
-      :path-params [id :- Long]
+      :path-params [id :- s/Str]
+      :query [params SightingsByIndicatorQueryParams]
       :summary "Gets all COAs associated with the Indicator"
-      (not-found))
+      (if-let [indicator (read-indicator @indicator-store id)]
+        (if-let [coas (list-coas-by-indicators @coa-store [indicator] params)]
+          (paginated-ok coas)
+          (not-found))
```
gbuisson: I think we already discussed this, 
imo it sounds wrong to return a not-found if there's no record, why not just return an empty list, or no-content ?

the 404 is justified for verdict route though.

test/ctia/test_helpers/http.clj
```diff
@@ -27,7 +27,7 @@
     (is (= 200 status))
     (when (= 200 status)
       (is (= new-entity
-             result))
+             (select-keys result (keys new-entity))))
```
gbuisson: good idea :-)

gbuisson:
Besides the 404 on empty lists I think it sounds good, but those were present before this task so I don't think it's a blocker.

src/ctia/http/routes/indicator.clj
polygloton: Its the standard RESTful way to handle it.  The record was not found, and entity is what is being pointed to.  If we replaced each entity with a static web page, and web page entity-123.html didn't exist, then you would get a 404.

gbuisson: It would be ok for a single document, but in the case of the list I'm not sure about that, imo a client getting a 404 on such a request would think he mistyped the url.

polygloton: I agree with you about the list routes.  An empty search result is still a success.

yogsototh:
Regarding to the conclusion of our discussion, I believe we should get rid of this PR entirely as it only added routes which we said should be discarded.

PR #287, opened 2016-05-12
Added feedback on all entities

### LINKED ISSUE #222: Feedback on everything

gbuisson:
This adds Feedback routes for all entities

closes #222


craigbro:
````
We discussed a radical simplification of feedback routes in the channel:

since we now have canonical IDs for all our entities

the entity/entity-id split in the paths is kinda bogus

seems to me....

that feedback just becoming

POST /feedback

GET /feedback?entity_id=http://ctia.alpha.....

and that's it

File: test/ctia/test_helpers/generators/schemas/feedback_generators.clj [hunk 1]
```diff
@@ -0,0 +1,30 @@
...
+  (gen/fmap
+   (fn [[id entity-id]]
+     (complete
+      Feedback
+      (cond-> {}
+        id (assoc :id id)
+        entity-id (assoc :entity_id entity-id))))
+   (gen/tuple
+    (gen-id/gen-short-id-of-type :feedback)
+    (gen-id/gen-short-id-of-type :judgement))))
```
polygloton: The entity-id should be a long ID.  I don't think it holds up the PR, though.

polygloton:
Looks good!

PR #288, opened 2016-05-12
IDs can be URLs and users can submit IDs

### LINKED ISSUE #212: Semantically valuable IDs
- [x] IDs are URLs that match the entity GET route
- [x] Users can optionally submit a semantically valuable ID
  - [x] CTIA will convert the user submitted string into a URL to the entity
  - [x] String should be URL encoded
- [x] If user supplies an ID, don't prepend the type to it (its already in the URL)

### LINKED ISSUE #252: Refactor handler tests for generative testing
In #251 we added some generative tests.  It needs some refactoring.

- [ ] Use defspec and forall in the tests so that we get failure reduction
- [ ] Add a new defspec-for-each-fixture type of macro, similar to the deftest one
- [ ] Make sure that we have enough coverage with these new tests
- [ ] If the tests are slow (which is likely with generative testing), then mark them with a :slow test selector that doesn't run by default

You might get inspiration from this code:

https://github.com/polygloton/ctia/blob/issue-169-spike/test/ctia/domain/id_test.clj

### LINKED ISSUE #269: Generative tests should shrink
We have generative tests that take, for example, 100 samples and then applies them to an expectation.  When there is test failure, however, that means that we get up to 100 test failure prints for each test (and there are several).  That sort of misses the point of test.check.  We need to take advantage of the features for shrinking tests failures to a single failure message.

See for example:
https://github.com/polygloton/ctia/blob/issue-169-spike/test/ctia/domain/id_test.clj

polygloton:
Also refactored tests
 - generative tests shrink
 - http route tests are separated from their generative tests

Closes #212, closes #252, and closes #269.


project.clj
```diff
@@ -10,7 +10,9 @@
              "-server"]
   :dependencies [[org.clojure/clojure "1.7.0"]
                  [clj-time "0.9.0"] ; required due to bug in lein-ring
-                 [metosin/schema-tools "0.7.0"]
+                 [prismatic/schema "1.0.5"]
```
polygloton: This is the version that compojure-api wants.  It was getting replaced by an older version.

src/ctia/http/routes/actor.clj
```diff
@@ -12,7 +12,7 @@
     (POST "/" []
       :return StoredActor
       :body [actor NewActor {:description "a new Actor"}]
-      :header-params [api_key :- s/Str]
+      :header-params [api_key :- (s/maybe s/Str)]
```
polygloton: The generative tests (using allow-all auth now) helped catch this bug in a few places.

src/ctia/http/routes/judgement.clj
```diff
@@ -40,7 +46,7 @@
       :summary "Adds a Feedback to a Judgement"
       :capabilities #{:create-feedback :admin}
       :login login
-      (ok (flows/create-flow :realize-fn #(realize-feedback %1 %2 %3 judgement-id)
+      (ok (flows/create-flow :realize-fn #(realize-feedback %1 %2 %3 (->id judgement-id))
```
polygloton: Stores use "short IDs" for the primary IDs of entities, but "long IDs" for references.  When the route has a short ID for a reference, it needs to be converted into a long ID.

project.clj
```diff
@@ -51,34 +54,37 @@
...
+                                    :es-producer-filtered-alias
+                                    :es-producer-aliased-index}
+                                  %)
+                   :default #(not= :disabled %)
+                   :integration #(#{:es-store
+                                    :integration
+                                    :es-producer
+                                    :es-producer-filtered-alias
+                                    :es-producer-aliased-index}
+                                  %)
```
quoll: The sets in these functions are already functions, and don't need to be wrapped in an anonymous function. Just leave them as sets.
e.g.
`:es-producer #{:es-producer
                :es-producer-filtered-alias
                :es-producer-aliased-index}`

src/ctia/schemas/vocabularies.clj
```diff
@@ -355,56 +355,50 @@
...
-   :process.file-scanner
-   :process.location-service
-   :process.network-scanner
-   :process.remediation-service
-   :process.reputation-service
-   :process.sandbox
-   :process.virtualization-service
-   :process.virtualization-service
-   :process.vulnerability-scanner))
+   Section Actuator Vocabulary, page 24."
```
polygloton: Using strings here instead of keywords because I think we should not leak the ':' char in our API.  Also there is value in following the established pattern and we should discuss diverging from the pattern before doing it.  In this case, I don't think there was enough benefit in using keywords to justify the awkwardness in the API (having the ':' char in our fields).

test/ctia/domain/id_test.clj
```diff
@@ -0,0 +1,73 @@
+(ns ctia.domain.id-test
```
polygloton: Notice the mixture of specs and tests in this file.  It works well because there are no fixtures and all of the functions being tested are referentially transparent.

project.clj
AlexBaranosky: What @quoll said :)

src/ctia/schemas/vocabularies.clj
AlexBaranosky: Consistency is a beautiful thing.

test/ctia/domain/id_test.clj
AlexBaranosky: I secretly despise clojure.test fixtures because of the way they glom onto everything in a ns.

test/ctia/http/generative/atom_store_spec.clj
```diff
@@ -0,0 +1,38 @@
+(ns ctia.http.generative.atom-store-spec
```
polygloton: This is kind of new and weird.  The tests are grouped by the test setup rather then by the entity; They are grouped by the store and not by the entity.  This makes a lot of sense for generative tests since each defspec below might trigger hundreds of API calls.  It is best to have a :once fixture at the top and not do any :each for each property being tested.  As a consolation, this is all just glue code and the specs are defined in one place.

test/ctia/http/generative/specs.clj
```diff
@@ -0,0 +1,87 @@
+(ns ctia.http.generative.specs
```
polygloton: This will probably grow and will need to be split up later, but it is concise at the moment, so its in one file.

src/ctia/domain/id.clj
```diff
@@ -0,0 +1,109 @@
...
+         "/ctia/"
+         type
+         "/"
+         (url/encode short-id))))
+
+(defn long-id?
+  "Make an educated guess if this is a long-form ID"
+  [^String id]
+  ;; TODO - More specific checking?
+  (boolean (.startsWith id "http")))
```
quoll: is it necessary to explicitly cast a java.lang.Boolean to a java.lang.Boolean? I suggest getting rid of it, and changing the defn to:
`(s/defn long-id? :- s/Bool [^String id] (.startsWith id "http"))`
(with docs, of course)

src/ctia/domain/id.clj
```diff
@@ -0,0 +1,109 @@
...
+    "The short part of the ID")
+  (long-id [this] [this url-params]
+    "Convert this record to a long ID string"))
+
+(defrecord CtiaId [hostname short-id path-prefix port protocol type]
+  ID
+  (localize [this url-show-params]
+    (merge this
+           (-> url-show-params
+               (update :path-prefix not-empty))))
```
quoll: Is this planning to have multiple update/assoc/dissoc in future? If not, then the threading macro isn't needed

test/ctia/http/routes/observable_test.clj
```diff
@@ -1,100 +1,312 @@
 (ns ctia.http.routes.observable-test
```
polygloton: There is value in readable examples in tests.  Even though they are long-winded, the artifact of the test is documentation, almost like a CURL example (but way better IMO).  *Maybe* generative testing will be able to replace it, but not yet.  For one thing, everyone on the team will need to be good at writing and debugging generative tests.  We're not there yet.  In the mean time, we need to keep scenario oriented, hand-written tests maintained while also maintaining separate generative tests when they augment the hand-written tests.

I *feel* like generative tests will usually give the most value for referentially-transparent code that doesn't state management.  That said, I do feel like the generative tests that we have written for the HTTP APIs were helpful.  (They were probably helpful mostly because our test coverage was poor).  IMO the jury is still out.

I've restored the original observable tests.  They may be ugly, but I actually feel like they are examples of very good testing (it tests a real scenario that a user might have to to follow using the API).

test/ctia/http/routes/pagination_test.clj
```diff
@@ -1,59 +1,59 @@
 (ns ctia.http.routes.pagination-test
```
polygloton: This test generates data, but I am not considering it a "generative test".  For one, it does not loop over the generated content and make (is (= ...)) style assertions (not after this commit at least).  Secondly, this test is now making assertions on the status of the POSTing of generated data.  This test is testing pagination, not the POSTing of generated data (we have generative tests that do that now).  Making assertions on the POSTed entities means that the test will fail fast if the test setup fails, and we won't end up with tens or hundreds of failing test cases being printed out.

src/ctia/domain/id.clj
polygloton: I think it was just cruft.  I'll get rid of the (boolean ...).

src/ctia/domain/id.clj
polygloton: Will fix, thanks.

test/ctia/test_helpers/generators/schemas.clj
```diff
@@ -0,0 +1,64 @@
...
+             [sighting-generators :as sg]
+             [ttp-generators :as tg]]))
+
+(def gen-new-indicator-with-new-sightings
+  (gen/fmap
+   (fn [[indicator sightings]]
+     [indicator
+      (map (fn [sighting]
+             (assoc sighting :indicators
+                    [{:indicator_id (:id indicator)}]))
```
polygloton: This is wrong.  The :indicator_id values need to be URLs ('long IDs'), and the store methods need to handle them correctly.

src/ctia/flows/crud.clj
```diff
@@ -1,14 +1,21 @@
...
             [ctia.events.obj-to-event :refer [to-create-event
                                               to-update-event
                                               to-delete-event]]))
 
+(defn find-id
+  "Lookup an ID in a given entity.  Parse it, because it might be a
+   URL, and return the short form ID.  Returns nil if the ID could not
+   be found."
+  [{id :id}]
+  (when (not-empty id)
```
quoll: When used as a predicate, `seq` is better use use than `not-empty`

src/ctia/lib/map.clj
```diff
@@ -1,7 +1,26 @@
...
     (apply merge-with rmerge vals)
     (last vals)))
+
+;; from http://stackoverflow.com/questions/21768802/how-can-i-get-the-nested-keys-of-a-map-in-clojure
+(defn keys-in [m]
+  (if (map? m)
+    (vec
+     (mapcat (fn [[k v]]
+               (let [sub (keys-in v)
+                     nested (map #(into [k] %) (filter (comp not empty?) sub))]
```
quoll: use `seq` rather than `(comp not empty?)`

test/ctia/test_helpers/generator.clj
```diff
@@ -1,67 +0,0 @@
-(ns ctia.test-helpers.generator
-  (:require [ctia.lib.time :as time]
```
gbuisson: I forgot it was still there :-)

polygloton: I think most of it is moved to other namespaces now

src/ctia/flows/crud.clj
polygloton: You are right.  Something to train myself on:  Only use not-empty when I care about the type of the result.

src/ctia/schemas/vocabularies.clj
yogsototh: Actually the `:` wasn't visible in the swagger model, and so didn't leaked in our API. I checked in the sighting model in swagger-ui. `compojure-api` use `name` for keyword values.

The JSON to POST at a sighting creation shouldn't contains `:`:
~~~
{ ...
  "source_device": "endpoint.printer",
  ...
~~~

That said, I'm completely ok to use strings as they are used for all other schema in vocabularies to maintain consistency.

yogsototh:
For me this look very good 👍

gbuisson:
sounds good too! well done!

PR #294, opened 2016-05-16
Refactor generative tests

### LINKED ISSUE #293: Refactor some generative tests to shrink
Any tests that generate data and then loop over the data, making expectations (is (= ...)) on the generated data, should be considered a generative test and should use defspec so that it shrinks.  Most of these have been refactored already, but some still remain that use deftest.

polygloton:
Closes #293

### DIFF
--- a/test/ctia/events/obj_to_event_test.clj
+++ b/test/ctia/events/obj_to_event_test.clj
@@ -1,43 +1,39 @@
 (ns ctia.events.obj-to-event-test
   (:require [clojure.data :refer [diff]]
+            [clojure.test.check.clojure-test :refer [defspec]]
+            [clojure.test.check.properties :refer [for-all]]
             [ctia.events.obj-to-event :as evt]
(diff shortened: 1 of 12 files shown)


test/ctia/events/obj_to_event_test.clj
```diff
@@ -1,43 +1,39 @@
...
-(t/deftest test-to-delete-event
-  (doseq [actor (generators/sample 100 StoredActor)]
-    (let [event (evt/to-delete-event actor)]
-      (t/is (s/validate DeleteEvent event)))))
+(defspec spec-to-delete-event
+  (for-all [actor (gen/gen-entity :actor)]
+           (s/validate DeleteEvent (evt/to-delete-event actor))))
 
-(t/deftest test-triplet-generation
-  (t/is
```
quoll: I was going to ask how this even worked before, but then realized that with no type checking the second parameter was considered to be the failure message

PR #295, opened 2016-05-17
Protect properties route with :admin capability

### LINKED ISSUE #253: Protect properties route with capability
Need a new capability, or maybe just require :admin.

polygloton:
Closes #253

### DIFF
--- a/src/ctia/http/routes/properties.clj
+++ b/src/ctia/http/routes/properties.clj
@@ -1,10 +1,14 @@
 (ns ctia.http.routes.properties
   (:require [compojure.api.sweet :refer :all]
+            [ctia.properties :refer [properties]]
             [ring.util.http-response :refer :all]
-            [ctia.properties :refer [properties]]))
+            [schema.core :as s]))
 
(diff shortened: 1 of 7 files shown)


src/ctia/http/routes/properties.clj
```diff
@@ -1,10 +1,14 @@
...
 
 (defroutes properties-routes
   (context "/properties" []
-    :tags ["Properties"]
-    (GET "/" []
-      (ok @properties))))
+           :tags ["Properties"]
+           :summary "The currently running properties"
+           :capabilities :admin
+           :header-params [api_key :- (s/maybe s/Str)]
```
quoll: This type information is not in the context of a schema.core macro. So the `:-` symbol and the `(s/maybe s/Str)` expression both end up being treated as allowable params. Of course, these will never be passed, so they are essentially ignored. But they're not type info, and they're not comments, so they shouldn't be there in this form.

polygloton: Isn't that part of the defroutes macro syntax?

PR #297, opened 2016-05-19
ES Producer Hook tests rewrite

### LINKED ISSUE #271: Rewrite ES event hook/producer tests
In PR #266, I deleted the old [producer tests](https://github.com/threatgrid/ctia/blob/e138cb9aa848d7c2a6ff9401ae5260008e01b8db/test/ctia/events/producer_test.clj), because I deleted the old producer implementation.

Need to write new tests as a sibling to the [redis event hook tests](https://github.com/polygloton/ctia/blob/issue-255-part1b/test/ctia/flows/hooks/redis_event_hook_test.clj).  The new ES hook tests would need to trigger the hook as a side effect, as is done for the redis event hook tests, and then query ES to find the events are there.

gbuisson:
closes #271 

- Tests the ES producer hook on both slicing strategies
- also had to change test selectors, see my comment for more details.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -54,15 +54,11 @@
                    :sql-store :sql-store
                    :es-store :es-store
                    :disabled :disabled
-                   :es-producer #{:es-producer
-                                   :es-producer-filtered-alias
-                                   :es-producer-aliased-index}
                    :default #(not= :disabled %)
-                   :integration #{:es-store
-                                   :integration
-                                   :es-producer
-                                   :es-producer-filtered-alias
-                                   :es-producer-aliased-index}
+                   :integration #(or (:es-store %)
+                                     (:integration %)
+                                     (:es-filtered-alias %)
+                                     (:es-aliased-index %))
                    :all #(not (:disabled %))}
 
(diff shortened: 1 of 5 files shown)


project.clj
```diff
@@ -54,15 +54,11 @@
...
                    :disabled :disabled
-                   :es-producer #{:es-producer
-                                   :es-producer-filtered-alias
-                                   :es-producer-aliased-index}
                    :default #(not= :disabled %)
-                   :integration #{:es-store
-                                   :integration
-                                   :es-producer
-                                   :es-producer-filtered-alias
-                                   :es-producer-aliased-index}
```
gbuisson: this selector syntax was not working, using `lein test :integration` launched 0 tests so I reverted back to use a fn instead. tell me if I got it wrong.

yogsototh:
LGTM

polygloton:
It says rewrite, but nothing was deleted.  Is it meant to be just new tests?

gbuisson:
Tests were deleted with PR #266 you added a task to implement them again.

PR #299, opened 2016-05-19
added reversed search routes for indicators

### LINKED ISSUE #19: All API endpoints for Judgements, Verdicts and Indicators working

yogsototh:
Closes #19 

I also added this route:

- [x] `/indicator/:id/indicators` for retrieving indicators related to some indicator id

 - handle relationship querying for ES
 - removed deprecated routes see #19
 - removed some comments
 - added test

### DIFF
--- a/src/ctia/http/routes/indicator.clj
+++ b/src/ctia/http/routes/indicator.clj
@@ -1,5 +1,7 @@
 (ns ctia.http.routes.indicator
   (:require [compojure.api.sweet :refer :all]
+            [ctia.domain.id :as id]
+            [ctia.properties :refer [properties]]
             [ctia.flows.crud :as flows]
             [ctia.http.routes.common :refer [PagingParams paginated-ok]]
             [ctia.schemas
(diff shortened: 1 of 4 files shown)


src/ctia/http/routes/indicator.clj
```diff
@@ -117,6 +91,65 @@
...
+    :tags ["Indicator"]
+    :return (s/maybe [StoredIndicator])
+    :summary "Gets all indicators referencing some judgement"
+    :query [params IndicatorsListQueryParams]
+    :path-params [id :- s/Str]
+    :header-params [api_key :- (s/maybe s/Str)]
+    :capabilities #{:list-indicators-by-title :admin}
+    (paginated-ok
+     (list-indicators @indicator-store
+                      {:judgements #{{:judgement_id id}}}
```
polygloton: This id would need to be a long-ID (a URL), but here it is a short-ID from the URL.  We need to either pass in a CtiaID instance instead of the reference map, or else we need to convert the ID into a long-ID.  Code for either approach can be found in ctia.domain.id.

test/ctia/test_helpers/http.clj
```diff
@@ -42,7 +42,9 @@
               :body new-entity
               :headers {"api_key" api-key})]
     (when (not= 200 status)
-      (throw (ex-info (str "Expected status to be 200 but was %s" status)
+      (throw (ex-info (str "Expected status to be 200 but was " status
+                           " for " path ":\n"
+                           (with-out-str (clojure.pprint/pprint response)))
```
polygloton: Good fix.  I think I had meant to use format instead of str.

yogsototh:
Argh! I rebased my PR so your comment was erased @polygloton . I use long-ids inside relationship now.

polygloton:
Looks good.

test/ctia/test_helpers/http.clj
polygloton: The ex-info fn takes a map as second arg.  The response should be that map.  If you want to print out errors like this, add a (try ... (catch ...)) to ctia.init somewhere that uses ex-data to get the map and pprint it.  Its more idiomatic and lets you avoid having a lot of this kind of (with-out-str ...) code all over.

polygloton:
Can you address this [commented out test](https://github.com/threatgrid/ctia/blob/master/test/ctia/http/generative/specs.clj#L41) in your PR?  Either fix it or delete it.

test/ctia/test_helpers/http.clj
AlexBaranosky: @yogsototh if you're just going to string-concat, then why use `ex-info` at all. You could just use `(Exception. (str a b c))`

AlexBaranosky: @yogsototh @polygloton I recommend you catch ExceptionInfos and also SQLExceptions as special cases, sort of like this: https://github.com/threatgrid/threatbrain/blob/master/src/threatbrain/lib/utils.clj#L6

This is because ex-infos have lots of useful info you'll want to handle specially. Also, SQLExceptions always hide the useful information in the .getNextException ... so if you don't get that next exception you will basically have a useless stacktrace.

yogsototh:
@polygloton I tried to fix the test, the first things to do was to add the long-ids but I guess I'm tired I couldn't make it work correctly, I made a WIP branch here:

https://github.com/yogsototh/ctia/tree/generative-sighting-test-fix

PR #301, opened 2016-05-19
Added commit number of the build on version info

### LINKED ISSUE #177: The /version should include the commit number of the build.
When I ask for the version, I should get the release number, as well as the build number (the commit ID).

gbuisson:
closes #177

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -55,18 +55,26 @@
                    :es-store :es-store
                    :disabled :disabled
                    :es-producer #{:es-producer
-                                   :es-producer-filtered-alias
-                                   :es-producer-aliased-index}
+                                  :es-producer-filtered-alias
+                                  :es-producer-aliased-index}
                    :default #(not= :disabled %)
                    :integration #{:es-store
-                                   :integration
-                                   :es-producer
-                                   :es-producer-filtered-alias
-                                   :es-producer-aliased-index}
+                                  :integration
+                                  :es-producer
+                                  :es-producer-filtered-alias
+                                  :es-producer-aliased-index}
                    :all #(not (:disabled %))}
 
   :java-source-paths ["hooks/ctia"]
   :javac-options  ["-proc:none"] ;; remove a warning
+  :filespecs [{:type :fn
+               :fn (fn [p]
+                     {:type :bytes :path "ctia-version.txt"
+                      :bytes (str (:out (clojure.java.shell/sh
+                                         "git" "log" "-n" "1" "--pretty=format:%H "))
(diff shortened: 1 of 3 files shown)


project.clj
```diff
@@ -63,6 +63,14 @@
...
   :javac-options  ["-proc:none"] ;; remove a warning
+  :filespecs [{:type :fn
+               :fn (fn [p]
+                     {:type :bytes :path "ctia-version.txt"
+                      :bytes (str (:out (clojure.java.shell/sh
+                                         "git" "log" "-n" "1" "--pretty=format:%H "))
+                                  (:out (clojure.java.shell/sh
+                                         "git" "symbolic-ref" "--short" "HEAD")))})}]
+
   :profiles {:dev {:dependencies [[cheshire "5.5.0"]
```
polygloton: This is cool

src/ctia/http/routes/version.clj
```diff
@@ -1,7 +1,17 @@
...
+            [clojure.string :as st]
+            [ctia.schemas.common :refer [VersionInfo]]
+            [compojure.api.sweet :refer :all]
+            [ring.util.http-response :refer :all]))
+
+(defn current-build []
+  (if-let [built-version (io/resource "ctia-version.txt")]
+    built-version
+    (str (:out (shell/sh "git" "log" "-n" "1" "--pretty=format:%H "))
+         (:out (shell/sh "git" "symbolic-ref" "--short" "HEAD")))))
```
polygloton: Let's memo this, or better yet use an atom explicitly like...

```clojure
(or @version (reset version (current-build)))
```

gbuisson: good idea

gbuisson: or just use a def instead ?

yogsototh: I would give my preference to the atom.

AlexBaranosky: avoiding running code at compile time is a rule thumb to save you much pain. .(that `def` would be running code on load 👎 )

Why not just:

```clj
(def current-build
  (memoize (fn []
             (if-let [built-version (io/resource "ctia-version.txt")]
               built-version
               (str (:out (shell/sh "git" "log" "-n" "1" "--pretty=format:%H "))
                    (:out (shell/sh "git" "symbolic-ref" "--short" "HEAD")))))))
```

gbuisson: agreed ;)

PR #302, opened 2016-05-20
Fix search sightings by indicator and other short/long id issues

yogsototh:
This PR should fix many bugs related to new short/long ids.

About this route: `/observable/:type/:value/indicators`
It currently works by listing all judgements related to an observable and from that list of judgements, retrieve all related indicators. 

Since each Judgement could reference external indicators, we might not want to keep this route for the same reason we deleted routes on issue #19.

Also, slightly related to this, do you think it would be a good idea to transform all short-ids provided in relationships to URLs at creation/update (maybe at realization)?

### DIFF
--- a/src/ctia/http/routes/observable.clj
+++ b/src/ctia/http/routes/observable.clj
@@ -1,9 +1,10 @@
 (ns ctia.http.routes.observable
   (:require
    [compojure.api.sweet :refer :all]
+   [ctia.lib.pagination :as pag]
    [ctia.http.routes.common :refer [paginated-ok PagingParams]]
    [ctia.schemas
-    [indicator :refer [StoredIndicator]]
+    [common :as c]
     [judgement :refer [StoredJudgement]]
     [sighting :refer [StoredSighting]]
     [vocabularies :refer [ObservableType]]]
@@ -21,16 +22,13 @@
                                :severity
                                :confidence)}))
 
-(s/defschema IndicatorsByObservableQueryParams
-  (st/merge
-   PagingParams
-   {(s/optional-key :sort_by) (s/enum :id :title)}))
+(s/defschema IndicatorRefsByObservableQueryParams
+  (st/dissoc PagingParams :sort_by :sort_order))
 
 (s/defschema SightingsByObservableQueryParams
   (st/merge
    PagingParams
    {(s/optional-key :sort_by) (s/enum :id :timestamp :confidence)}))
-
 
 (defroutes observable-routes
   (GET "/:observable_type/:observable_value/judgements" []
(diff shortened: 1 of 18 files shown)


craigbro:
I DO think we should coerce all references to long URLs.

Maybe we should change the `observable-type/value/indicators` to return a list of indicator URLs, deduplicated?

I think we need to review the use cases there.  The idea is that we DO want to be able to look up what indicators are associated with an observable.  Since Observables are not objects in themselves, we cannot ask for the observable and get a list of "references" and then iterate over them, we have to synthesize that list from the judgements.

polygloton:
My preference was to not have make the realize functions have the responsibility of parsing IDs.  At worst, the realize functions could take a CtiaId instance and call the method that id needs to get the correct ID format.

polygloton:
Returning the list of URLs feels nice.  It seems "RESTful" and also solves the problem of remote entities.

yogsototh:
If we return a list of URLs, we might want to provide a route able to retrieve a list of entities from a list of ids to help the client.

craigbro:
We are not adding CtiaId to the data model, and tsince the realize functions should be taking one of our schmea objects  as their input.....  I don't see how they are going to "take a ctaid"

src/ctia/http/routes/observable.clj
```diff
@@ -51,16 +51,25 @@
     :query [params IndicatorsByObservableQueryParams]
     :path-params [observable_type :- ObservableType
```
gbuisson: I think the previous paging params become meaningless, perhaps you should change them.

gbuisson:
Looks good to me!

src/ctia/http/routes/observable.clj
```diff
@@ -48,19 +46,28 @@
...
-       (list-indicators-by-judgements @indicator-store $ params))))
+     (let [res (as-> {:type observable_type
+                      :value observable_value} $
+                 (list-judgements-by-observable @judgement-store $ nil)
+                 (:data $ [])
+                 (mapcat :indicators $)
+                 (map :indicator_id $)
+                 (set $) ;; remove duplicate URIs
+                 (vec $)
+                 (sort $))]
```
polygloton: I'm not in love with '$ as a variable name :-/

polygloton: Also I think you could just let the $ and move the rest of the sexprs into a the body.

src/ctia/stores/es/indicator.clj
```diff
@@ -19,12 +12,15 @@
 
 (def ^{:private true} mapping "indicator")
 
+(defn long-id->short-id [long-id]
+  (id/short-id (id/long-id->id long-id)))
```
polygloton: Use _ctia.domain.id/str->short-id_ instead, or if you want this as an optimization, add it to ctia.domain.id.

test/ctia/http/generative/specs.clj
```diff
@@ -38,49 +38,49 @@
 (def-property spec-exploit-target-routes 'exploit-target)
 
 (def-property spec-indicator-routes 'indicator)
```
polygloton: Need to delete line 40.

test/ctia/test_helpers/generators/schemas.clj
```diff
@@ -18,24 +19,29 @@
              [sighting-generators :as sg]
              [ttp-generators :as tg]]))
 
+(def ->sighting-long-id
+  (id/long-id-factory :indicator
+                      #(get-in @properties [:ctia :http :show])))
```
polygloton: I think this should be ->indicator-long-id.

polygloton:
LGTM except for some nits.

src/ctia/http/routes/observable.clj
AlexBaranosky: @yogsototh 
I recommend using `->>`. Also, note the two things are unneeded. 
1) don't need the `[]` arg to `(:data $ [])`  (which enables everything to be a `->>`), because it'll default to a nil result which you can `mapcat` over no problem.
2) don't need the `vec` call, which immediately gets turned into a non-vector type when you call `sort`

SEE, for example:
```clj
(class (sort [1 2 3]))
;;=> clojure.lang.ArraySeq
```

MY VERSION:
```clj
(->> (list-judgements-by-observable @judgement-store {:type observable_type
                                                      :value observable_value} nil)
     :data
     (mapcat :indicators)
     (map :indicator_id)
     distinct ;; remove duplicate URIs
     sort)
```

yogsototh: Great! Thanks! I didn't knew `distinc`

Issue #306, opened 2016-05-20
Transient test failure


quoll:
During generative tests we've seen a transient failure on the test-helpers.core/encode function. We've run the test manually, and have been unable to reproduce. Stack trace and failure record below:

``` clojure
{:result #error {
 :cause "Assert failed: (string? s)"
 :via
 [{:type java.lang.AssertionError
   :message "Assert failed: (string? s)"
   :at [ctia.test_helpers.core$encode invoke "core.clj" 284]}]
 :trace
 [[ctia.test_helpers.core$encode invoke "core.clj" 284]
  [ctia.http.generative.specs$fn__8676 invoke "specs.clj" 85]
  [clojure.lang.AFn applyToHelper "AFn.java" 154]
  [clojure.lang.AFn applyTo "AFn.java" 144]
  [clojure.core$apply invoke "core.clj" 630]
  [clojure.test.check.properties$apply_gen$fn__1882$fn__1883 invoke "properties.cljc" 16]
  [clojure.test.check.properties$apply_gen$fn__1882 invoke "properties.cljc" 16]
  [clojure.test.check.rose_tree$fmap invoke "rose_tree.cljc" 78]
  [clojure.test.check.generators$fmap$fn__625 invoke "generators.cljc" 89]
  [clojure.test.check.generators$gen_fmap$fn__603 invoke "generators.cljc" 55]
  [clojure.test.check.generators$call_gen invoke "generators.cljc" 41]
  [clojure.test.check$quick_check doInvoke "check.cljc" 62]
  [clojure.lang.RestFn applyTo "RestFn.java" 142]
  [clojure.core$apply invoke "core.clj" 634]
  [ctia.http.generative.es_store_spec$spec_feedback_routes_es_store doInvoke "es_store_spec.clj" 26]
  [clojure.lang.RestFn applyTo "RestFn.java" 139]
  [clojure.core$apply invoke "core.clj" 632]
  [ctia.http.generative.es_store_spec$spec_feedback_routes_es_store invoke "es_store_spec.clj" 26]
  [ctia.http.generative.es_store_spec$fn__8567 invoke "es_store_spec.clj" 26]
  [clojure.test$test_var$fn__7670 invoke "test.clj" 704]
  [clojure.test$test_var invoke "test.clj" 704]
  [clojure.test$test_vars$fn__7692$fn__7697 invoke "test.clj" 722]
  [clojure.test$default_fixture invoke "test.clj" 674]
[...]

PR #307, opened 2016-05-20
Improvement to test assertions

polygloton:
Trying to better capture the failing input in #306

### DIFF
--- a/test/ctia/test_helpers/core.clj
+++ b/test/ctia/test_helpers/core.clj
@@ -281,11 +281,21 @@
                   time/format-date-time)))
 
 (defn encode [s]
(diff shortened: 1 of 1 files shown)


gbuisson:
LGTM!

test/ctia/test_helpers/core.clj
```diff
@@ -281,11 +281,21 @@
                   time/format-date-time)))
 
 (defn encode [s]
-  {:pre [(string? s) (seq s)]}
+  (assert (string? s)
```
AlexBaranosky: I've always found preconditions to:
1) be a complicated way to basically just do an assert
2) generate not-so-hot error messages.

PR #308, opened 2016-05-20
Refactor capabilities to require all and remove :admin capability

### LINKED ISSUE #108: Refactor the way that auth :capabilities work
I think the implementation for capabilities isn't quite right.  Currently you have to have just one of the capabilities defined for a route to be able to use the route.  Typically there is a user level capability and an admin capability, as in...

:capabilities #{:create-judgement :admin}

I think that :admin should be implicit, and we should be able to require all of the capabilities before granting access.  That way we could protect routes that use multiple stores with just the capabilities to read those stores.  The user would need to have all of the capabilities required by the compound data route.  This way we could reduce the number of permission grants that we give by not having as many :list-thing-A-by-thing-B style permissions.

Need to get more input on this idea.

polygloton:
Refactored auth to require the user to have all the capabilities for a particular route. Admin and allow-all auth mode gets all capabilities rather than the special :admin capability.

Closes #108

### DIFF
--- a/src/ctia/auth.clj
+++ b/src/ctia/auth.clj
@@ -2,9 +2,10 @@
 
 (defprotocol IIdentity
   (authenticated? [this])
+  (admin? [this])
   (login [this])
   (allowed-capabilities [this])
-  (allowed-capability? [this capability]))
+  (capable? [this capabilities]))
 
 (defprotocol IAuth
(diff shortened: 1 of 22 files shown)


saintx:
This seems to be much cleaner.

Issue #313, opened 2016-05-24
Strings get split


nrezvani:
Strings in indicator.source are split on "-". It is suspected that this is an ElasticSearch mapping tokenizer issue. 

### Elasticsearch request body  

```
{
  "query": {
    "query_string": {
      "query": "*",
      "analyze_wildcard": true
    }
  },
  "size": 0,
  "aggs": {
    "2": {
      "terms": {
        "field": "indicators.source",
        "size": 10,
        "order": {
          "_count": "desc"
        }
      },
      "aggs": {
        "3": {
          "date_histogram": {
            "field": "created",
            "interval": "1h",
            "time_zone": "America/Los_Angeles",
            "min_doc_count": 1
          }
        }
      }
    }
  }
}
```

### Kibana screenshot 
![screen shot 2016-05-24 at 12 55 40 pm](https://cloud.githubusercontent.com/assets/16615619/15518129/297b5efc-21b0-11e6-82b6-54387a7652a7.png)

gbuisson:
I think there are multiple issues here:
- Indicator entity currently has no `source` field, I'm not sure if it is because we don't want it or we forgot to add it.
- You use `indicators.source` which will not match the `indicator` entity but references to ìndicators` on other entities, with that query it can match judgement, TTP, Campaign or Sighting, is it what you intend to do? If not I suggest adding a restriction on the `type`field and set it to `indicator`. 

We use the same kind of data structure for all references, which looks like this currently:

```clojure 
(def related
  {:confidence {:type "string"}
   :source {:type "string"}
   :relationship {:type "string"}})
```
- we didn't disable analyzers for those fields, thus we get tokenization. shall we do so?

craigbro:
We should disable analysis on those fields.

We need to track the source of an indicator, the same as Judgement, so if Judgement has source, we should ensure it is there for indicators as well.

And I agree, Nargol may be using the incorrect field.

gbuisson:
Alright, let's merge #314 then and I shall create an issue to add the source field on Indicator.

gbuisson:
@saintx  In order to test that, we need to redeploy CTIA and update the import task to populate the source field on indicator, I also recommend adding a boolean filter to select only "indicator" documents.

Then the ES query shall look something like that:

```json
{
  "query": {
    "query_string": {
      "query": "*",
      "analyze_wildcard": true
    }
  },
  "size": 0,
  "aggs": {
    "2": {
      "terms": {
        "field": "indicator.source",
[...]

PR #314, opened 2016-05-25
disabled es analyzers on relation mapping

gbuisson:
connect #313 
This disables analyzers and tokenization for reference fields.


yogsototh:
LGTM

PR #315, opened 2016-05-25
Issue 305 bulk insert

### LINKED ISSUE #304: Provide get by many ids routes
We should provide some route to which, providing a list of ids returns a list of entities.

See discussion in #302

### LINKED ISSUE #305: Routes to post multiple entities instead of one
For each entity type, we should have a route to add a collection of that kind of entity.  This would reduce HTTP roundtrip overhead and speed up imports.

yogsototh:
I made one commit per route, tell me if you prefer only one big squashed commit.

Also, I'm not fond of the route naming I choose. For now, it is: `/:entity/multi`

I didn't want to call the route `bulk` because the current implementation focus on reducing the number of HTTP request not really on being able to handle big volume of insertions, yet.

Closes #305
Closes #304

### DIFF
--- a/src/ctia/http/routes/bulk.clj
+++ b/src/ctia/http/routes/bulk.clj
@@ -1,0 +1,162 @@
+(ns ctia.http.routes.bulk
+  (:require [compojure.api.sweet :refer :all]
+            [clojure.string :as str]
+            [ctia.flows.crud :as flows]
+            [ctia.domain.entities :as ent]
+            [ctia.schemas
+             [bulk :refer [BulkRefs NewBulk StoredBulk]]]
+            [ctim.schemas.common :as c]
+            [ctia.store :refer :all]
+            [ring.util.http-response :refer :all]
+            [schema.core :as s]
+            [clojure.tools.logging :as log]))
+
+(defn singular [k]
+  "remove the last s of a keyword see test for an example."
+  (-> k
+      name
+      (str/replace #"s$" "")
+      keyword))
+
+(defn realize [k]
+  "return the realize function provided an entity key name"
+  (condp = k
+    :actor          ent/realize-actor
+    :campaign       ent/realize-campaign
+    :coa            ent/realize-coa
+    :exploit-target ent/realize-exploit-target
+    :feedback       ent/realize-feedback
+    :incident       ent/realize-incident
+    :indicator      ent/realize-indicator
+    :judgement      ent/realize-judgement
+    :sighting       ent/realize-sighting
+    :ttp            ent/realize-ttp))
+
+(defn create-fn [k]
+  "return the create function provided an entity key name"
+  (condp = k
+    :actor          #(create-actor @actor-store %)
+    :campaign       #(create-campaign @campaign-store %)
+    :coa            #(create-coa @coa-store %)
+    :exploit-target #(create-exploit-target @exploit-target-store %)
+    :feedback       #(create-feedback @feedback-store %)
+    :incident       #(create-incident @incident-store %)
+    :indicator      #(create-indicator @indicator-store %)
+    :judgement      #(create-judgement @judgement-store %)
+    :sighting       #(create-sighting @sighting-store %)
+    :ttp            #(create-ttp @ttp-store %)))
+
+(defn read-fn [k]
+  "return the create function provided an entity key name"
+  (condp = k
+    :actor          #(read-actor @actor-store %)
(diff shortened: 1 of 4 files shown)


saintx:
One big squashed commit is best in nearly every case.  The reason is to facilitate potential rollback.  It's highly unlikely we will ever need to roll back part of this, but not the rest of it.

WRT naming, in general, I like [/actor & /actors], [/campaign & /campaigns] etc for single & multi, and not the sub-route.  The name of the route should not be chosen based on implementation details (such as whether the implementation is optimized for large batches or small ones), because the user will not and should not need to know or concern themselves with those details.  The *only* thing the user needs to know, is that you use "/actor" for a single entity, and "/actors" for multiple entities, in my opinion.  This keeps it simple, and lets us choose how we'll optimize under the hood later on.

yogsototh:
as we should also provide routes returning a list of entities provided a list of ids, I'm following your proposal to use "/actors"

And I'll squash all my commit together.

gbuisson:
LGTM (Looks Good To Me)

polygloton:
maybe we could have a different top level route, something like /ctia/entities (POST & GET)

polygloton:
Also I think there wold only need to be one route.  We can allow any mix of entity types to be posted if require that the :type filed be set.

polygloton:
We should have a configured maximum number of entities that may be posted.  Its the kind of thing you want to fail when someone is doing their integration testing so that they adjust their client implementation.  I think it should be a property.  This is the kind of thing that could cause production issues down the road if there is no maximum.

saintx:
Agreed.  I found myself contemplating "Why stop at 1k?  Why not 10k?  Why not 100k?  What happens?"

gbuisson:
You might also want to enforce max content size in order to prevent upload bombs.

saintx:
I agree that guarding against upload bombs is good, but at some point, I want us to actually merge #305 and stop tacking new features onto it.  I need this thing.  New features and improvements can go into their own issue.

src/ctia/http/routes/bulk.clj
```diff
@@ -0,0 +1,162 @@
...
+    :actor          ent/realize-actor
+    :campaign       ent/realize-campaign
+    :coa            ent/realize-coa
+    :exploit-target ent/realize-exploit-target
+    :feedback       ent/realize-feedback
+    :incident       ent/realize-incident
+    :indicator      ent/realize-indicator
+    :judgement      ent/realize-judgement
+    :sighting       ent/realize-sighting
+    :ttp            ent/realize-ttp))
```
polygloton: This can go into ctia.domain.entities now.

src/ctia/http/routes/bulk.clj
```diff
@@ -0,0 +1,162 @@
...
+                      :read-campaign
+                      :read-coa
+                      :read-exploit-target
+                      :read-feedback
+                      :read-incident
+                      :read-indicator
+                      :read-judgement
+                      :read-sighting
+                      :read-ttp}
+      (let [bulk (into {} (filter (comp not empty? second)
```
polygloton: you could use "seq" instead of "not empty?" I think

src/ctia/http/routes/bulk.clj
```diff
@@ -0,0 +1,162 @@
...
+            [ring.util.http-response :refer :all]
+            [schema.core :as s]
+            [clojure.tools.logging :as log]))
+
+(defn singular [k]
+  "remove the last s of a keyword see test for an example."
+  (-> k
+      name
+      (str/replace #"s$" "")
+      keyword))
```
polygloton: Maybe move this to a lib?  Not sure what it would be.  Perhaps ctia.lib.keyword.  Also move tests.  :)

polygloton:
Looks good.  I'm merging this.  Minor nits may be addressed in another PR.  :)

src/ctia/http/routes/bulk.clj
AlexBaranosky: I prefer `(remove (comp empty? val))` :)

src/ctia/http/routes/bulk.clj
AlexBaranosky: @yogsototh `case` is both more efficient, and less code.

src/ctia/http/routes/bulk.clj
yogsototh: double Linting 👍 :)

Issue #234, opened 2016-04-18
Use only the URI for ES setup in properties


polygloton:
Currently we have both a URI property, and two properties (one each for host and port), multiplied by two for the store and the producer.  If we could get by with just the URI, then let's use that.

gbuisson:
It is totally doable
- REST uses a single http uri, 
- native uses pairs [[:ip :port]]
- native allows multiple hosts

we could transform the props value to extract the relevant parts for native, It won't be as explicit though.

polygloton:
I can't help thinking we should stop supporting native mode.

PR #320, opened 2016-05-26
Allow per Entity Store configuration

### LINKED ISSUE #175: Allow per store index configuration for ES stores
I whould be able to override the index name used for storing a given entity by that entities ES store.

gbuisson:
closes #175 

This allows a more fine grained Store configuration, you can now set an implementation and its options for each entity.

In order to keep the properties file short, I added a `default` store config namespace, this sets the base configuration for all the stores, you can then declare overrides for each entity.

I also updated the default properties file as well as the one for demo so we set an index by entity.


resources/ctia-default.properties
```diff
@@ -12,25 +12,27 @@ ctia.http.show.protocol=http
...
+ctia.store.campaign.indexname=ctia_campaign
+ctia.store.coa.indexname=ctia_coa
+ctia.store.exploit-target.indexname=ctia_exploit-target
+ctia.store.feedback.indexname=ctia_feedback
+ctia.store.identity.indexname=ctia_entity
+ctia.store.incident.indexname=ctia_incident
+ctia.store.indicator.indexname=ctia_indicator
+ctia.store.judgement.indexname=ctia_judgement
+ctia.store.sighting.indexname=ctia_sighting
+ctia.store.ttp.indexname=ctia_ttp
```
polygloton: Can we split the notion of specifying the store type from setting the store configuration?  I don't really like the _ctia.store.default.type_ property; I prefer the old way.  Instead you could have something like _ctia.store.es.default.uri_, and override that with _ctia.store.es.actor.uri_.  That way it is easy to immediately see what store type is being used without having to scan for a default and then scan again for any overrides.

gbuisson: Thinking about your proposal and the chat with Craig, I think your proposal would be more in line with the possibility of instantiating multiple store instances per entity, is it something we want?

polygloton: That's a interesting point.  We could have properties like...

```
ctia.store.judgement=es,sql
ctia.store.es.default.foo=bar
ctia.store.sql.default.spam=eggs
```

Note the comma separated list.  Seems to work.

polygloton:
I didn't work when I tried it.  The jar file got a different name then expected.

```shell
-> % lein run -m ctia.task.codegen test-codegen
Java HotSpot(TM) 64-Bit Server VM warning: ignoring option MaxPermSize=256m; support was removed in 8.0
Compiling 1 source files to /Users/stevsloa/code/cisco/ctia/target/classes
Java HotSpot(TM) 64-Bit Server VM warning: ignoring option MaxPermSize=256m; support was removed in 8.0
starting CTIA...
Starting nREPL server on port 3001
Starting HTTP server on port 3000
generating clojure client...
Error: Unable to access jarfile /tmp/swagger-codegen.jar

generating java client...
Error: Unable to access jarfile /tmp/swagger-codegen.jar

generating haskell-servant client...
Error: Unable to access jarfile /tmp/swagger-codegen.jar

generating python client...
Error: Unable to access jarfile /tmp/swagger-codegen.jar

generating ruby client...
Error: Unable to access jarfile /tmp/swagger-codegen.jar

done
-> % ls /tmp/swagger-codegen-2.1.6.jar
/tmp/swagger-codegen-2.1.6.jar
```

resources/ctia-default.properties
gbuisson: yes that works, however how do you quickly set all stores to use the same implementation? for instance with my previous work you could configure all stores to use ES with a single line, how could it be done with your proposal?

polygloton: We don't need to quickly set all stores to a single implementation.

There is a default properties file, which has each store type listed on a single line already, so there is only a need for someone to enter a store type if they are overriding the default.  We don't have to optimize for the least number of characters in the properties file.  Nor to we need to optimize for the least amount of typing.  We should optimize for making it easy to mentally parse and verify.  We have cut-and-paste!  I believe that most use cases in production will require multiple types of stores.  

As a system operator, if I can scan the config file and see a property line like "ctia.store.judgement=es", I can quickly determine what store the judgement entity is using.  If I have to scan once for "ctia.store.default.type=es", and then scan again for "ctia.store.judgement.type=sql", its a extra work for the human reader.  We should optimize for readability since humans will read the properties file more often then they will write it.

Its time to stop manicuring CTIA.  It was fine the way it was, and we have better things to do!

gbuisson:
I changed my PR based on feedback, tell me if it looks better.
We'll need to rebuild the CTIA docker demo image to reflect configuration changes.

src/ctia/properties.clj
```diff
@@ -20,13 +22,40 @@
...
+  {(str "ctia.store." store) s/Str})
+
+(defn sql-store-impl-properties [store]
+  {(str "ctia.store.sql." store ".classname") s/Str
+   (str "ctia.store.sql." store ".subprotocol") s/Str
+   (str "ctia.store.sql." store ".subname") s/Str
+   (str "ctia.store.sql." store ".delimiters") s/Str})
+
+(defn es-store-impl-properties [store]
+  {(str "ctia.store.es." store ".uri") s/Str
```
yogsototh: Here we have `uri` and also `port`, `cluster name`, `indexname` and `refresh` while apparently `port` is never used and is part of the `uri` (`http://localhost:9200`).

For me an uri is something that could contain more informations than just host and port, for example: `http://user:password@elasticsearch:9200/index_name` for example.

Couldn't we rename `uri` as `host` and assume the protocol to be `http` and use `port`?

I know this is certainly an older issue not directly related to this PR. So this shouldn't be a blocker.

saintx:
LGTM

src/ctia/properties.clj
saintx: This is probably something we could standardize around, Yann.   Probably shouldn't block this PR, but we could address it in future refactoring and come up with a standard way of doing it moving forward.

gbuisson: there is an issue for that #234

src/ctia/properties.clj
```diff
@@ -20,13 +22,40 @@
...
+   (str "ctia.store.es." store ".port") s/Int
+   (str "ctia.store.es." store ".clustername") s/Str
+   (str "ctia.store.es." store ".indexname") s/Str
+   (str "ctia.store.es." store ".refresh") s/Bool})
+
+(s/defschema StorePropertiesSchema
+  "All entity store properties for every implementation"
+  (let [configurable-stores (map name (keys store/stores))
```
polygloton: The result of this map could just be def'ed in the store ns.

src/ctia/init.clj
```diff
@@ -41,80 +40,76 @@
...
-  (doseq [[store-type store-atom] store/stores]
-    (let [selected-store (get-in @p/properties
-                                 [:ctia :store store-type])
-          builder (get-in store-factories
-                          [:builder selected-store]
-                          (fn default-builder [f] (f)))
+  (doseq [[store-name store-atom] store/stores]
+    (let [store-type (keyword (get-in @p/properties [:ctia :store store-name] "none"))
+          store-properties (merge (get-in @p/properties [:ctia :store store-type :default] {})
+                                  (get-in @p/properties [:ctia :store store-type store-name] {}))
```
polygloton: It is a good idea to move the properties deref to here.  The earlier that we do it, the better for our code.

resources/ctia-default.properties
```diff
@@ -24,13 +27,23 @@ ctia.store.judgement=es
...
+ctia.store.es.campaign.indexname=ctia_campaign
+ctia.store.es.coa.indexname=ctia_coa
+ctia.store.es.exploit-target.indexname=ctia_exploit-target
+ctia.store.es.feedback.indexname=ctia_feedback
+ctia.store.es.identity.indexname=ctia_entity
+ctia.store.es.incident.indexname=ctia_incident
+ctia.store.es.indicator.indexname=ctia_indicator
+ctia.store.es.judgement.indexname=ctia_judgement
+ctia.store.es.sighting.indexname=ctia_sighting
+ctia.store.es.ttp.indexname=ctia_ttp
```
polygloton: Hurray for this!

polygloton:
LGTM

src/ctia/init.clj
AlexBaranosky: @polygloton will your `properties` atom be updatable at runtime?

If so, then this kind of code where you pull different fields off in succession is open to possible race conditions, where the properties are not consistent. It's a small edge case, but thought I'd point it out regardless.

polygloton: It should not be changed after the server starts.  In our tests we start and stop the server, changing the properties each time.  As long as we deref properties during startup, we're good.

AlexBaranosky: Cool then: no change at runtime == no problem.

PR #321, opened 2016-05-26
A task to automatically generate swagger client libraries

### LINKED ISSUE #26: Python client library generated from swagger.json
Once swagger.json is sufficiently stable, we want to generate a python client library for the CIA, and make it available for users.

It should include some basic use case examples in it's documentation.

gbuisson:
closes #26 

This automatically generates CTIA swagger client libraries for `["clojure" "java" "haskell-servant" "python" "ruby"]` using swagger-codegen.

invoke it with `lein run -m ctia.task.codegen <output-dir>`


src/ctia/task/codegen.clj
```diff
@@ -0,0 +1,56 @@
...
+  (let [port (get-in @properties [:ctia :http :port] 3000)]
+    (str "http://localhost:" port "/swagger.json")))
+
+(defn exec-command [& args]
+  (when-let [err (:err (apply shell/sh args))]
+    (println err)))
+
+(defn setup []
+  (println "starting CTIA...")
+  (start-ctia!)
```
polygloton: Might want to specify ":join? false, :silent? true" incase the defaults ever change.

polygloton:
I didn't work when I tried it.  The jar file got a different name then expected.

```shell
-> % lein run -m ctia.task.codegen test-codegen
Java HotSpot(TM) 64-Bit Server VM warning: ignoring option MaxPermSize=256m; support was removed in 8.0
Compiling 1 source files to /Users/stevsloa/code/cisco/ctia/target/classes
Java HotSpot(TM) 64-Bit Server VM warning: ignoring option MaxPermSize=256m; support was removed in 8.0
starting CTIA...
Starting nREPL server on port 3001
Starting HTTP server on port 3000
generating clojure client...
Error: Unable to access jarfile /tmp/swagger-codegen.jar

generating java client...
Error: Unable to access jarfile /tmp/swagger-codegen.jar

generating haskell-servant client...
Error: Unable to access jarfile /tmp/swagger-codegen.jar

generating python client...
Error: Unable to access jarfile /tmp/swagger-codegen.jar

generating ruby client...
Error: Unable to access jarfile /tmp/swagger-codegen.jar

done
-> % ls /tmp/swagger-codegen-2.1.6.jar
/tmp/swagger-codegen-2.1.6.jar
```

src/ctia/task/codegen.clj
AlexBaranosky: @polygloton I'm loving that attention to detail ^^^

yogsototh:
tried it locally, seems to work. It would be great if we could add more metadata information to provide the right authors information for example, the right project page, etc...

gbuisson:
I added support for additional properties by lang.

Seems that each template has its own vars for defining package name, description and such.
for now it works well for clojure, python and ruby are okay.
Java doesn't seem to compile because it fails to parse the spec for NewFeedback.

gbuisson:
I think you may review it again

yogsototh:
There were a problem regarding names, Haskell forbid "-" in names, I used a capitalized version.
I've fixed this particular problem, I made a PR on your PR.

Unfortunately, the RelationInfo is problematic, as `s/Any` isn't authorized by Haskell. So it won't compile up until we'll make RelationInfo more constrained.

polygloton:
I'm still getting an error when I try to run in.  Its long, so I put it in a gist.

https://gist.github.com/polygloton/9004792a994d3ab87a27c5aefaea6d30

The code looks good.  If its just me, let's go ahead and merge it.

gbuisson:
Yeah I can't get the java client to compile, there must be something on the schemas that is incompatible, maybe we shall disable it.

saintx:
The java client is a good candidate to split into its own issue.

PR #322, opened 2016-05-26
Added :version after realization for all entities

### LINKED ISSUE #318: Populate document :version on realize
In order to ease data migrations, It would be useful if each document would contain its schema version.
We could populate the :version field at realize time.

yogsototh:
Closes #318


yogsototh:
After a discussion with @saintx we decided to add some test that may fail if any major schema changes in order to be certain not to forget to upgrade the version.

We might imagine a better approach in which we could detect the changes and depending on it we could advise the use that this will be a major upgrade (modifying the type of a field, removing a field or adding a mandatory field) or a minor upgrade (adding an optional field).

But before that we would like to make this proposal. Any better approach or idea would be more than welcome.

PR #327, opened 2016-05-31
Refactor to use CTIM

### LINKED ISSUE #324: Use CTIM for schema definitions
Rather than defining schemas in CTIA.

polygloton:
Closes #324


src/ctia/stores/sql/judgement.clj
```diff
@@ -24,11 +24,11 @@
...
-(defn realize-judgements [login new-judgements]
-  (for [new-judgement new-judgements]
-    (judgement-schema/realize-judgement new-judgement
-                                        (str "judgement-" (UUID/randomUUID))
-                                        login)))
+;; (defn realize-judgements [login new-judgements]
+;;   (for [new-judgement new-judgements]
+;;     (realize-judgement new-judgement
+;;                        (str "judgement-" (UUID/randomUUID))
+;;                        login)))
```
polygloton: I wasn't sure if this was being used so I commented it out.  I could probably delete it.

yogsototh: I believe we should remove the commented code here and also the first commented require.

src/ctia/stores/sql/judgement.clj
```diff
@@ -24,11 +24,11 @@
               (k/table :judgement_indicator)
               (k/database @db))))
 
-(defn realize-judgements [login new-judgements]
-  (for [new-judgement new-judgements]
-    (judgement-schema/realize-judgement new-judgement
-                                        (str "judgement-" (UUID/randomUUID))
-                                        login)))
+;; (defn realize-judgements [login new-judgements]
```
saintx: Does this still need to be here?

polygloton: Probably not.

yogsototh:
That's great! Look goods to me except my minor comment.

I just thought about a potential problem. For example, when we will correct the #7 on CTIM, and we will push to clojars using the same SNAPSHOT version number, their might be a problem for people having not yet merged the fix on CTIA in their own branch.

I believe we would only need to use a `lein install` in a local CTIM fixed at an older release to fix this problem locally.

polygloton:
Got a test failure

```clojure
{:end_time "2016-07-05T21:17:42.012Z"}} ({:description "Z))KM9Du>6", :id "sighting-n+", :timestamp #inst "2016-05-03T21:17:42.021-00:00", :indicators [{:indicator_id "http://localhost:56209/ctia/indicator/indicator-%23"}]})]], :shrunk {:total-nodes-visited 11, :depth 4, :result false, :smallest [[{:description "wz?`V", :producer "(\\O7)9x2", :title "b", :id "indicator-!", :valid_time {:end_time "2016-05-31T21:17:42.112Z"}} ({:description "X::$)", :id "sighting-!", :timestamp #inst "2016-08-09T21:17:42.341-00:00", :indicators [{:indicator_id "http://localhost:56209/ctia/indicator/indicator-%21"}]})]]}, :test-var "spec-indicator-routes-atom-store"}

lein test :only ctia.http.generative.atom-store-spec/spec-indicator-routes-atom-store

FAIL in (spec-indicator-routes-atom-store) (clojure_test.cljc:21)
expected: result
  actual: false
```

polygloton:
Will deal with the test failure as a separate issue.

gbuisson:
LGTM

PR #333, opened 2016-06-01
Add CLI to feed importer

### LINKED ISSUE #323: Add CLI to TG Feed Importer
Add a CLI to the TG Feed import mechanism, so we can use it on the alpha.ctia demo server without an NREPL connection

saintx:
Adds command line interface for the threatgrid feed import tools,
allowing us to run the importer from the alpha.ctia server without
access to the NREPL.  Also adds support for relations to the sightings.

Closes #323

### DIFF
--- a/src/ctia/import/threatgrid/feed.clj
+++ b/src/ctia/import/threatgrid/feed.clj
@@ -1,16 +1,16 @@
 (ns ctia.import.threatgrid.feed
   (:require [cheshire.core :as cjson]
-            [clojure.data.json :as json]
             [clojure.java.io :as io]
+            [clojure.data.json :as json]
             [clojure.string :as str]
-            [clj-time.core :refer [date-time days]]
-            [clj-time.format :refer [unparse formatter]]
-            [ctia.lib.time :refer [date-range] :as ctime]
+            [clj-time.core :refer [date-time days now] :as time]
+            [clj-time.format :refer [parse unparse formatter] :as tf]
+            [ctia.lib.time :refer [date-range after?] :as ctime]
+            [ctia.import.threatgrid.http :as http]
             [ctia.import.threatgrid.feed.indicators :as fi]
             [ctia.import.threatgrid.feed.sightings :as fs]
             [ctia.import.threatgrid.feed.judgements :as fj]
-            [clj-http.client :as http]
-            [clj-http.conn-mgr :as conn]))
+            [clojure.tools.cli :refer [parse-opts]]))
 
 (defn tg-feed-names []
   [;;"autorun-registry" ;; always empty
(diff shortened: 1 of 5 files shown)


src/ctia/import/threatgrid/feed.clj
```diff
@@ -28,102 +28,168 @@
...
-                  :observable {:type "domain" :field :domain}
-                  :entries entries}]
-        (let [judgements (fj/feed->judgements feed)]
-          (println "Posting" (count judgements) "judgements to ctia")
-          (post-to-ctia judgements "judgement" ctia-uri conn-mgr))
-        (let [sightings (fs/feed->sightings feed)]
-          (println "Posting" (count sightings) "sightings to ctia")
-          (post-to-ctia sightings "sighting" ctia-uri conn-mgr)))
-      (println "Done loading " file-name))))
+(defn usage [options-summary]
```
AlexBaranosky: Multiline strings make this a little less noisy, imho:

```clj
(defn usage [options-summary]
  (str "Import Threat Grid feeds into CTIA.

Usage: lein run -m ctia.import.threatgrid.feed [options]

Options:
"
      options-summary
      "
Valid Feed Names:
  banking-dns
  dll-hijacking-dns
  doc-net-com-dns
  downloaded-pe-dns
  dynamic-dns
  irc-dns
  modified-hosts-dns
  parked-dns
  public-ip-check-dns
  rat-dns
  sinkholed-ip-dns
  stolen-cert-dns"))
```

src/ctia/import/threatgrid/feed/indicators.clj
```diff
@@ -13,7 +13,7 @@
    :conn-timeout 10000})
 
 (defn stored-ctia-indicator [title ctia-uri]
-  (let [target-url (str ctia-uri "/ctia/indicator/title/" title)
+  (let [target-url (str ctia-uri "/ctia/indicator/title/" "tg-feed-" title)
```
AlexBaranosky: @saintx you can just combine these strs to:

```clj
(str ctia-uri "/ctia/indicator/title/tg-feed-" title)
```

src/ctia/import/threatgrid/feed/sightings.clj
```diff
@@ -7,76 +7,53 @@
...
-
-(defn entries->sightings
-  [entries observable-type observable-field
-   indicator-id indicator-source source-uri & {:as options}]
-  (apply transform entries observable-type observable-field
-         indicator-id indicator-source source-uri options))
+  [feed & {:keys [tlp confidence]
+           :or {tlp "green"
+                confidence "High"}}]
+  (let [{:keys [title description source source_uri
```
AlexBaranosky: @saintx in case you didn't know this is valid Clojure:

```clj
{title :title
 description :description
 indicator-source :source
 source-uri :source_uri
 observable :observable
 indicator :indicator
 entries :entries} feed
```

This means you wouldn't need to have the `let` lines for `indicator-source source` and `source-uri source_uri`

You can actually do this, and you wouldn't need the `let` at all:

```clj
(defn transform 
  [{title :title
            description :description
            indicator-source :source
            source-uri :source_uri
            {observable-type :type
             observable-field :field
             :as observable} :observable
            {indicator-id :id
             :as indicator} :indicator
            entries :entries
            :as feed}
 & {:keys [tlp confidence]
           :or {tlp "green"
                confidence "High"}}]

...
)
```

src/ctia/import/threatgrid/http.clj
```diff
@@ -0,0 +1,66 @@
...
+
+(defn retry
+  [retries f & args]
+  (let [res (try {:value (apply f args)}
+                 (catch Exception e
+                   (if (= 0 retries)
+                     (throw e)
+                     {:exception e})))]
+    (if (:exception res)
+      (do
```
AlexBaranosky: @saintx this `do` can be removed. It's not doing anything here.

src/ctia/import/threatgrid/http.clj
```diff
@@ -0,0 +1,66 @@
...
+      (Thread/sleep 1000) ;; a courtesy
+      (let [body (:body (http/get (tg-uri feed-file-name) options))]
+        (if (not (body-empty? body))
+          (let [file-path (str feed-path "/" feed-file-name)]
+            (println "Downloading" file-path)
+            (with-open [wrtr (io/writer file-path)]
+              (.write wrtr body)))
+          (println feed-file-name "is empty."))))))
+
+(defn retry
```
AlexBaranosky: Seems like it could go in a utility ns somewhere.

src/ctia/import/threatgrid/http.clj
saintx: I think I had some tracing in there at one point, which called for the do, and forgot to remove it when I removed the trace.

src/ctia/import/threatgrid/feed/sightings.clj
saintx: This is exactly what I needed--I had tried destructuring it initially, but all the tutorials and notes that I could find on it were for let blocks, and I ended up having an uneven number of symbols in the form.   In particular, in the example  `{foo :foo bar :bar :as baz} :baz`, the trailing `:baz` is what I was missing--this is a very helpful comment, thank you.

src/ctia/import/threatgrid/http.clj
saintx: I agree in principle, but don't think that should be a blocking issue for this PR.

polygloton:
LGTM

PR #334, opened 2016-06-02
Support multiple store instance for each entity

gbuisson:
This PR is following #175 

I believe this feature to be optional as we don't need it immediately, however I have it working and pass all tests so I propose it.

Also having tests using all stores at once helped solve small inconsistencies between all implementations.

It allows instantiating multiple store backends per entity:
let's say you configure `ctia.store.judgement=es,sql,atom`, this would spawn 3 stores: Elasticsearch, SQL and atom.
Writes would be replicated to all stores, Reads will be randomly done on one of them.

In order to achieve that, I added 2 proxy methods: `read-store` and `write-store`, these wrap all store access on routes/crud and may be tweaked to better distribute load or protect the system in case of a component failure.

The downside of my approach is that I use a lot of closures, there's a lot of room for simplification/improvement, PRs are welcome!


src/ctia/store.clj
```diff
@@ -85,40 +85,21 @@
...
+                       :sighting []
+                       :incident []
+                       ;;:relation relation-store
+                       :identity []}))
+
+(defn write-store [e write-fn]
+  (first (doall (map write-fn (e @stores)))))
+
+(defn read-store [e read-fn]
+  (read-fn (rand-nth (e @stores))))
```
yogsototh: I would rather prefer `first` as default instead of `rand-nth` simply because statistically rand-nth will tend to go a lot toward the worst result:

If queries from store1 takes 2ms and queries from store2 takes 100ms, the average query-time will be 51ms.

I would propose to have a configurable via properties strategy function: there should be at least `first` and `rand-nth`.

yogsototh: I would use:

~~~clojure
(defn read-store [store write-fn & args]
    (first (doall (map (apply write-fn (e @stores) args))))))
~~~

Which could replace

~~~clojure
(read-store :sighting (fn [s] (read-sighting s id)))
~~~

by

~~~clojure
(read-store :sighting read-sighting id)
~~~

The same remarks can be done for the `write-store` function.

Also as the `read-store` and `write-store` read and write to many store, I would have put a plural like: `read-stores`, `write-stores` or `read-from-stores`, `write-to-stores`.

yogsototh:
As a general remark, this PR also modify a lot of different files in a similar way. I've seen this pattern in lot of other PR.

It might be the occasion to adopt a CRUD abstraction to be able make the same kind of modification faster. Typically having functions: `(flow/create :actor actor)` instead of the actual more complex `(flow/create-flow :type :actor .... )` and maybe having a macro which provide standard CRUD routes for some entity.

polygloton:
I think @yann might be right.  I've been adding logic to the flow/crud NS that might be better off in a layer above it (like ID format validation).  However, we should wait until someone is experiencing enough pain to motivate them to write that layer.  Don't be proactive on it.  :)

gbuisson:
I updated the PR with @yogsototh suggestions

yogsototh:
LGTM

src/ctia/init.clj
```diff
@@ -96,20 +96,29 @@
...
+                                  {:properties (merge (get-in @p/properties [:ctia :store impl :default] {})
+                                                      (get-in @p/properties [:ctia :store impl store-key] {}))
+
+                                   :builder (get-in store-factories [:builder impl]
+                                                    (fn default-builder [f p] (f)))
+
+                                   :factory (get-in store-factories [store-key impl]
+                                                    #(throw (ex-info (format "Could not configure %s store" impl)
+                                                                     {:store-key store-key
+                                                                      :store-type (keyword %)})))}) store-impl-keys)
```
polygloton: Try not to leave "dangling" parameters, like 'store-impl-keys here.  They are hard to find, and can be easily missed by the reader.  Whenever you have multi-line expressions like this, close it on one line and start others on the next line.

src/ctia/store.clj
```diff
@@ -85,40 +85,23 @@
...
+                       :sighting []
+                       :incident []
+                       ;;:relation relation-store
+                       :identity []}))
+
+(defn write-store [store write-fn & args]
+  (first (doall (map #(apply write-fn % args) (store @stores)))))
+
+(defn read-store [store read-fn & args]
+  (apply read-fn (rand-nth (store @stores)) args))
```
polygloton: I really think this should be (first (store ...)).  We shouldn't be trying to do load balancing at this point.  The user should be able to deterministically specify the fastest store to be used in reads, and the easiest way to do that is to put it first in the list of stores.

test/ctia/test_helpers/es.clj
```diff
@@ -8,18 +8,19 @@
 
 (defn recreate-state-index [state]
   (when (:conn state)
-    (es-index/delete! (:conn state)
-                      (:index state))
+    (do (es-index/delete! (:conn state)
```
polygloton: To 'do here is redundant (when has an implicit 'do).

polygloton:
LGTM, except that I would like to see stores used during reads to be the first store.

gbuisson:
Your wishes are granted!

PR #336, opened 2016-06-02
Refactoring IDs so that only <type>-<UUID> is valid

### LINKED ISSUE #311: IDs follow a required format
- [x] IDs must be the entity type followed by a UUID.  Use a regex to validate user submitted IDs.
- [ ] Refactor ctia.domain.id.  Since we won't allow arbitrary IDs, it can be simplified.
  - [x] Transformation from short-id -> long-id
  - [x] Transformation from long-id -> short-id
  - [ ] Remove ID protocol
- [x] Add tests...
  - [x] Assert that only admins may set an ID
  - [x] Assert that the submitted IDs match the pattern
- [ ] Long IDs are made from short IDs in the realize functions?

polygloton:
Enforces the ID format
Not directly using CtiaId records (except in tests)
Closes #311

### DIFF
--- a/src/ctia/domain/id.clj
+++ b/src/ctia/domain/id.clj
@@ -3,7 +3,7 @@
   (:import java.lang.String))
 
 (def short-id-re
-  (re-pattern (str "(" url/url-chars-re ")")))
+  #"([a-z][-a-z]+-[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12})")
 
 (def url-re
   #"(https?):\/\/([-\da-zA-Z][-\da-zA-Z.]*)(:(\d+))?((\/[-\w.]+)*)\/ctia\/([a-z]+)\/")
@@ -36,6 +36,9 @@
 (def short-id?
   "Make an educated guess if this is a short-form ID"
   (complement long-id?))
+
+(defn valid-short-id? [short-id]
+  (boolean (re-matches short-id-re short-id)))
 
 (defprotocol ID
   (localize [this url-params]
@@ -70,13 +73,14 @@
 
 (defn short-id->id
   [type short-id {:keys [hostname path-prefix port protocol]}]
-  (map->CtiaId
-   {:hostname hostname
-    :short-id short-id
-    :path-prefix (not-empty path-prefix)
-    :port port
-    :protocol protocol
-    :type (name type)}))
+  (if (valid-short-id? short-id)
+    (map->CtiaId
+     {:hostname hostname
+      :short-id short-id
+      :path-prefix (not-empty path-prefix)
+      :port port
+      :protocol protocol
+      :type (name type)})))
 
 (defn ->id
   "Given a string ID, build an ID instance with provided URL
@@ -98,10 +102,21 @@
     (last (re-matches long-id-re s))
     s))
 
-(defn long-id-factory
+(defn factory:short-id->long-id
   "Build a fn that takes a short-id and returns a long-id"
   [type url-params-fn]
   (fn [short-id]
+    (long-id
+     (short-id->id type
+                   short-id
+                   (url-params-fn)))))
+
+(def ^:deprecated long-id-factory factory:short-id->long-id)
+
+(defn factory:short-id+type->long-id
+  "Build a fn that takes an id-type and a short-id and returns a long-id"
+  [url-params-fn]
+  (fn [type short-id]
     (long-id
      (short-id->id type
                    short-id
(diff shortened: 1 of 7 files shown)


src/ctia/domain/id.clj
```diff
@@ -98,11 +102,22 @@
...
+(defn factory:short-id->long-id
   "Build a fn that takes a short-id and returns a long-id"
   [type url-params-fn]
   (fn [short-id]
     (long-id
      (short-id->id type
                    short-id
                    (url-params-fn)))))
+
+(def ^:deprecated long-id-factory factory:short-id->long-id)
```
saintx: Out of curiosity, new code that's already deprecated, is this intentional?

saintx:
The regex looks right to me, as does the rest of it, with the sole exception of my question about that deprecated def.

src/ctia/domain/id.clj
polygloton: It is an alias to the old fn name.

src/ctia/domain/id.clj
```diff
@@ -70,13 +73,14 @@
...
 (defn short-id->id
   [type short-id {:keys [hostname path-prefix port protocol]}]
-  (map->CtiaId
-   {:hostname hostname
-    :short-id short-id
-    :path-prefix (not-empty path-prefix)
-    :port port
-    :protocol protocol
-    :type (name type)}))
+  (if (valid-short-id? short-id)
```
yogsototh: We might prefer to use `when` instead of `if`.

yogsototh:
LGTM!

src/ctia/domain/id.clj
AlexBaranosky: @yogsototh it'd be interesting to hear why you might prefer `when` over `if` in this spot? This is one of my top five bikeshedding discussion subjects :D /cc @polygloton

src/ctia/flows/crud.clj
```diff
@@ -7,20 +7,35 @@
...
   (str (name entity-type) "-" (UUID/randomUUID)))
 
+(defn- find-entity-id [identity entity-type entity prev-entity]
+  (or (find-id prev-entity)
+      (when-let [entity-id (find-id entity)]
+        (when-not (auth/capable? identity :specify-id)
+          (throw (http-response/bad-request!
+                  {:error "Missing capability to specify entity ID"
+                   :entity entity})))
+        (if (id/valid-short-id? entity-id)
```
AlexBaranosky: @polygloton seems more consistent to use `when-not` twice here, instead of the combo of `when-not` and `if`

```clj
(defn- find-entity-id [identity entiaty-type entity prev-entity]
  (or (find-id prev-entity)
      (when-let [entity-id (find-id entity)]
        (when-not (auth/capable? identity :specify-id)
          (throw (http-response/bad-request!
                  {:error "Missing capability to specify entity ID"
                   :entity entity})))
        (when-not (id/valid-short-id? entity-id)
          (throw (http-response/bad-request!
                  {:error (format "Invalid entity ID: %s" entity-id)
                   :entity entity})))
        entity-id)
      (make-id entity-type)))
```

src/ctia/domain/id.clj
AlexBaranosky: ... I've heard arguments on both sides of the issue. By this point in tie, I figure the debate is so 50-50 that people can just use whichever they like best.

yogsototh: I believe the first reason is because I also code in Haskell I believe. And in Haskell you are forced to provide both values which are also statically typed to have the same type.
If you want to make a one way it is more explicit. So in my habits, when I see an `if` I tend to search the else value while when I see `when` it is easier for me to read.

Also, if you treat your `if` as a `when`, adding another effectful line (like a println) won't change the meaning. But this is mainly a matter of taste.

AlexBaranosky: I used to use `when` exclusively in spots like this. Then one time I contributed a PR to one of Technomancy's projects and he shared his opinion which was that you should avoid `when` because it's implicit `do` implies side effects. (and side effects are something to be avoided)

I personally, don't generally concern myself with it anymore :) Simply because there are valid arguments to both sides and it mostly is a very small matter.

polygloton: I confess I normally use when in this case (there is no alternate branch).  But reality shows I probably use whatever pops out of my brain first, as I did in this case.

src/ctia/flows/crud.clj
polygloton: I thought about it, and I agree, it looks nicer.  I ended up not doing it because I was afraid it makes the entity-id value harder to spot when scanning the code.  Your suggestion is probably better though.

AlexBaranosky: @polygloton I pretty much always use a `when` when throwing exceptions.

AlexBaranosky: .. for what it's worth...

PR #342, opened 2016-06-04
Refactor to use ID code and schema generator code from CTIM

### LINKED ISSUE #340: Move all entity ID code to CTIM
The generator code is there already (out of date now).  Need to move ctia.domain.id to CTIM.  Not sure exactly what it should be called.  Get rid of ID generators in CTIA (after updating CTIM equivalent code).

### LINKED ISSUE #341: Remove schema generators from CTIA
They are duplicated in CTIM.  We should have it only in CTIM.

For example, we should remove ctia.test-helpers.generators.schemas and its children namespaces.

polygloton:
Closes #340 and Closes #341


gbuisson:
LGTM

PR #344, opened 2016-06-08
Log configuration on launch

### LINKED ISSUE #316: Output configuration on launch
In order to ease ops and debugging, at CTIA launch we should output:

- [ ] Configuration sources that are detected/used
- [ ] The properties map with its configured values
- [ ] CTIA release version

gbuisson:
closes #316 

- configured console log appender
- added a properties option to enable event file logging
- tweaked logging format to display time

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -6,6 +6,7 @@
 
   :jvm-opts [ "-Xmx4g" ;; On some OSX VMs, this is needed to increase available memory
              "-Djava.awt.headless=true"
+             "-Dlog.console.threshold=INFO"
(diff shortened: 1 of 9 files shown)


yogsototh:
I like the second commit!

Just to note, it might just relate to #316 and not close it.

I believe it is much more work to tell what is the source for each line of configuration. So we should merge this PR, but not close #316, as I believe it would be really helpful to know where some configuration come from.

So LGTM if you replace `Closes` by `Relates to` in your PR comment.

gbuisson:
I added a logging step that logs properties by source on debug level

PR #348, opened 2016-06-09
Add a limit to bulk size.

### LINKED ISSUE #326: Limit the size in number of elements for the bulk route
The limit should be fixed with a property (see discussion here: https://github.com/threatgrid/ctia/pull/315)

yogsototh:
Closes #326 

I used a default limit of 2000 elements. @saintx does 2000 seems correct to you?

### DIFF
--- a/src/ctia/http/routes/bulk.clj
+++ b/src/ctia/http/routes/bulk.clj
@@ -5,6 +5,7 @@
             [ctia.flows.crud :as flows]
             [ctia.lib.keyword :refer [singular]]
             [ctia.schemas.bulk :refer [BulkRefs NewBulk StoredBulk]]
+            [ctia.properties :refer [properties]]
             [ctia.store :refer :all]
             [ctim.schemas.common :as c]
             [ring.util.http-response :refer :all]
@@ -99,6 +100,14 @@
           {}
           (keys bulk)))
 
+(defn bulk-size [bulk]
+  (reduce + 0 (map count (vals bulk))))
+
+(defn get-bulk-max-size []
+  (get-in @properties
+          [:ctia :http :bulk :max-size]
+          2000))
+
 (defroutes bulk-routes
   (context "/bulk" []
     :tags ["Bulk"]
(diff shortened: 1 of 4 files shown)


gbuisson:
LGTM

saintx:
Yes, 2k is a fine default for the time being.

src/ctia/http/routes/bulk.clj
```diff
@@ -118,7 +127,9 @@
                       :create-sighting
                       :create-ttp}
       :identity login
-      (ok (gen-bulk-from-fn create-entities bulk login)))
+      (if (> (bulk-size bulk) (get-bulk-max-size))
+        (forbidden (str "Bulk max nb of entities: " (get-bulk-max-size)))
```
polygloton: Should use 'bad-request instead of 'forbidden

src/ctia/http/routes/bulk.clj
```diff
@@ -99,6 +100,14 @@
           {}
           (keys bulk)))
 
+(defn bulk-size [bulk]
+  (reduce + 0 (map count (vals bulk))))
```
polygloton: I don't know if its worth changing, but you could also do...

```clojure
(apply + (map count (vals bulk)))
```

test/ctia/http/routes/bulk_test.clj
```diff
@@ -188,3 +186,55 @@
...
+                  :ttps (map mk-new-ttp (range nb))}]
+    (is (= (bulk-size new-bulk)
+           (* nb 10)))))
+
+(deftest-for-each-store bulk-max-size-post-test
+  (helpers/set-capabilities! "foouser" "user" all-capabilities)
+  (whoami-helpers/set-whoami-response "45c1f5e3f05d0" "foouser" "user")
+
+  (helpers/with-properties ["ctia.http.bulk.max-size" 50]
+    #(let [nb 5
```
polygloton: The 'with-properties macro expects a body rather than an fn, so this test doesn't work.  (It adds syntactic sugar around 'with-properties-, which does take an fn).

polygloton:
Looks good except for a couple small issues.

test/ctia/http/routes/bulk_test.clj
AlexBaranosky: @polygloton Steve, you have the eagle eye!

PR #358, opened 2016-06-13
some cleaning via my fork of lein-kibit

yogsototh:
to note `(not (empty? x))` IS NOT equivalent to `(seq x)`. So I removed that rule from lein kibit.

### DIFF
--- a/src/ctia/stores/atom/common.clj
+++ b/src/ctia/stores/atom/common.clj
@@ -82,15 +82,14 @@
      filter-map :- {s/Any s/Any}]
 
     (when-not (empty? filter-map)
-      (into []
-            (filter (fn [model]
-                      (every? (fn [[k v]]
-                                (let [found-v (if (sequential? k)
-                                                (get-in model k ::not-found)
-                                                (get model k ::not-found))]
-                                  (match? found-v v)))
-                              filter-map))
-                    (vals (deref state)))))))
+      (vec (filter (fn [model]
+                     (every? (fn [[k v]]
+                               (let [found-v (if (sequential? k)
+                                               (get-in model k ::not-found)
+                                               (get model k ::not-found))]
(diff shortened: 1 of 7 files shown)


src/ctia/stores/atom/common.clj
```diff
@@ -82,15 +82,14 @@
...
-      (into []
-            (filter (fn [model]
-                      (every? (fn [[k v]]
-                                (let [found-v (if (sequential? k)
-                                                (get-in model k ::not-found)
-                                                (get model k ::not-found))]
-                                  (match? found-v v)))
-                              filter-map))
-                    (vals (deref state)))))))
+      (vec (filter (fn [model]
```
AlexBaranosky: @polygloton `filterv` is better, imho

yogsototh: Nice one, I forked kibit here and I added the `(vec (filter` to `(filterv` rule.

https://github.com/yogsototh/kibit

Also there is a task about verifying how much code don't have any docstring and lein bikeshed provide such information.

it might be a nice idea to put both kibit and bikeshed in the CI to fail the test if some conditions aren't made. It would help us encode some common coding style rules.

polygloton:
I wouldn't go so far as to fail a test.  We shouldn't be forced to comply with a lint checker.  :)  Also, your changes to kibit sound solid.  You should send them a PR.

yogsototh:
Yes I've done some PR. Yes you're right, failing test is too much, but I liked the idea to somehow write some of the coding rules into an automated tool.

polygloton:
Maybe print out a warning when we run tests locally

yogsototh:
How about a `prepush` lein alias that will do a `lein do clean, bikeshed, kibit, test`?

polygloton:
that is a good plan

PR #362, opened 2016-06-14
Prepush

yogsototh:
The action is not a chained do as any message from bikeshed or kibit
	is considered as a fail. So I used a shell script for that.
	Also I consider to warn only on >100 columns lines as there are too
	much >80 column for now.

also replace `vec filter` by `filterv`


gbuisson:
Looks good!

PR #365, opened 2016-06-15
Add error logs on parsing and validation errors.

### LINKED ISSUE #364: Capture jetty standard out/error
Either log it (preferably) or print it.  We must have lost this feature when we stopped using lein-ring.

One way to do this would be to add a custom middleware that catches exceptions and logs them before re-throwing them.

yogsototh:
See: https://github.com/metosin/compojure-api/wiki/Exception-handling
	and: https://github.com/metosin/compojure-api/blob/master/src/compojure/api/exception.clj

Should close #364

### DIFF
--- a/src/ctia/http/exceptions.clj
+++ b/src/ctia/http/exceptions.clj
@@ -1,28 +1,40 @@
 (ns ctia.http.exceptions
   "This ns declare all handler for server exceptions.
 
-  See https://github.com/metosin/compojure-api/wiki/Exception-handling"
+  See <https://github.com/metosin/compojure-api/wiki/Exception-handling>"
   (:require
    [compojure.api.impl.logging :as logging]
    [compojure.api.exception :as ex]
-   [ring.util.http-status :refer [internal-server-error]]))
+   [ring.util.http-status :refer [internal-server-error]])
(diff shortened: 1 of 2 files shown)


polygloton:
I think you should override the default as well, so that we can be explicit about it.

polygloton:
Also check for ExceptionInfo and included the ex-data in the log message if there is any.

src/ctia/http/exceptions.clj
```diff
@@ -1,28 +1,40 @@
...
+  (logging/log! :error e (ex-message e))
   (ex/response-validation-handler e data request))
 
-
+(defn default-error-handler
+  "Handle default error"
+  [^Exception e data request]
+  (logging/log! :error e (ex-message e))
+  (internal-server-error {:type "unknown-exception"
+                          :class (.getName (.getClass e))}))
```
quoll: It's usually better to use clojure.core functions when possible. So:
`   ... :class (.getName (class e)) ...`

yogsototh: I've copy/pasted from the metosin code as the goal was to keep mostly the same function but only change its logs. But yes, I agree, `class` is better.

PR #366, opened 2016-06-15
get rid of the nested documents on sighting

gbuisson:
closes https://github.com/threatgrid/ctim/issues/21

This enables us to get rid of ES nested objects for sighting observables by computing an observable hashes list just before ES document storage and clearing it at retrieval.

I would have used something like https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-transform.html
but it is now deprecated so I do it from CTIA.

### DIFF
--- a/src/ctia/stores/es/sighting.clj
+++ b/src/ctia/stores/es/sighting.clj
@@ -8,14 +8,25 @@
             [schema.core :as s]))
 
 (s/defschema ESStoredSighting
-  (st/merge StoredSighting
-            {:observables_hash [s/Str]}))
+  (st/assoc StoredSighting :observables_hash [s/Str]))
 
+(def ESStoredSightingList (list-response-schema ESStoredSighting))
+(def StoredSightingList (list-response-schema StoredSighting))
 (def es-coerce! (crud/coerce-to-fn [(s/maybe ESStoredSighting)]))
+(def create-fn (crud/handle-create :sighting ESStoredSighting))
+(def read-fn (crud/handle-read :sighting ESStoredSighting))
+(def update-fn (crud/handle-update :sighting ESStoredSighting))
+(def list-fn (crud/handle-find :sighting ESStoredSighting))
 
 (s/defn observable->observable-hash :- s/Str
-  [{:keys [type value] :as obsrvable :- Observable}]
+  "transform an observable to a hash"
+  [{:keys [type value] :as o :- Observable}]
   (str type ":" value))
+
+(s/defn obs->hashes :- [s/Str]
+  "transform a list of observables into hashes"
+  [observables :- [Observable]]
+  (map #(observable->observable-hash %) observables))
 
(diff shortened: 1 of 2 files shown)


src/ctia/stores/es/sighting.clj
```diff
@@ -1,28 +1,74 @@
...
+            [schema.core :as s]))
+
+(s/defschema ESStoredSighting
+  (st/merge StoredSighting
+            {:observables_hash [s/Str]}))
+
+(def es-coerce! (crud/coerce-to-fn [(s/maybe ESStoredSighting)]))
+
+(s/defn observable->observable-hash :- s/Str
+  [{:keys [type value] :as obsrvable :- Observable}]
```
yogsototh: - the `:as obsrvable` is not necessary

- `type` conflict with `clojure.core/type` why not `observable-type`?

src/ctia/stores/es/sighting.clj
```diff
@@ -1,28 +1,74 @@
...
+  (let [read-fn (crud/handle-read :sighting ESStoredSighting)]
+    (-> (read-fn state id)
+        es-stored-sighting->stored-sighting)))
+
+(s/defn handle-update-sighting :- StoredSighting
+  [state id realized]
+  (let [update-fn (crud/handle-update :sighting ESStoredSighting)
+        transformed (stored-sighting->es-stored-sighting realized)]
+    (-> (update-fn state id transformed)
+        es-stored-sighting->stored-sighting)))
```
yogsototh: To my eye I would prefer, also, it might save a bit of work not to generate the `update-fn` for each call. The same apply to creation:

~~~
(def update-fn (crud/handle-update :sighting ESStoredSighting)

(s/defn handle-update-sighting :- StoredSighting
  [state id realised]
  (->> realized
       stored-sighting->es-stored-sighting
       (update-fn state id)
       es-stored-sighting->stored-sighting))
~~~

src/ctia/stores/es/sighting.clj
```diff
@@ -1,28 +1,74 @@
...
+             [:data]
+             #(map es-stored-sighting->stored-sighting (es-coerce! %))))
+
+(defn handle-list-sightings
+  [state filter-map params]
+
+  (let [list-fn (crud/handle-find :sighting ESStoredSighting)
+        res (list-fn state filter-map params)]
+
+    (es-paginated-list->paginated-list res)))
```
yogsototh: same remark here:

~~~clojure
(def list-fn (crud/handle-find :sighting ESStoredSighting))

(defn handle-list-sightings
  [state filter-map params]
  (-> (list-fn state filter-map params)
      es-paginated-list->paginated-list))
~~~

yogsototh:
Looks good to me except some minor coding style details.

It would also have been a good idea to start having benchmark metrics to verify the benefits of these optimisations.

src/ctia/stores/es/sighting.clj
polygloton: Personally I'm not against variable shadowing.

PR #368, opened 2016-06-15
Support for schema changes in 0.1.4-SNAPSHOT CTIM

craigbro:
Most changes involve adding new required fields for sightings, and handling the addition of "default" values for things like the 'count' field.


saintx:
LGTM

PR #374, opened 2016-06-16
Use metrics-clojure

yogsototh:
This create a route `/stats` with this JSON:

https://gist.github.com/yogsototh/fda31016be6c634f4e3d57609fe1f9fc

The library used make adding new metrics really easy:

http://metrics-clojure.readthedocs.io/en/latest/index.html

We can then send the data to graphite, riemann or ganglia


polygloton:
Maybe move the big clode block into a gist that you link to?

src/ctia/http/middleware/metrics.clj
```diff
@@ -0,0 +1,11 @@
+(ns ctia.http.middleware.metrics
+  "Middleware to control all metrics of the server"
+  (:require [metrics.core :refer [new-registry]]
+            [metrics.meters :refer [meter]]
+            [metrics.ring.expose :refer [expose-metrics-as-json]]
+            [metrics.ring.instrument :refer [instrument]]))
+
+(defn wrap-metrics [handler]
+  (-> handler
+      (expose-metrics-as-json "/stats") ;; WARNING PUBLIC
```
saintx: We can control access to this route with nginx.

saintx:
LGTM

yogsototh:
Also I separated a `metrics` middleware from a `riemann` init function.
The `metrics` file really generate metrics only using swagger routes, so adding new routes will automatically create new metrics for them.

I put the `riemann.clj` in the middlewares folder but I don't feel it is its place. I'm not sure where to move the `riemann.clj` file for now.

I might forgot an evident thing but I can't connect to Riemann. I've opened an issue here:

https://github.com/sjl/metrics-clojure/issues/106

polygloton:
You could make a ctia.lib.riemann namespace

project.clj
```diff
@@ -85,8 +94,11 @@
                                   [prismatic/schema-generators "0.1.0"
                                    :exclusions [prismatic/schema]]]
                    :resource-paths ["model"
-                                    "test/resources"]}
-
+                                    "test/resources"]
+                   :jvm-opts ["-Dcom.sun.management.jmxremote"
+                              "-Dcom.sun.management.jmxremote.ssl=false"
+                              "-Dcom.sun.management.jmxremote.authenticate=false"
+                              "-Dcom.sun.management.jmxremote.port=5555"]}
```
polygloton: What is JMX for?  Can we put it into a different profile, so that it can be added on demand?

saintx: "JMX technology provides the tools for building distributed, Web-based, modular and dynamic solutions for managing and monitoring devices, applications, and service-driven networks." http://www.oracle.com/technetwork/articles/java/javamanagement-140525.html

polygloton: I read that much already.  ;-)  I guess I meant, "Why do we use it?"  I assume it is required for the metrics to work.  We still might want it to be in a different profile that we mix in as needed.

yogsototh: In fact I needed them to try to fix some bug, I added them in the commit.

It is not required for the metrics to work. I'll remove them.

polygloton:
The metric routes should be protected with a capability so that they are not public.  There is a :developer capability that would be perfect.

polygloton:
This looks good, except for a few nits.  I wonder if we are adding too much complexity with three ways to get metrics (YAGNI).  Also we need to make sure the metrics routes are secured so that they can't be abused (lets not make it public by default).

PR #389, opened 2016-07-06
initial cache control headers implementation

### LINKED ISSUE #370: Add request caching
supporting http headers like cacheing ETAGS and if-modified-since

gbuisson:
closes #370 

This adds HTTP Cache control headers support:

- Last-Modified is computed from :created or :updated for all entities
- ETag is composed from a sha1 of the output if it's data, else if it's a file we generate through File metas.

The result is applied as a couple of middleware to every GET routes.

### DIFF
--- a/src/ctia/http/middleware/cache_control.clj
+++ b/src/ctia/http/middleware/cache_control.clj
@@ -15,14 +15,19 @@
     File (str (.lastModified body) "-" (.length body))
     (sha1 (.getBytes (pr-str body) "UTF-8"))))
 
-(defn update-headers [headers etag body]
-  (merge headers
-         {"ETag" etag}
-         (when-let [last-modified (or (:updated body)
-                                      (:created body))]
-           {"Last-Modified" (format-rfc822-time last-modified)})))
+(defn update-headers
+  [headers etag {:keys [created updated] :as body}]
+  (let [last-modified (or updated created)]
+    (cond-> headers
+      etag (assoc "ETag" etag)
+      last-modified (assoc "Last-Modified"
+                           (format-rfc822-time last-modified)))))
 
-(defn wrap-cache-control-headers [handler]
+(defn wrap-cache-control-headers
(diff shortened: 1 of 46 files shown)


yogsototh:
LGTM

src/ctia/http/middleware/cache_control.clj
```diff
@@ -0,0 +1,33 @@
...
+  (= (:status response) 200))
+
+(defn calculate-etag [body]
+  (case (class body)
+    String (sha1 body)
+    File (str (.lastModified body) "-" (.length body))
+    (sha1 (.getBytes (pr-str body) "UTF-8"))))
+
+(defn update-headers [headers etag body]
+  (merge headers
```
polygloton: You could use cond-> with assoc instead of merge (it makes the when-let below a little cleaner, IMO).  For the "ETag" key, you could assoc it under the :else predicate.  Just a nit, though.  :)

polygloton:
Could the middleware declarations be added to ctia.http.handler/app?  Then it would only have to be declared in one place.

test/ctia/http/middleware/cache_control_test.clj
```diff
@@ -0,0 +1,77 @@
...
+                                    helpers/fixture-properties:clean
+                                    whoami-helpers/fixture-server]))
+
+(use-fixtures :each whoami-helpers/fixture-reset-state)
+
+(defn get-actor [id headers]
+  (select-keys (get (str "ctia/actor/" id)
+                    :headers (merge headers {"api_key" "45c1f5e3f05d0"})) [:status :headers :parsed-body]))
+
+(deftest-for-each-store test-cache-control-middleware
```
polygloton: Could you get away with just testing one of the stores?

polygloton:
Looks good, just a couple of nits.

gbuisson:
I improved my PR from feedback

PR #395, opened 2016-07-11
fixed issues related to latest CTIM version

### LINKED ISSUE #394: ES mapping is incorrect
Our current tests are passing because we don't generate optional keys on schema generative tests.

- Update CTIM so that optional keys are generated
- Fix CTIA/CTIM related issues

gbuisson:
depends on https://github.com/threatgrid/ctim/pull/31
closes #394 

- updated ES mapping with missing fields
- small fixes linked to latest CTIM version

### DIFF
--- a/src/ctia/stores/es/mapping.clj
+++ b/src/ctia/stores/es/mapping.clj
@@ -668,6 +668,3 @@
          exploit-target-mapping
          sighting-mapping
          identity-mapping))
-
-
-(keys  incident-mapping)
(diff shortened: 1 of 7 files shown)


src/ctia/stores/es/mapping.clj
```diff
@@ -609,3 +668,6 @@
          exploit-target-mapping
          sighting-mapping
          identity-mapping))
+
+
+(keys  incident-mapping)
```
yogsototh: I believe this is a forgotten call for local testing. This should be removed.

gbuisson: oops!

yogsototh:
LGTM!

PR #396, opened 2016-07-11
Removed URI properties for Redis and ES (standardizing on host & port)

### LINKED ISSUE #392: ElasticSearch store URI needs to be removed
The ElasticSearch store and hook URIs in the properties have been replaced with host/port values instead. The hook uri property was removed from the schema, meaning that using this URI now gives the user an error.

The store uri property was left in the schema, meaning that it can still appear in properties, with no errors. However, this uri value is now ignored, and the defaults will be used instead, which may cause confusion if it gets used accidentally.

The property to remove is `ctia.store.es.default.uri`

polygloton:
- Also removed commented-out SQL properties

Closes #392


yogsototh:
LGTM!

gbuisson:
LGTM

PR #397, opened 2016-07-12
Cleaned up shutdown behavior

### LINKED ISSUE #268: Cleaner Shutdown Hook
We have a few places in the code where we set shutdown hooks.

```clojure
(.addShutdownHook ...)
```

We usually do this in some initialization function.  However, initialization code may be run multiple times, particularly when we run tests.  This is resulting in lots of pending shutdown hooks.  Its probably not a big deal, but it isn't very clean.  Let's have a single shutdown hook, and we can allow initializer functions to register shutdown functions.  They can be registered with a keyword, so that each time the initializer runs, it replaces is previous registration.  Since each initializer function should have a sibling shutdown function, and since the tests should be calling the shutdown function, there should be no issues with left-over state between tests.

polygloton:
Added a shutdown ns as a sibling to init
Consolidates shutdown behavior
Uses shutdown hooks to avoid multiple ".addShutdownHook" calls

Closes #268

### DIFF
--- a/src/ctia/shutdown.clj
+++ b/src/ctia/shutdown.clj
@@ -1,0 +1,22 @@
+(ns ctia.shutdown
+  (:require [clojure.tools.logging :as log]))
+
+(defonce shutdown-hooks (atom {:registered? false :hooks {}}))
+
+(defn register-hook! [key f]
+  (swap! shutdown-hooks assoc-in [:hooks key] f))
+
+(defn shutdown-ctia! []
+  (doseq [[name hook] (:hooks @shutdown-hooks)]
+    (try
+      (hook)
+      (catch Exception excep
+        (log/error excep (format "Shutdown hook %s failed" name)))))
+  (swap! shutdown-hooks assoc :hooks {}))
+
+(defn init! []
+  (when-not (:registered? @shutdown-hooks)
+    (.addShutdownHook
+     (Runtime/getRuntime)
+     (Thread. shutdown-ctia!))
+    (swap! shutdown-hooks assoc :registered? true)))
--- a/src/ctia/shutdown.clj
+++ b/src/ctia/shutdown.clj
@@ -1,22 +1,34 @@
 (ns ctia.shutdown
   (:require [clojure.tools.logging :as log]))
 
-(defonce shutdown-hooks (atom {:registered? false :hooks {}}))
+(defonce shutdown-hooks (ref {:registered? false :hooks {}}))
 
 (defn register-hook! [key f]
-  (swap! shutdown-hooks assoc-in [:hooks key] f))
+  (dosync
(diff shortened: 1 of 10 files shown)


src/ctia/shutdown.clj
```diff
@@ -0,0 +1,22 @@
...
+      (catch Exception excep
+        (log/error excep (format "Shutdown hook %s failed" name)))))
+  (swap! shutdown-hooks assoc :hooks {}))
+
+(defn init! []
+  (when-not (:registered? @shutdown-hooks)
+    (.addShutdownHook
+     (Runtime/getRuntime)
+     (Thread. shutdown-ctia!))
+    (swap! shutdown-hooks assoc :registered? true)))
```
yogsototh: I think there could still be a race condition here if we call `init!` concurrently. I'm not sure this is really necessary as `init!` shouldn't be called concurrently. But just in case I believe we could prevent any problem by using something like:

~~~clojure
(let [can-add-hook (ref false)]
       (dosync
           (when-not (:registered? @shutdown-hooks)
                (swap! shutdown-hook assoc registered? true)
                (ref-set can-add-hook true)))
        (when @can-add-hook 
            (.addShutdownHook
                 (Runtime/getRuntime)
                 (Thread. shutdown-ctia)))
~~~

yogsototh:
LGTM except if we really need to support call to `init!` concurrently.

gbuisson:
LGTM

src/ctia/shutdown.clj
polygloton: That's a good point, but we really don't need to support concurrency.  If we did, I'd want to use an agent (for side-effects) to set the hook.  Its not needed, but I'll change it, just to be pedantic.  :)

polygloton:
@yogsototh Please re-review :-)

src/ctia/shutdown.clj
```diff
@@ -0,0 +1,28 @@
...
+(defn register-hook! [key f]
+  (dosync
+   (alter shutdown-hooks assoc-in [:hooks key] f)))
+
+(defn shutdown-ctia! []
+  (doseq [[name hook] (:hooks @shutdown-hooks)]
+    (try
+      (hook)
+      (catch Exception excep
+        (log/error excep (format "Shutdown hook %s failed" name)))))
```
polygloton: I didn't put this in a transaction because I don't know how safe it is to rely on the agent-send-off thread pool at shutdown time.  (Hooks would have side effects, so I'd want to use send-off).  So calling shutdown-ctia! isn't completely thread safe.  Two threads theoretically could call it at the same time and trigger the hooks concurrently.  Its not likely to happen in practice, however.  Shutdown hooks should be written defensively anyhow.

yogsototh:
LGTM!

PR #398, opened 2016-07-15
Durable atoms

### LINKED ISSUE #72: Implement durable atoms in memory store

polygloton:
Closes #72


polygloton:
Not sure if we really still want this, but if so, here it is!  :)

test/ctia/test_helpers/atom.clj
```diff
@@ -0,0 +1,83 @@
...
+                    "ctia.store.atom.sighting.mode" "memory"
+                    "ctia.store.ttp" "atom"
+                    "ctia.store.atom.ttp.mode" "memory"]
+    (f)))
+
+(defn fixture-properties:atom-durable-store [f]
+  ;; Set properties to enable the atom store
+  (with-properties ["ctia.store.actor" "atom"
+                    "ctia.store.atom.actor.mode" "durable"
+                    "ctia.store.atom.actor.path" "/tmp/durable-actor.edn"
```
gbuisson: I think we might want to have all those files on a shared top folder like for example /tmp/ctia/<file>.edn so we can easily delete everything.

yogsototh: `/tmp` is also usually deleted after a reboot, I would find it safer to use `/var/tmp` instead.

yogsototh: In the github interface I didn't saw this was the directory set only for tests, so we should stick to `/tmp`.

yogsototh:
LGTM!

gbuisson:
LGTM if you move the edn files ;-)

test/ctia/test_helpers/atom.clj
polygloton: The problem with a directory under /tmp is that something would need to create that directory.   We can just "rm durable-*.edn" when we want to.

saintx:
Agreed with @yogsototh above. /var is the correct location for variable data.  Nothing under /tmp can reasonably be called "durable".

polygloton:
I doesn't need to be durable, though.  It is test data that gets reset between each test.

craigbro:
I think durable atoms is a useful store, it should default to storing somewhere under the current directory, but be configurable.

PR #401, opened 2016-07-27
fix-init-properties-script: fix init-properties script

yogsototh:


scripts/init-properties-for-docker.sh
```diff
@@ -21,7 +21,7 @@ ctia.store.es.default.host=$ip
 ctia.store.es.default.port=9200
 ctia.hook.es.host=$ip
 ctia.hook.es.port=9200
-ctia.hook.redis.uri=http://${ip}:6379
```
quoll: The expression should be wrapped in double quotes. So if there is ever an error that puts a space into the "ip" var, you don't accidentally set the uri to only include the first thing, and then get an error as it tries to execute the next term

yogsototh: It should be good now.

scripts/init-properties-for-docker.sh
```diff
@@ -17,11 +17,11 @@ if [ -f $resourcefile ]; then
 fi
 
 cat > $resourcefile <<EOF
-ctia.store.es.default.host=$ip
+ctia.store.es.default.host="$ip"
```
quoll: I did not see the:
`cat > $resourcefile <<EOF`
in the previous context, so I did not realize it was going to a properties file, sorry. In that case you do not want quotes. But the move to host/port is preferable to URI.

yogsototh: so back to previous version :).

The `uri` doesn't exists anymore, this is why I had to make this change, to use `host` instead.

PR #404, opened 2016-08-08
Search entities by external id

### LINKED ISSUE #376: Endpoints for all entities, which allow suser to search for exact match on external_ids
We recently added external_ids to the BaseEntity mixin, and we want to provide an API call that lets users return 0-N objects that have the exact match (case insensitive) for a given external_id.

While the existing filter-map object listing endpoints will also support this, I think it's an operation that we want to have a specific capability for, and be able to grant it to a device, and only let them do external IDs.  It's one way of letting other devices manage their data in our system, without being able to can it all.

gbuisson:
closes #376 

- Added search by external id routes for all entities
- Added relevant tests

### DIFF
--- a/src/ctia/auth.clj
+++ b/src/ctia/auth.clj
@@ -16,49 +16,41 @@
   #{;; Actor
     :create-actor
     :read-actor
-    :list-actors-by-external-id
     :delete-actor
 
     ;; Campaign
     :create-campaign
     :read-campaign
-    :list-campaigns-by-external-id
     :delete-campaign
 
     ;; COA
     :create-coa
     :read-coa
-    :list-coas-by-external-id
     :delete-coa
 
     ;; Exploit-Target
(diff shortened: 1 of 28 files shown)


src/ctia/auth.clj
```diff
@@ -16,62 +16,71 @@
   #{;; Actor
     :create-actor
     :read-actor
+    :list-actors-by-external-id
```
polygloton: I think you should just add one capability called :external-id, and then compose it with the proper :read-<entity> capabilities (eg ":capabilities #{:read-actor :external-id}").  Notice that there are no other :list-<thingy> style capabilities here now.  I'm not necessarily against :list-<thingy> style capabilities, but I think we should prefer composing simpler capabilities when practical.

polygloton: Could be :list-by-external-id instead.

PR #417, opened 2016-08-31
CTIM 0.1.8 support

### LINKED ISSUE #415: Typo on sighting test specs
There's a typo on our tests:
Sightings specs test TTPs instead of Sightings

see: 
- https://github.com/threatgrid/ctia/blob/master/test/ctia/http/generative/atom_store_spec.clj#L40
- https://github.com/threatgrid/ctia/blob/master/test/ctia/http/generative/es_store_spec.clj#L40

gbuisson:
This PR adds CTIM 0.1.8 support which includes the Bundle Entity

also closes #415

### DIFF
--- a/src/ctia/auth.clj
+++ b/src/ctia/auth.clj
@@ -68,6 +68,11 @@
     ;; Verdict
     :read-verdict
 
+    ;; Bundle
+    :create-bundle
+    :read-bundle
+    :delete-bundle
+
     ;; Other
     :developer
     :specify-id
--- a/test/ctia/http/generative/atom_store_spec.clj
+++ b/test/ctia/http/generative/atom_store_spec.clj
@@ -38,7 +38,10 @@
   specs/spec-judgement-routes)
 
 (defspec spec-sighting-routes-atom-store
-  specs/spec-ttp-routes)
+  specs/spec-sighting-routes)
 
 (defspec spec-ttp-routes-atom-store
   specs/spec-ttp-routes)
+
+(defspec spec-bundle-routes-atom-store 50
+  specs/spec-bundle-routes)
--- a/test/ctia/http/routes/bundle_test.clj
+++ b/test/ctia/http/routes/bundle_test.clj
@@ -1,0 +1,117 @@
+(ns ctia.http.routes.bundle-test
+  (:refer-clojure :exclude [get])
+  (:require [clj-momo.test-helpers.core :as mth]
+            [clojure.test :refer [is join-fixtures testing use-fixtures]]
+            [ctia.domain.entities :refer [schema-version]]
+            [ctim.schemas.common :as c]
+            [ctia.test-helpers
+             [auth :refer [all-capabilities]]
+             [core :as helpers :refer [delete get post put]]
+             [fake-whoami-service :as whoami-helpers]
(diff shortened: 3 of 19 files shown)


polygloton:
Here are the [CTIM changes](https://github.com/threatgrid/ctim/compare/c53d96...34cd91)

test/ctia/http/generative/atom_store_spec.clj
```diff
@@ -38,7 +38,10 @@
   specs/spec-judgement-routes)
 
 (defspec spec-sighting-routes-atom-store
-  specs/spec-ttp-routes)
+  specs/spec-sighting-routes)
 
 (defspec spec-ttp-routes-atom-store
   specs/spec-ttp-routes)
+
+(defspec spec-bundle-routes-atom-store 50
```
polygloton: I would like us to limit complexity in the generators rather than override the number of tests.  Seeing 50 here is a smell, I think.  I believe that if we are specifying the number of tests, then there is something wrong with our generators.

test/ctia/http/generative/atom_store_spec.clj
```diff
@@ -38,7 +38,10 @@
   specs/spec-judgement-routes)
 
 (defspec spec-sighting-routes-atom-store
-  specs/spec-ttp-routes)
+  specs/spec-sighting-routes)
```
polygloton: Wow, good catch.

src/ctia/auth.clj
```diff
@@ -68,6 +68,11 @@
     ;; Verdict
     :read-verdict
 
+    ;; Bundle
+    :create-bundle
+    :read-bundle
+    :delete-bundle
+
```
polygloton: Do you think we should add :read-bundle to default-capabilities?

test/ctia/http/routes/bundle_test.clj
```diff
@@ -0,0 +1,117 @@
...
+          (is (= 200 (:status response)))
+          (is (deep=
+               new-bundle
+               (dissoc bundle
+                       :id
+                       :created
+                       :modified
+                       :tlp
+                       :owner)))))
+
```
polygloton: I think we should add tests for getting an individual entity that was added in the bundle.

polygloton:
It doesn't seem like bundles create different entities, though they do embed entities.  Is that on purpose?  It seems to me like the bundle handler should invoke other store handlers to create those entities.  Even if you don't want to do that in this PR, maybe we should capture it in a new issue.

gbuisson:
Currently I see Bundle Storage more like some kind of "carbon copy" for the result of an enrichment of a sighting across many instances of CTIA at some point in time, this is to ensure that the produced material would be easily available and unaltered in any way for the rest of the processing loop.

In this PR we use our Atom and ES Stores but we could store them as compressed files on a distributed FS for instance.

I definitely see an overlap with the bulk routes though, and we could somehow merge these 2 behaviours, does that sound like a good idea?

src/ctia/auth.clj
gbuisson: I'm not sure, this entity can hold huge records and impact performance, I would make it so that a restricted set of specific applications are allowed to use it.

polygloton:
That makes sense.

PR #418, opened 2016-09-06
Using correct IDs in API results

### LINKED ISSUE #212: Semantically valuable IDs

polygloton:
APIs return long-style IDs for entities.

Stores are still use short-style IDs, and are decoupled from the API which handles the ID conversion.

Part of #212


src/ctia/http/routes/judgement.clj
```diff
@@ -2,11 +2,13 @@
   (:require
    [compojure.api.sweet :refer :all]
    [ctia.domain.entities :refer [realize-feedback realize-judgement]]
+   [ctia.domain.entities.judgement :refer [with-long-id page-with-long-id]]
    [ctia.flows.crud :as flows]
-   [ctia.http.routes.common :refer [created paginated-ok PagingParams]]
-   [ctia.properties :refer [properties]]
+   [ctia.http.routes.common
+    :refer [created paginated-ok PagingParams]]
```
polygloton: It turned pink in my emacs :)

src/ctia/domain/entities/actor.clj
```diff
@@ -0,0 +1,13 @@
...
+    [ctim.domain.id :as id]))
+
+(def short-id->long-id
+  (id/factory:short-id->long-id :actor get-http-show))
+
+(defn with-long-id [{id :id, :as entity}]
+  (assoc entity :id (short-id->long-id id)))
+
+(defn page-with-long-id [{data :data, :as m}]
+  (assoc m :data (map with-long-id data)))
```
quoll: My first inclination here is to use `update-in` rather than `assoc`, because it's more descriptive of what's happening, and there are fewer things going on. 

`(defn page-with-long-id [{data :data, :as m}]
   (update-in m [:data] (partial map with-long-id))`

src/ctia/domain/entities/campaign.clj
```diff
@@ -0,0 +1,13 @@
...
+    [ctim.domain.id :as id]))
+
+(def short-id->long-id
+  (id/factory:short-id->long-id :campaign get-http-show))
+
+(defn with-long-id [{id :id, :as entity}]
+  (assoc entity :id (short-id->long-id id)))
+
+(defn page-with-long-id [{data :data, :as m}]
+  (assoc m :data (map with-long-id data)))
```
quoll: Can also use `update-in` for `with-long-id` and `page-with-long-id`

src/ctia/http/routes/bulk.clj
```diff
@@ -13,9 +24,9 @@
             [schema.core :as s]))
 
 (defn realize
-  "return the realize function provided an entity key name"
+  "return the realize function provided an entity type key"
   [k]
-  (condp = k
+  (case k
```
quoll: Good catch

src/ctia/domain/entities/campaign.clj
AlexBaranosky:
```clj
(defn page-with-long-id [m]
  (update m :data #(map with-long-id %)))
```

src/ctia/http/routes/bulk.clj
```diff
@@ -28,7 +39,7 @@
     :ttp            ent/realize-ttp))
 
 (defn create-fn
-  "return the create function provided an entity key name"
+  "return the create function provided an entity type key"
   [k]
   #(write-store k (case k
```
quoll: I appreciate that this isn't in the current PR, but thought I should comment.
`#()` anonymous functions are great for in place quick calls. But for a function that returns a function I don't think I've ever seen anything used except `(fn []...)`. It also lets you document the parameter for the returned function.

The other approach I'd use would be to use `partial` on `k` and the `create-*` function returned in the case. That does lose the label for the parameter though

src/ctia/http/routes/bulk.clj
```diff
@@ -57,27 +68,47 @@
                    :sighting       read-sighting
                    :ttp            read-ttp) %))
 
+(defn with-long-id-fn
+  "return the with-long-id function provided an entity type key"
+  [k]
+  (case k
```
quoll: Given that it's a value to value mapping (keyword to function), then wouldn't a map literal be better?

src/ctia/http/routes/bulk.clj
```diff
@@ -57,27 +68,47 @@
...
-                   :store-fn (create-fn entity-type)
-                   :identity login
-                   :entity %)
-                  (catch Exception e
-                    (do (log/error (pr-str e))
-                        nil))))
-       (map :id)))
+  (let [with-long-id (with-long-id-fn entity-type)]
+    (->> entities
+         (map #(try
```
quoll: This function that starts with `try` is quite large, and breaks up the threading macro. I suggest labeling it in the `let` block and calling map on that function.

src/ctia/domain/entities/verdict.clj
```diff
@@ -0,0 +1,6 @@
+(ns ctia.domain.entities.verdict
+  (:require
+    [ctia.domain.entities.judgement :as judgement]))
+
```
polygloton: Nope.

src/ctia/http/routes/bulk.clj
polygloton: I was just following the pattern of the surrounding code.  Perhaps open an issue and submit a PR to refactor this NS?

src/ctia/http/routes/bulk.clj
polygloton: I agree, the fn should be extracted out.

src/ctia/domain/entities/campaign.clj
polygloton: I agree update is nicer than update-in.  If it will become more readable, I will change assoc to update.

src/ctia/properties.clj
```diff
@@ -113,3 +113,6 @@
 (def init! (mp/build-init-fn files
                              PropertiesSchema
                              properties))
+
+(defn get-http-show []
+  (get-in @properties [:ctia :http :show]))
```
yogsototh: It might be overkill to do it right now. But `get-http-show` will be called a lot and the properties won't change much often. Wouldn't it be a good idea to memoize it for at least few seconds?

PR #422, opened 2016-09-23
CTIM PR #61 Support - Bundle Changes

gbuisson:
This PR holds the necessary changes to make it work with CTIM Bundle latest changes.

- Updated ES Mapping
- ring-swagger seems to translates `set` container as `vector`, added a custom bundle spec with a coercion before validation
- updated bundle tests


yogsototh:
LGTM!

PR #432, opened 2016-10-12
Flanders upgrade & DataTable support

gbuisson:
depends on https://github.com/threatgrid/ctim/pull/80

- Upgraded the project to use schemas generated from Flanders
- DataTable support including store, routes tests and spec


src/ctia/schemas/core.clj
```diff
@@ -0,0 +1,85 @@
...
+             [relationships :as rels]
+             [sighting :as ss]
+             [ttp :as ttps]
+             [verdict :as vs]
+             [vocabularies :as vocs]]
+            [flanders.schema :as fs]
+            [schema.core :as s]))
+
+;; actor
+(def NewActor (fs/->schema-tree as/NewActor))
```
polygloton: If you wanted to avoid def'ing these, you could use the memoized get-schema fn.  Not a preference, just an option.

gbuisson: I was thinking about going this way too and I actually started like that, but having to repeat the same call each time was a deal breaker, also it greatly simplifies our ns require statements.

project.clj
```diff
@@ -38,13 +39,20 @@
                  [org.clojure/tools.cli "0.3.5"]
                  [pandect "0.6.0"]
                  [threatgrid/clj-momo "0.2.1"]
-                 
+
                  ;; Schemas
                  [prismatic/schema ~schema-version]
                  [metosin/schema-tools ~schema-tools-version]
-                 [threatgrid/ctim "0.1.9"]
+                 [threatgrid/ctim "0.2.0-SNAPSHOT"
```
polygloton: 0.2.0 is on clojars now

project.clj
```diff
@@ -38,13 +39,20 @@
...
+                  ;; Exclusions:
+                  ;; - ring-swagger as 0.22.9 lacks s/Any support
+                  :exclusions [metosin/ring-swagger]]
 
                  ;; Web server
+                 ;; ring-swagger 0.22.10 provided by compojure-api
+                 ;; 1.1.8 fails to generate JSON schema for s/Any
+                 ;; overriding here
+                 [metosin/ring-swagger "0.22.11"]
```
polygloton: I opened threatgrid/ctim#81 to update this in CTIM.

PR #434, opened 2016-10-13
RedisMQ Event Queue Hook

craigbro:
Add support for RedisMQ Event Queue, defaults to disabled

### DIFF
--- a/test/ctia/flows/hooks/redismq_event_hook_test.clj
+++ b/test/ctia/flows/hooks/redismq_event_hook_test.clj
@@ -1,0 +1,66 @@
+(ns ctia.flows.hooks.redismq-event-hook-test
+  (:require [clj-momo.test-helpers.core :as mth]
+            [clojure.test :refer [deftest is join-fixtures testing use-fixtures]]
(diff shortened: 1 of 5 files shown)


test/ctia/flows/hooks/redismq_event_hook_test.clj
```diff
@@ -0,0 +1,66 @@
...
+                                :type "ip"}
+                   :disposition 3
+                   :source "source"
+                   :tlp "green"
+                   :priority 100
+                   :severity 100
+                   :confidence "Low"
+                   :valid_time {:start_time "2016-02-11T00:40:48.212-00:00"}})
+      (is (= 3 (rmq/current-depth queue)))
+      )))
```
yogsototh: I believe it will be useful to check what's occurring when post are made concurrently. So one other test might be to post n judgements (about 100?) concurrently and wait for the queue depth to be n or fail if it doesn't reach n.

yogsototh: Also it might be safer to check the values of queued entities.

PR #446, opened 2016-11-01
Add support for filtered query string searches for ES stores

### LINKED ISSUE #444: Add support for query to ctia.lib.es.document/search-docs
We need to pass a query string into a ES document search, and we currently can pass in queries as part of the paras map. This is confusing, so I made search-doc take an explicit query param.  I also made the filter-map->term-filter function accept a query.

craigbro:
Closes #444


craigbro:
Including in #450

PR #448, opened 2016-11-02
Improve bulk insert performance

polygloton:
Performance is (hopefully) improved in three ways:

- Uses the bulk API for all creates when inserting into ES
- The flow code is changed to be asynchronous to allow for create methods on the store to happen in parallel (uses core.async and dedicated threads)
- In the flow code, events and store methods are handled in parallel

### DIFF
--- a/src/ctia/flows/crud.clj
+++ b/src/ctia/flows/crud.clj
@@ -1,14 +1,36 @@
 (ns ctia.flows.crud
   "This namespace handle all necessary flows for creating, updating and deleting entities."
   (:import java.util.UUID)
-  (:require [clojure.tools.logging :as log]
+  (:require [clojure.core.async :as a]
+            [clojure.core.async.impl.protocols :as ap]
+            [clojure.tools.logging :as log]
             [ctia.auth :as auth]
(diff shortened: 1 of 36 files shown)


src/ctia/flows/crud.clj
```diff
@@ -22,64 +44,147 @@
...
+            :update (realize-fn entity entity-id login prev-entity)
+            :delete entity))))
+
+(defn- apply-create-event-fn
+  [create-event-fn realized prev-entity flow-type login]
+  (try
+    (if (= :update flow-type)
+      (create-event-fn realized prev-entity)
+      (create-event-fn realized))
+    (catch Throwable e
```
polygloton: Need to move this try/catch into ctia.flows.hooks-pipe

craigbro:
Adding core.async here means that we cannot use hystrix.

polygloton:
I loaded data into CTIA using hydrant.  With code from master it took 1:34:57.38 total time.  With code form this branch it took 52:42.82 total time.  We aren't going to use this code, but it is an interesting reference for improving performance.

PR #450, opened 2016-11-02
Entity search

### LINKED ISSUE #445: Add query string search support for entites.
We want to allow people to use Lucene/ES query string queries combined with filter maps to list and search for entities.  This requires a new capability as well.

New endpoint is `/ENTITY/search`

It takes a query string, as well as limit and offset arguments.

It requires the user have a new capability `search-ENTITY`

craigbro:
This adds query string search to all entities.

Closes #445

### DIFF
--- a/src/ctia/auth.clj
+++ b/src/ctia/auth.clj
@@ -22,11 +22,13 @@
     :create-campaign
     :read-campaign
     :delete-campaign
+    :search-campaign
 
     ;; COA
     :create-coa
     :read-coa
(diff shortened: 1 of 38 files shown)


gbuisson:
LGTM!

craigbro:
Does not include support for the following entities, they are being broken out into distinct stories:

* Incident
* Exploit Target
* COA - implemented but failing for unknown reasons
* Relationship

PR #451, opened 2016-11-03
Modified stores and the flow to create multiple entities in bulk

### LINKED ISSUE #442: Improve ES write throughput
See description in my comment below:

polygloton:
- Store methods that create entities take multiple items
- Flow code is refactored and modified to support the new store methods
- Updated HTTP create routes
- When ES is the backing store, create methods use the ES bulk API

Related to #442

### DIFF
--- a/src/ctia/flows/crud.clj
+++ b/src/ctia/flows/crud.clj
@@ -1,5 +1,6 @@
 (ns ctia.flows.crud
-  "This namespace handle all necessary flows for creating, updating and deleting entities."
+  "This namespace handle all necessary flows for creating, updating
+  and deleting entities."
   (:import java.util.UUID)
   (:require [clojure.tools.logging :as log]
             [ctia.auth :as auth]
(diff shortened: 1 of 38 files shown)


src/ctia/flows/crud.clj
```diff
@@ -139,10 +219,16 @@
...
+         :prev-entity entity
+         :identity identity
+         :store-fn delete-fn
+         :create-event-fn to-delete-event}
+        apply-before-hooks
+        create-events
+        apply-store-fn
+        apply-event-hooks
+        apply-after-hooks
+        make-result)))
```
yogsototh: I really like it. It is really easier to read and understand.

src/ctia/store.clj
```diff
@@ -2,13 +2,13 @@
 
 (defprotocol IActorStore
   (read-actor [this id])
-  (create-actor [this new-actor])
+  (create-actor [this new-actors])
```
gbuisson: I think I would prefer that we keep the existing create-<entity> and add a create-<entity>s

polygloton:
I ran a timed import (the same that I did for threatgrid/ctia#448), and this PR is the fastest yet.  With this code, the import took 10 minutes.  When my previous PR, it took 50 minutes.  Code in master takes 1.5 hours.  This seems to be the winner!

polygloton:
Note that I think the difference in time vs my last PR is related to ES batch size being to small and maybe some non-optimimum timing issues with the threads, but its moot.  I think this PR is good.

src/ctia/store.clj
craigbro: Discussed in channel.  The bulk codepath in ES is used in Face and Kiries with single events, alot.  No problems observed.

src/ctia/store.clj
```diff
@@ -17,13 +17,13 @@
   (add-indicator-to-judgement [this judgement-id indicator-relationship]))
 
 (defprotocol IVerdictStore
-  (create-verdict [this new-verdict])
+  (create-verdict [this new-verdicts])
```
craigbro: Let's pluralize the create-ENTITY func names

src/ctia/stores/es/store.clj
```diff
@@ -72,8 +72,8 @@
 
 (defrecord RelationshipStore [state]
   IRelationshipStore
-  (create-relationship [_ new-relationship]
-    (rel/handle-create-relationship state new-relationship))
+  (create-relationship [_ new-relationships]
+    (rel/handle-create-relationship state new-relationships))
```
craigbro: all of these, since they are namespaced already could just be handle-create and handle-read ...

src/ctia/store.clj
polygloton: Good idea

PR #452, opened 2016-11-03
bulk-bench: bulk benchmark

yogsototh:
You should launch the bulk benchmark with

~~~
lein bench bulk
~~~

It benchmark the time to create a bulk with 20 entities of all types. Each entity being randomly generated.

Maybe it would be a good idea to launch benchmarks in the CI with goals to detect if a PR has a really bad effect on efficiency.

### DIFF
--- a/benchmarks/ctia/http/routes/bulk_bench.clj
+++ b/benchmarks/ctia/http/routes/bulk_bench.clj
@@ -1,0 +1,84 @@
+(ns ctia.http.routes.bulk-bench
+  "Benchmark could be launched by `lein bench bulk`"
+  (:require [ctia.test-helpers
+             [benchmark :refer [cleanup-ctia!
+                                setup-ctia-atom-store!
+                                setup-ctia-es-store!
+                                setup-ctia-es-store-native!]]
+             [core :as helpers :refer [delete post]]]
+            [clj-momo.lib.time :refer [now]]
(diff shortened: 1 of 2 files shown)


yogsototh:
Case:  :bulk-es-store
Evaluation count : 12 in 12 samples of 1 calls.
             Execution time mean : 10.122367 sec
    Execution time std-deviation : 6.569266 ms
   Execution time lower quantile : 10.110776 sec ( 2.5%)
   Execution time upper quantile : 10.131369 sec (97.5%)

Case:  :bulk-es-store-native
Evaluation count : 12 in 12 samples of 1 calls.
             Execution time mean : 10.117286 sec
    Execution time std-deviation : 5.384372 ms
   Execution time lower quantile : 10.107968 sec ( 2.5%)
   Execution time upper quantile : 10.124828 sec (97.5%)

Case:  :bulk-atom-store
Evaluation count : 12 in 12 samples of 1 calls.
             Execution time mean : 413.985203 ms
    Execution time std-deviation : 24.424459 ms
   Execution time lower quantile : 372.847559 ms ( 2.5%)
   Execution time upper quantile : 457.209562 ms (97.5%)

gbuisson:
LGTM

saintx:
Nice work!  This is going to be a very useful data point for us.

PR #464, opened 2016-11-07
Make bulk ES event insertions

### LINKED ISSUE #442: Improve ES write throughput

polygloton:
Currently we are inserting events into ES one at a time.  We want to do bulk insertions.  To accomplish this, we will move the ES insertion code out of the event-hook code and put it into the flow code.

Closes #442.


src/ctia/http/routes/data_table.clj
```diff
@@ -27,15 +27,15 @@
...
-                      :store-fn #(write-store :data-table create-data-tables %)
-                      :entity-type :data-table
-                      :identity identity
-                      :entities [data-table])))))
+                  (first
+                   (flows/create-flow
+                    :entity-type :data-table
+                    :realize-fn realize-data-table
+                    :store-fn #(write-store :data-table create-data-tables %)
+                    :long-id-fn with-long-id
```
craigbro: If every long-id-fn is with-long-id ...  then why are you passing it?

polygloton: It is actually bound to a different function in each namespace.

polygloton:
Lots of merge conflicts!  :)  I will tackle them Wednesday morning.

PR #466, opened 2016-11-09
Index settings support

craigbro:
Added ability ot pass index settings to the index creation calls.

This let us define analyzers.  I then defined custom analyzers that lowercase all of the terms, and also set up default search analyzers that do the same.  Lastly, I made our terms searched lowercase their terms, because they are doing exact term matches.

The end result is that full-text search works better, is not case sensitive, but we still have the ability to search for values like IP addresses and other observables that have punctuation in them.  We just need to quote them.

### DIFF
--- a/test/ctia/http/generative/es_store_spec.clj
+++ b/test/ctia/http/generative/es_store_spec.clj
@@ -49,5 +49,5 @@
 (defspec spec-ttp-routes-es-store
   specs/spec-ttp-routes)
 
-(defspec spec-bundle-routes-es-store
-  specs/spec-bundle-routes)
+(comment (defspec spec-bundle-routes-es-store
+           specs/spec-bundle-routes))
--- a/test/ctia/http/routes/judgement_test.clj
+++ b/test/ctia/http/routes/judgement_test.clj
@@ -40,6 +40,7 @@
                        :priority 100
                        :severity 100
                        :confidence "Low"
+                       :reason "This is a bad IP address that talked to some evil servers"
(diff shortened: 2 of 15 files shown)


test/ctia/http/routes/judgement_test.clj
```diff
@@ -80,13 +82,45 @@
...
                 response (get (str "ctia/judgement/search")
                               :headers {"api_key" "45c1f5e3f05d0"}
                               :query-params {"query" term})]
             (is (= 200 (:status response)))
             (is (= "Malicious" (first (map :disposition_name (:parsed-body response))))
-                "quoted term works"))
+                "IP quoted term works"))
+
+          (let [term "1.2.3.4"
+                response (get (str "ctia/judgement/search")
```
polygloton: Take a look at the destructuring on line 31.  It makes these tests a little more concise.

test/ctia/http/generative/es_store_spec.clj
```diff
@@ -49,5 +49,5 @@
 (defspec spec-ttp-routes-es-store
   specs/spec-ttp-routes)
 
-(defspec spec-bundle-routes-es-store
-  specs/spec-bundle-routes)
+(comment (defspec spec-bundle-routes-es-store
+           specs/spec-bundle-routes))
```
polygloton: Is this comment intentional?

craigbro: Yes, I think I should remove it entirely actually.

craigbro:
Going to remove bundle in a different PR

PR #468, opened 2016-11-09
safer docker properties init 

### LINKED ISSUE #465: using docker for mac, init-properties task makes bogus configuration
when using docker for mac, cita.pproperties generated documentation contains blank ips

```
ctia.store.es.default.host=
ctia.store.es.default.port=9200
ctia.hook.es.host=
ctia.hook.es.port=9200
ctia.hook.redis.host=
ctia.hook.redismq.host=
```

gbuisson:
closes #465 
- fallback to localhost when docker-machine isn't running
- set redismq hook ip

### DIFF
--- a/scripts/init-properties-for-docker.sh
+++ b/scripts/init-properties-for-docker.sh
@@ -3,7 +3,11 @@
 
 ip=127.0.0.1
 command -v docker-machine >/dev/null 2>&1 && \
-ip=$(docker-machine ip)
+    ip=$(docker-machine ip)
+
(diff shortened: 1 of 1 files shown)


scripts/init-properties-for-docker.sh
```diff
@@ -3,7 +3,11 @@
 
 ip=127.0.0.1
```
yogsototh: you should remove this line then.

polygloton:
LGTM

PR #470, opened 2016-11-09
use path-params instead of query-params for external_id lookup

### LINKED ISSUE #425: external_id in /ctia/sightings should be wrapped in { .. }
screen shot 2016-10-04 at 10 35 07 pm

gbuisson:
closes #425 

changed from /entity/external_id?external_id=<external_id>
to /entity/external_id/<external_id>


saintx:
LGTM.

PR #479, opened 2016-11-14
ctim-64-fix: fix related to issue 64 of CTIM

yogsototh:
related to https://github.com/threatgrid/ctim/pull/109

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -43,7 +43,7 @@
                  ;; Schemas
                  [prismatic/schema ~schema-version]
                  [metosin/schema-tools ~schema-tools-version]
-                 [threatgrid/ctim "0.3.2"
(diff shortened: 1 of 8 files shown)


yogsototh:
Test won't pass before a 0.3.3-SNAPSHOT release will be pushed to clojars for CTIM

polygloton:
Do you need me to push to clojars?

polygloton:
Also I think we should make a release of CTIM (non-snapshot) whenever we are about to merge a new version into CTIA.  In other words, CTIA master never uses a CTIM snapshot version.  Snapshot is good enough in a branch (also good enough in a PR that is still "In Progress").

polygloton:
Other than my comments above, this looks good to me!

PR #483, opened 2016-11-15
add configurable CORS support

### LINKED ISSUE #482: Setup CORS support
Setup CORS Access Control to allow CTIA-UI from a different domain.

- [x] add a middleware that sets the correct headers
- [x] make the allowed origins configurable
- [x] make the allowed methods configurable
- [x] set the defaults origins to localhost for easy hacking

https://developer.mozilla.org/en/docs/Web/HTTP/Access_control_CORS

gbuisson:
closes #482 

- Use ring-cors
- expose access-control-allow-origin and access-control-allow-methods config
- default development friendly config

### DIFF
--- a/resources/ctia-default.properties
+++ b/resources/ctia-default.properties
@@ -3,7 +3,7 @@ ctia.auth.threatgrid.cache=true
 
 ctia.http.enabled=true
 ctia.http.port=3000
-ctia.http.access-control-allow-origin=http://localhost,http://localhost:6886
+ctia.http.access-control-allow-origin=http://(localhost|127.0.0.1)(:\\d+)?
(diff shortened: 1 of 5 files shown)


resources/ctia-default.properties
```diff
@@ -3,6 +3,8 @@ ctia.auth.threatgrid.cache=true
 
 ctia.http.enabled=true
 ctia.http.port=3000
+ctia.http.access-control-allow-origin=http://localhost,http://localhost:6886
```
yogsototh: why not `http://(localhost:127.0.0.1)(:\d+)?` to allow all ports on localhost?

PR #487, opened 2016-11-21
CTIM issue 81 support PR

gbuisson:
related to: https://github.com/threatgrid/ctim/pull/115

will upgrade the CTIM dep to non snapshot once it is released on clojars


polygloton:
CTIM v0.3.4 is now on clojars :)

Issue #471, opened 2016-11-09
Update CTIA to use Relationship objects for all relations, and simplify API


craigbro:
See https://github.com/threatgrid/ctim/issues/107 for the changes to the CTIM that we are adjusting too.

We want to radically simplify the model for interacting with the CTIA.  That makes this a radical change to the CTIA.  I think it is worth it in that it will eliminate alot of CTIA code, and make the act of relating two entities together consistent.  Previously, adding a relationship to our graph, which would be a common operation, would require getting the object from the store, reading the list of values from the `related_COA` field, for example, adding our new value, and then PUTing it.  All the time we have to hope noone else is editing the object.

With this model, adding a relationship is simply a POST of the `Relationship` object to the `ctia/relationship` endpoint.

It does complicate reading and traversing the data, because you now have to ask for the relationships an object has in a second query.  I propose that the way to ease this complexity is to provide a GraphQL endpoint, which lets the UI compose their own views of the data according to a schema.  Additionally, we can define one or more common, compound views of an entity and it's relationships that will populate common view patterns.  For example, `ctia/judgement/:id/views/expanded` would return the base `Judgement`, as well as some minimal set of fields from up to 10 of it's `Indicators`. 

## Base Entity API

Each entity should have the following endpoints, which already exist for all the Base Entities

* POST ENTITY/ - Create the Entity
* PUT ENTITY/:ID - Update the Entity
* GET ENTITY/:ID - Get the entity with that ID
* GET ENTITY/external_id/:id - Get the entities with that external ID
* GET ENTITY/search - Query String Search
* DELETE ENTITY/:id - Delete the entity (if allowed for that entity)

### Entity Views

To support the core views of entities that we will be using in the CTIA "CRUD" UI, we should provide a set of pre-defined views of an entity.  These views will assemble data from the various objects it has relationships with, and perhaps some statistics, and return those in a well defined schema that can be used to populate the UI.

The contents of these views will be define by the UI/U team, and are specific to each entity.

#### GET ENTITY/:ID/views/listing

Returns the data need to populate the UI's "list" view of the entity.

#### GET ENTITY/:ID/views/expanded

Returns the data needed to populate the "expanded" view of the entity, which will include details from other objects it has relationships to, as well counts and first/last seen and other elements that may be needed.

#### GET ENTITY/:ID/views/historic

Returns the data needed to populate the historic view of the entity.  May include edit history, a histogram of Sightings, or a first/last seen indicator.

### GraphQL API endpoint

[...]

PR #495, opened 2016-11-28
Experimental Graphql Support

craigbro:
Added a ctia.schema.graphql namespace, which contains a GraphQL schema.  Also added ctia.http.routes.graphql namespace which contains a /graphql endpoint which should be compatible with Relay.


polygloton:
You might want to target ctim-release-4 instead of master.  The judgement.clj changes would be different, at least.  Also that is where more work for #471 is being done.

project.clj
```diff
@@ -92,9 +92,17 @@
                  [markdown-clj "0.9.86"]
                  [hiccup "1.0.5"]
 
+                 ;; CORS support
+                 [ring-cors "0.1.8"]
```
saintx: Need to bump this to 0.1.9

PR #502, opened 2016-12-06
ES5 compatibility changes

gbuisson:


gbuisson:
closing this for now as it seems that there's an incompatibility with travis and ES5

PR #503, opened 2016-12-06
ES5 Compatibility changes

gbuisson:
This PR makes CTIA compatible with ES5

This is a breaking change, previous versions of ES (2.x) will not work anymore.
other Elastic Product such as logstash and kibana also need to be upgraded to 5x as reflected on the demo docker-compose config.

- Changed Travis configuration to support ES5
- Dropped Elastisch, native transport support
- use a simple custom library based on clj-http
- Fixed queries to support ES5 changes
- Removed leftover obsolete mapping fields
- Upgraded docker-compose configuration
- Fixed benchmarks

for what it's worth, here's the result for big-actor:

`Case:  :big-actor-es-store
Evaluation count : 1440 in 60 samples of 24 calls.
             Execution time mean : 45.449006 ms
    Execution time std-deviation : 3.535146 ms
   Execution time lower quantile : 42.091023 ms ( 2.5%)
   Execution time upper quantile : 56.006288 ms (97.5%)`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,8 @@ jdk:
     - oraclejdk8
 
 before_install:
+    # ES5 needs this
+    - sudo sysctl -w vm.max_map_count=262144
     # create log & bin dir if missing
     - mkdir -p $LOG_DIR
     - mkdir -p $BIN_DIR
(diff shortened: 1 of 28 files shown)


gbuisson:
When this gets merged, you will need to reset your docker-compose setup:

on the ctia project folder run:

`docker-compose -f containers/demo/docker-compose.yml kill &&
docker-compose -f containers/demo/docker-compose.yml rm &&
docker-compose -f containers/dev/docker-compose.yml up`

PR #504, opened 2016-12-07
Validate references in API routes

### LINKED ISSUE #419: Validate references in API routes
When entities are POST/PUT to the API, validate that references are URLs, and return a 400 status/error when not.

In other words, add some sort of validation feature.

Related to #212 
Depends on threatgrid/ctim#58

polygloton:
Closes #419


src/ctia/schemas/core.clj
```diff
@@ -16,82 +16,169 @@
...
              [vocabularies :as vocs]]
             [flanders
-             [schema :as fs]
+             [schema :as f-schema]
+             [spec :as f-spec]
              [utils :as fu]]
-            [schema.core :refer [defschema Str Bool]]))
+            [schema.core :refer [Str Bool] :as sc]))
+
+(defmacro defschema [name-sym ddl spec-kw-ns]
```
gbuisson: why not using a plain fn instead of a macro here?

test/ctia/test_helpers/core.clj
```diff
@@ -139,3 +146,21 @@
   (id/->id type-kw
            (crud/make-id (name type-kw))
            (get-in @properties [:ctia :http :show])))
+
+(defn fixture-spec-validation [t]
+  (with-redefs [cs/registry-ref (atom (cs/registry))]
+    (cs/check-asserts true)
+    (t)
```
gbuisson: does it means we need to enable spec validation on prod deployment to validate references?

gbuisson: nevermind ;-)

src/ctia/flows/crud.clj
```diff
@@ -198,15 +210,18 @@
              store-fn
              identity
              entities
-             long-id-fn]}]
+             long-id-fn
+             spec]}]
```
polygloton: Good question, and I'm not sure.  Here is my thinking.  We do a round of validation with the schemas, and then maybe do another round of validation with the specs.  Since there is a performance hit, we can chose to only do the spec validation when there is a good reason.  For now, the only extra benefit is in the Reference spec, which only really applies to the Relationship entity.

src/ctia/schemas/core.clj
polygloton: I'll look into it.

polygloton: I think the main benefit is that looks cleaner.  It if was a fn, it would have to quote the name-sym in the caller.

Issue #507, opened 2016-12-09
Need java example using redismq


polygloton:
Depends on #508

PR #510, opened 2016-12-14
More interesting redismq Java example

polygloton:
Adding more to #507

### DIFF
--- a/examples/redismq-java-client/test/redismq_java_client/example_test.clj
+++ b/examples/redismq-java-client/test/redismq_java_client/example_test.clj
@@ -1,32 +1,84 @@
 (ns redismq-java-client.example-test
   (:require [cheshire.core :as json]
             [clj-momo.properties :as mp]
-            [clojure.test :refer [deftest is testing]]
+            [clojure.spec :as cs]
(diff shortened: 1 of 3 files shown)


examples/redismq-java-client/test/redismq_java_client/example_test.clj
```diff
@@ -1,32 +1,84 @@
...
+
+(def gen-coa-create-event
+  (gen/fmap (fn [[coa event-id]]
+              {:entity coa
+               :type CreateEventType
+               :id event-id})
+            (gen/tuple (cs/gen ::coa)
+                       (gen-id/gen-short-id-of-type :event))))
+
+(def gen-some-events
```
quoll: I realize that the name "gen-" is supposed to indicate that it was created using the gen library, but it reads as a verb. So it looks like a function, even though it's a vector (and every time I look at it, it's throwing me). Worth addressing?

polygloton: Actually this is a generator that returns a vector, so I think a verb is appropriate.

PR #514, opened 2017-01-09
Issue 501: ES purge task

### LINKED ISSUE #501: A lein task to nuke ES data
make a lein task that would walk all the stores, nuking indexes and templates.

gbuisson:
closes #501 

Add a task to easily purge all ES Stores through a single command.

### DIFF
--- a/src/ctia/stores/es/store.clj
+++ b/src/ctia/stores/es/store.clj
@@ -40,6 +40,10 @@
              [verdict :as ve]]
             [schema.core :as s]))
 
+(defn delete-state-indexes [{:keys [conn index config]}]
+  (when conn
+    (es-index/delete! conn (str index "*"))))
+
 (s/defn init-store-conn :- ESConnState
   "initiate an ES store connection returns
    a map containing a connection manager, dedicated store index properties"
--- a/src/ctia/task/purge_es_stores.clj
+++ b/src/ctia/task/purge_es_stores.clj
@@ -27,5 +27,5 @@
   (log/info "purging all ES Stores data")
   (setup)
   (delete-store-indexes)
-  (println "done")
+  (log/info "done")
   (System/exit 0))
--- a/src/ctia/task/purge_es_stores.clj
+++ b/src/ctia/task/purge_es_stores.clj
@@ -1,0 +1,31 @@
+(ns ctia.task.purge-es-stores
+  (:require [clojure.tools.logging :as log]
+            [ctia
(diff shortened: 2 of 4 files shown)


src/ctia/stores/es/store.clj
```diff
@@ -40,6 +40,10 @@
              [verdict :as ve]]
             [schema.core :as s]))
 
+(defn delete-state-indexes [{:keys [conn index config]}]
+  (when conn
+    (es-index/delete! conn (str index "*"))))
```
saintx: From the ES documentation:
https://www.elastic.co/guide/en/elasticsearch/reference/5.0/indices-delete-index.html

> In order to disable allowing to delete indices via wildcards or _all, set action.destructive_requires_name setting in the config to true. This setting can also be changed via the cluster update settings api.

Do we need to change the config for our local ES before this ES purge task will work correctly?

saintx: It's probably better to do a call per store that explicitly lists the indices that we want to remove.  It's dangerous to do it this way, for example, as it would needlessly clear our kibana indices.  Even if we never use it in production, it's still kind of overkill.

gbuisson: @saintx no by default if you didn't enforce wildcard you don't need to change anything, actually it works with ES vanilla config from docker.

I think you miss the real intent of the wildcard, a call is made for each store, however as indexes can be split, for instance the event index slicing I need the wildcard.

src/ctia/task/purge_es_stores.clj
```diff
@@ -0,0 +1,31 @@
...
+    (log/infof "deleting index %s" (:index state))
+    (delete-state-indexes state)))
+
+(defn -main
+  "invoke with lein run -m ctia.task.purge-es-stores"
+  []
+  (log/info "purging all ES Stores data")
+  (setup)
+  (delete-store-indexes)
+  (println "done")
```
yogsototh: just a nitpick, I would have used `log/info` instead of `println`. Except if log/info aren't displayed on the terminal by default.

quoll:
I'm good with this. Craig was requested, so I'll leave until he sees it.

polygloton:
Please consider my feedback before merging.

craigbro:
No, do not prompt user, then it becomes useless for automation.  If a user can figure out what command to call and it's called purge-es-stores we should assume they know what they want to do.

PR #518, opened 2017-01-27
Expired verdicts are recalculated when retrieved in the API

### LINKED ISSUE #516: Expired verdicts should not be returned by API routes
Routes that return verdicts should not have expired verdicts.  Currently expired verdicts are being returned.

Also, double check that start/end times are being stored in the verdict store.

polygloton:
Closes #516


polygloton:
If you are reviewing this, please read the comments in the issue too.  There is some thinking that I need help with.

PR #522, opened 2017-02-15
Removing store fields from API responses

### LINKED ISSUE #367: Some entity fields should not be returned
Such as :owner, :created, and :modified.

Maybe we could put all of these sort of keys under a :meta keyword in a nested map, or maybe we should just remove them all before sending responses.

polygloton:
Closes #367 

Using route specific middleware to implement removing the fields.  Also the cache-control middleware is moved to individual routes.


src/ctia/http/routes/exploit_target.clj
```diff
@@ -69,16 +74,19 @@
...
                 :summary "List ExploitTarget by external id"
                 :capabilities #{:read-exploit-target :external-id}
+                :middleware [wrap-un-store wrap-cache-control]
                 (paginated-ok
                  (page-with-long-id
                   (read-store :exploit-target list-exploit-targets
                               {:external_ids external_id} q))))
+
            (GET "/search" []
                 :return (s/maybe [StoredExploitTarget])
```
gbuisson: I wonder how this PR doesn't break the tests.
Here in this route you state that it shall return a vector of `StoredExploitTarget`.
`StoredExploitTarget`requires the owner field.
so either the middleware somehow disables response schema validation or it never worked.

gbuisson: I think that we should define a new schema, we have `New*` and `Stored*` schemas, I think that in this case it doesn't return nor a `NewExploitTarget` or a `StoredExploitTarget` but just an `ExploitTarget`

polygloton: The schema is applied to the return value of the handler, and the middleware is applied after that.  At least that is my understanding.

polygloton: We could move the 'un-store' operation into the handler and change the :return schema to ExploitTarget.  I used middleware because it felt cleaner.

gbuisson: well I think this is cleaner, otherwise we should remove the :return param of the route as it is wrong.

polygloton: It needs to be done in middleware so that cache-control can access the :created or :modified key.

gbuisson: and why not using the middleware and just change the schema to ExploitTarget?
It's a shortcut...

craigbro:
I was about to say "NO DON'T USE MIDDLEWARE!" but then I read Steve's comments and was like, "Oh"...

This requires some discussion in person

polygloton:
After discussing this with Craig and GB, we decided to remove the "Modified-Time" from the cache-control middleware, move the un-store operation into the handler, and set the correct :return schemas.

PR #528, opened 2017-02-23
Undoing changes to verdict chema

### LINKED ISSUE #527: In the API, return a Verdict that conforms to CTIM v0.4.3
So undo the CTIM 0.4.4 change

polygloton:
Closes #527 

- I removed the :owner field from verdict.  Its not desired and its not part of the schema (honestly I don't know how it was passing schema validation when we used CTIM 0.4.3).

- I did not remove :owner from the ES mapping (yet).  I need to find out if that would cause a problem first.

- CTIM 0.4.5 makes the :schema_version fields accept any string.

This is needed so that we can fix iroh-enrich and allow us to deploy without reloading data.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -42,7 +42,7 @@
                  ;; Schemas
                  [prismatic/schema ~schema-version]
                  [metosin/schema-tools ~schema-tools-version]
-                 [threatgrid/ctim "0.4.4"]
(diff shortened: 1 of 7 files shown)


polygloton:
I'm merging this because two PRs that depend on it have already been merged: threatgrid/iroh#255 and threatgrid/iroh-enrich#112.

PR #537, opened 2017-02-28
Fixing verdict fail on :start_date of 'today'

### LINKED ISSUE #536: When creating a judgement with :start_time of today, create verdict fails
Because of the way that the judgement search query is written in the handle-calculate-verdict code, a verdict is not created with the :start_time is today.  This is because the query rounds dates to the most recent day and then does a less-than filter.  It does the same thing for :end_time (but looks for greater-than).

Probably need to stop rounding and set the filter to be *-or-equal-to.

polygloton:
Closes #536.


polygloton:
The ES search query DSL doesn't seem to act intuitively (to me).  I need to learn more about it and certainly I need to add more tests around this issue.

src/ctia/stores/es/query.clj
```diff
@@ -3,22 +3,22 @@
...
 
-(def unexpired-time-range
+(defn- unexpired-time-range
   "ES filter that matches objects which
   valid time range is not expired"
-
+  [time-str]
   [{:range
-    {"valid_time.start_time" {"lt" "now/d"}}}
+    {"valid_time.start_time" {"lte" time-str}}}
```
saintx: good catch.

PR #543, opened 2017-03-15
Fixing verdict 500/null-pointer error

### LINKED ISSUE #530: Server error when looking up expired verdicts

polygloton:
Closes #530 

Added better tests that insert objects directly into ES (bypassing CTIA APIs).  This makes it much easier to test exactly what we want (without sleeping), but comes with some extra maintenance costs (probably).

There were two problems in the (GET "/:observable_type/:observable_value/verdict" ...) route.  Notice the new implementation.  It solves at least two failing cases.
* It was trying to return the result of write-store, which wasn't working (IDs were invalid when with-long-id was called).  Not sure why this didn't happen in the old tests.
* When calculating a new verdict returned null, the code wasn't handling nil correctly.

The existing tests were not effective in finding these bugs.  I didn't dig into *exactly* why not, but I think it was related to the timing of expirations vs API calls and getting things into ES.  By inserting objects directly into ES we avoid a lot of timing complexity.


src/ctia/http/routes/observable.clj
```diff
@@ -66,12 +67,17 @@
...
-                           (#(write-store :verdict create-verdicts [%])))
+                       (when-let [realized-verdict
+                                  (some-> (read-store :judgement
+                                                      calculate-verdict
+                                                      {:type observable_type
+                                                       :value observable_value})
+                                          entities/realize-verdict)]
+                         (log/info "Generated New Verdict:"
+                                   (pr-str realized-verdict))
+                         (write-store :verdict create-verdicts
```
craigbro: We should catch any errors here, and log it, and then continue to return the realized-verdict.  This will allow this end-point to work correctly even when the index goes into a read-only mode, or we have a transient write error.

PR #547, opened 2017-03-15
Removed the verdict store

### LINKED ISSUE #546: Remove the verdict store
Determine verdicts dynamically using a query against the judgement store (the old behavior).

polygloton:
Verdicts are determined by querying judgements in ES.

Closes #546


resources/ctia-default.properties
```diff
@@ -44,7 +44,6 @@ ctia.store.incident=es
 ctia.store.indicator=es
 ctia.store.judgement=es
 ctia.store.relationship=es
-ctia.store.verdict=es
```
saintx: We'll need a tracking PR for tenzin to update the ctia.properties file there, as well.

gbuisson:
LGTM though I'm not sure I would have removed the verdict store, perhaps I would have created another implementation and keep the ability to choose.

resources/ctia-default.properties
saintx: Nevermind, we don't need this--the default properties ship with the package itself.

PR #549, opened 2017-03-21
Multisort support

### LINKED ISSUE #541: Enable search query sort by disposition and then by date
In PR #535 we added some additional sort fields.  We would also like to be able to sort by disposition, and then by the [:valid_time :start_time] fields.  I don't know if this can be done in ES or not.  Investigate and implement if possible.

This would be a feature for searching for entities such as judgments that have a :disposition field.  It would sort by importance, as implemented in threatgrid/ctim#139.

gbuisson:
closes #541 

Added multisort support on our APIs as well as the ability to sort judgements by disposition then valid_time.

### DIFF
--- a/src/ctia/lib/es/document.clj
+++ b/src/ctia/lib/es/document.clj
@@ -159,7 +159,7 @@
                               sort_order)}}))
              (clojure.string/split sort-field-str #","))]
 
-    {:sort (apply merge {} sort-fields)}))
+    {:sort (into {} sort-fields)}))
 
 (defn params->pagination
   [{:keys [sort_by sort_order offset limit]
--- a/src/ctia/lib/es/document.clj
+++ b/src/ctia/lib/es/document.clj
@@ -146,6 +146,21 @@
       safe-es-read
       :found))
 
+(defn sort-params
+  [sort_by sort_order]
+  (let [sort-field-str (if (keyword? sort_by)
+                         (name sort_by)
+                         sort_by)
+        sort-fields
+        (map (fn [field]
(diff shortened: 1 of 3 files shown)


src/ctia/lib/es/document.clj
```diff
@@ -146,6 +146,21 @@
...
+                         sort_by)
+        sort-fields
+        (map (fn [field]
+               (let [sp (clojure.string/split field #":")]
+                 {(first sp)
+                  {:order (or (second sp)
+                              sort_order)}}))
+             (clojure.string/split sort-field-str #","))]
+
+    {:sort (apply merge {} sort-fields)}))
```
polygloton: Should be (into {} sort-fields), I think.

test/ctia/http/routes/observable/judgements_test.clj
```diff
@@ -65,5 +87,14 @@
           (get "ctia/ip/10.0.0.1/judgements"
                :headers {"api_key" "45c1f5e3f05d0"})]
       (is (= 200 status))
-      (is (= #{"judgement-1" "judgement-2"}
-             (set (mapcat :external_ids judgements)))))))
+      (is (= #{"judgement-1" "judgement-2" "judgement-3"}
+             (set (mapcat :external_ids judgements))))))
+
+  (testing "GET /:observable_type/:observable_value/judgements?sort_by=disposition%3Aasc%2Cvalid_time.start_time%3Adesc"
```
polygloton: Maybe think about putting the query parameters into the get parameters so its easier to read, so (get ... :query-params {...}).

Issue #552, opened 2017-03-31
Searching an IPV6 address without escaping returning 500 Server error (unknown-exception)


jodiazgo:
closing and moving to another repo

gbuisson:
Reopening it as it is a valid issue, an invalid lucene query shouldn't be a 500 but a 400 http status.

PR #556, opened 2017-04-26
Fixed Search query error feedback

gbuisson:
connected to #552 
needs https://github.com/threatgrid/clj-momo/pull/8

This PR improves API feedback regarding malformed ES lucene queries.
instead of returning a `500`it will return a `400` with a glimpse of the ES error message.

Also removed CTIA ES code and switched to the namespaces provided by clj-momo.


gbuisson:
This PR will fail until https://github.com/threatgrid/clj-momo/pull/8/files is merged.

polygloton:
+62 −647

PR #558, opened 2017-05-01
Extracting metrics code to clj-momo

polygloton:
Supports threatgrid/iroh#624

Depends on threatgrid/clj-momo#9 and threatgrid/ctim#152.


project.clj
```diff
@@ -43,6 +43,7 @@
                  [prismatic/schema ~schema-version]
                  [metosin/schema-tools ~schema-tools-version]
                  [threatgrid/ctim "0.4.9"]
+                 [threatgrid/clj-momo "0.2.6-SNAPSHOT"]
```
polygloton: This should come from CTIM.  Release an updated CTIM once the clj-momo changes are stable.

Issue #550, opened 2017-03-31
/ctia/graphql endpoint duplicated in Swagger UI


folbricht:
![screen shot 2017-03-30 at 9 41 02 pm](https://cloud.githubusercontent.com/assets/3505107/24535530/ed811834-1591-11e7-8323-87f7b98690f3.png)

PR #568, opened 2017-05-17
Refactoring of GraphQL interface

msprunck:
- Integration of [GraphiQL](https://github.com/graphql/graphiql) at `/graphiql`
- Integration of [GraphQL Voyager](https://github.com/APIs-guru/graphql-voyager) at `/voyager`
- Definition of GraphQL types for CTIM entities using Flanders
- Wrapper to the [graphql-java](https://github.com/graphql-java) Java API (`ctia.schemas.graphql.helpers` namespace)
- New GraphQL scalar type for dates
- Fix graphql `variables` and add the `operationName` parameter to select the query to execute if more than one is provided
- GraphQL cursor based pagination based on the Relay specification (use limit/offset)
- Fix graphql response format (can be requested in a format other than JSON)
- Added all fields to the Sighting and Indicator entities
- Added the Feedback entity with a `feedbacks` connection on Sighting, Indicator and Judgement entities.

Limitations:

- Enums containing non valid GraphQL type names (ex: "Agent Disclosure" or "amp-device") are defined as a GraphQL String type

Depends on threatgrid/flanders#11
Depends on threatgrid/ctim#158

Connected to #550


src/ctia/schemas/graphql/pagination.clj
```diff
@@ -0,0 +1,212 @@
...
+  (when cursor
+    (try
+      (-> cursor
+          b64/decode
+          (Integer/parseInt))
+      (catch Exception e
+        (log/warn e "Unable to unserialize cursor")
+        0))))
+
+(s/defn validate-paging :- PagingDirection
```
yogsototh: Would it be possible to return something like {:error s/Str :data PaginDirection} instead?
So we'll be able to use return value instead of thrown exceptions.

msprunck: Not really. `validate-paging` is called from a [`DataFetcher`](https://github.com/graphql-java/graphql-java/blob/master/src/main/java/graphql/schema/DataFetcher.java). We are in the Java world of the `graphql-java` library where the only way to add an error the `ExecutionResult` is by throwing an exception.

msprunck:
We can not use GraphQL enums for all CTIM vocabularies because enum names must match `[_A-Za-z][_0-9A-Za-z]*`. Some CTIM vocabularies contain a space or a dash.

One alternative would be to use String unions but `GraphQLUnionType` can not be used as an input type. Furthermore, `GraphQLUniontype` does not support scalar types.

polygloton:
I gave feedback on threatgrid/flanders#11

PR #570, opened 2017-06-02
Adding another sort_by option to the judgement search

polygloton:
Experimenting with a sort order that is similar to the [sorting logic](https://github.com/threatgrid/ctim/blob/master/src/ctim/domain/sorting.cljc) in CTIM.


gbuisson:
Is it ready to be reviewed?

polygloton:
I think we won't do this.  It isn't correct.  I had forgotten about it.

Issue #571, opened 2017-06-12
Handling ordering in GraphQL connections


msprunck:
All nodes in connection are currently sorted by `ID` in ascending order.

A connection should provide a new argument to specify the ordering.

As suggested [here](https://github.com/graphql/graphql-relay-js/issues/20#issuecomment-220494222) we could add an `orderBy` argument that take `[Ordering]!` where ordering is an InputObject:

```graphql
input Ordering {
  sort: Sort!
  direction: Direction! = ASC
}

enum Sort { VALID_TIME_START_TIME, VALID_TIME_END_TIME , DISPOSITION }
enum Direction { ASC, DESC }
```

You'd then do:

```graphql
{
  judgements(orderBy: [{sort: VALID_TIME_END_TIME, direction: DESC}, 
                       {sort: DISPOSITION, direction: ASC}, 
                       {sort: VALID_TIME_START_TIME, direction: DESC}])
}

```

PR #572, opened 2017-06-13
Graphql sorting

msprunck:
Connected to #571

### DIFF
--- a/test/ctia/test_helpers/graphql.clj
+++ b/test/ctia/test_helpers/graphql.clj
@@ -1,5 +1,9 @@
 (ns ctia.test-helpers.graphql
-  (:require [ctia.test-helpers.core :as helpers]))
+  (:require [clojure.test :refer [is testing]]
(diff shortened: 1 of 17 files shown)


test/ctia/test_helpers/graphql.clj
```diff
@@ -29,3 +33,135 @@
...
+      (is (= 200 status))
+      (is (empty? errors) "No errors")
+      (let [{:keys [data errors status]}
+            (query graphql-query
+                   (into variables
+                         {:sort_field (sorting/sorting-kw->enum-name sort-field)})
+                   operation-name)
+            connection-data (get-in data connection-path)
+            nodes (->> (:nodes connection-data)
+                       (map sort-field-fn)
```
yogsototh: as sort-field-fn is used by the the graphql query. I think it would be nice to have one manual test which compare datas directly to check that sort is doing something.

PR #576, opened 2017-06-29
Add single broker kafka cluster to dev docker-compose

craigbro:
This adds a single node kafka broker, with a default topic of "test"

Kafka is listening on 127.0.0.1:9092

### DIFF
--- a/containers/dev/docker-compose.yml
+++ b/containers/dev/docker-compose.yml
@@ -30,7 +30,7 @@ services:
       - "9092:9092"
     environment:
       KAFKA_ADVERTISED_HOST_NAME: 127.0.0.1
-      KAFKA_CREATE_TOPICS: test:1:1"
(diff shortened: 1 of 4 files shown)


containers/dev/docker-compose.yml
```diff
@@ -1,20 +1,36 @@
...
+     ports:
+       - "2181:2181"
+  kafka:
+    image: wurstmeister/kafka
+#    build: .
+    ports:
+      - "9092:9092"
+    environment:
+      KAFKA_ADVERTISED_HOST_NAME: 127.0.0.1
+      KAFKA_CREATE_TOPICS: test:1:1"
```
saintx:
```
Error while executing topic command : Cannot parse argument '1"' of option replication-factor
```

Just remove the trailing `"` character to fix this.

PR #585, opened 2017-08-21
JWT middleware integration

yogsototh:
Related to https://github.com/threatgrid/iroh/issues/902.

Added the JWT middleware in such a way it still work with other Auth systems, typically it is able to handle the `api-key` one.


src/ctia/http/middleware/auth.clj
```diff
@@ -3,22 +3,37 @@
             [compojure.api.meta :as meta]
             [ring.util.http-response :as http-response]))
 
-(defn wrap-authentication [handler]
+(defn add-id-to-request
+  "Given a request and a discoverd id and auth header
+  change the request accordingly"
```
msprunck: **Typo**: Given a request, a discover**E**d id and auth header, change the request accordingly.

src/ctia/http/middleware/auth.clj
```diff
@@ -3,22 +3,37 @@
...
+  [request id login auth-header]
```
msprunck: The login param is not needed. You can get it with `(auth/login id)`.

src/ctia/http/middleware/auth.clj
```diff
@@ -3,22 +3,37 @@
...
-           (assoc-in [:headers "authorization"] Authorization))))))
+    (handler
+     (if (:login request)
+       request
+       (let [auth-header (or (get-in request [:headers "authorization"])
+                             (get-in request [:query-params "Authorization"]))
+             id (auth/identity-for-token auth-service auth-header)
+             login (auth/login id)]
+         (cond-> request
+           (some? id) (add-id-to-request id login auth-header)))))))
```
msprunck: I suggest to move this logic to the `add-id-to-request` function

src/ctia/http/middleware/auth.clj
```diff
@@ -3,22 +3,37 @@
...
+(defn wrap-authentication [handler]
+  (testable-wrap-authentication handler @auth-service))
 
 (defn require-capability! [required-capability id]
-  (if (and required-capability
-           (auth/require-login? @auth/auth-service))
+  (if required-capability
     (cond
-      (not (auth/authenticated? id))
+      (or (nil? id)
```
msprunck: You could extend the `IIdentity` protocol for `nil` to only use `(auth/authenticated? id)`

src/ctia/http/server.clj
```diff
@@ -33,16 +37,30 @@
...
+         true auth/wrap-authentication
+
+         (:enabled jwt)
+         auth-jwt/wrap-jwt-to-ctia-auth
+
+         (:enabled jwt)
+         ((rjwt/wrap-jwt-auth-fn {:pubkey-path (:public-key-path jwt)
+                                  :no-jwt-handler rjwt/authorize-no-jwt-header-strategy}))
+
+         true wrap-params
```
msprunck: Why `wrap-params` has to be used a second time ? It is already wrapped by the `defapi` macro.

src/ctia/http/server.clj
```diff
@@ -33,16 +37,30 @@
...
                     (allow-methods access-control-allow-methods))
+
+         true auth/wrap-authentication
+
+         (:enabled jwt)
+         auth-jwt/wrap-jwt-to-ctia-auth
+
+         (:enabled jwt)
+         ((rjwt/wrap-jwt-auth-fn {:pubkey-path (:public-key-path jwt)
+                                  :no-jwt-handler rjwt/authorize-no-jwt-header-strategy}))
```
msprunck: Why is it not possible to add these middlewares in the defapi macro ?

src/ctia/http/middleware/auth.clj
yogsototh: I prefer to keep `add-id-to-request` pure, be it just for testing purpose.

src/ctia/http/server.clj
yogsototh: I believe this is because init ordering issue. The `api-handler` is generated before the options have been dynamically loaded.

src/ctia/http/server.clj
yogsototh: That's because of middleware ordering issue. `wrap-params` must be called before `wrap-authentication`

PR #587, opened 2017-08-31
Private CTIA - Store Access Control

gbuisson:


gbuisson:
I think this is ready for a first round of reviews.

src/ctia/schemas/core.clj
```diff
@@ -20,196 +20,225 @@
              [schema :as f-schema]
              [spec :as f-spec]
              [utils :as fu]]
-            [schema.core :refer [Str Bool] :as sc]))
+            [schema.core :refer [Str Bool] :as sc]
+            [schema-tools.core :as st]))
+
+(sc/defschema ACLStoredEntity
```
msprunck: You could merge `ACLEntity` with a map containing `:groups`

src/ctia/schemas/core.clj
```diff
@@ -20,196 +20,225 @@
...
+  (st/optional-keys
+   {:groups [Str]
```
msprunck: What is the difference between `:groups` and `authorized_groups`? If `:groups` is related to the owner, could you rename it to `owner_groups`?

src/ctia/stores/es/crud.clj
```diff
@@ -37,47 +47,117 @@
...
+  [doc ident]
+  (boolean
+   (and
+    (= (:owner doc)
+       (:login ident))
+    (clojure.set/subset?
+     (set (:groups ident))
+     (set (:groups doc))))))
+
+(s/defn allow-write? :- s/Bool
```
msprunck: Could you document these rules in a developer documentation?

src/ctia/stores/es/crud.clj
```diff
@@ -37,47 +47,117 @@
...
+          (clojure.set/subset?
+           (set (:groups ident))
+           (set (:groups current-doc)))) true
+
+     ;;CTIM models with TLP red that is owned by user FOO
+     (and (= (:tlp current-doc) "red")
+          (= (:owner current-doc)
+             (:login ident))) true)))
+
+(s/defn allow-read? :- s/Bool
```
msprunck: Could you add tests for `allow-write?`, and  `allow-read?` to check all use cases?

src/ctia/stores/es/event.clj
```diff
@@ -83,5 +84,8 @@
   [state :- ESConnState
    filter-map :- {s/Any s/Any}
    params]
-  (update (handle-list-raw state filter-map params)
+  (update (handle-list-raw state
+                           filter-map
+                           {:login "Unknown"
```
msprunck: Why the current identity is not used here?

src/ctia/stores/es/query.clj
```diff
@@ -2,6 +2,34 @@
...
+  [{:keys [login groups]}]
+  ;; TODO do we really want to discard case on that?
+  (let [login (str/lower-case login)
+        groups (map str/lower-case groups)]
+    {:bool
+     {:should
+      [{:terms {"tlp" public-tlps}}
+       ;; Document Owner
+       {:bool {:filter [{:term {"owner" login}}
+                        {:terms {"groups" groups}}]}}
```
yogsototh: Does that means intersection is not empty?

src/ctia/stores/es/query.clj
```diff
@@ -2,6 +2,34 @@
...
             [clojure.string :as str]))
 
+(def public-tlps
+  ["white" "green"])
+
+(defn find-restriction-query-part
+  [{:keys [login groups]}]
+  ;; TODO do we really want to discard case on that?
+  (let [login (str/lower-case login)
+        groups (map str/lower-case groups)]
```
yogsototh: I think we should take care that the logins and groups  cannot contains any special chars that would have a special meaning in Lucene queries.

The question is, where should we protect them, here or in the auth? Or maybe in both places.

gbuisson: it looks like the `terms` query part expects an exact match 
https://www.elastic.co/guide/en/elasticsearch/reference/5.4/query-dsl-terms-query.html

you need to use `wildcard` to perform wildcard searches:
https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-wildcard-query.html

src/ctia/stores/es/query.clj
gbuisson: looks like it according to the docs:
https://www.elastic.co/guide/en/elasticsearch/reference/5.4/query-dsl-terms-query.html

src/ctia/schemas/core.clj
```diff
@@ -20,196 +20,225 @@
...
+       (st/merge ACLStoredEntity
+                 (f-schema/->schema ~ddl)))
+     (f-spec/->spec ~ddl ~spec-kw-ns)))
+
+(defmacro def-acl-schema [name-sym ddl spec-kw-ns]
+  `(do
+     (sc/defschema ~name-sym
+       (st/merge ACLEntity
+                 (f-schema/->schema ~ddl)))
+     (f-spec/->spec ~ddl ~spec-kw-ns)))
```
polygloton: What are the consequences of having the spec and the schema not match?  We are validating new entities with spec now in CTIA.  Do they need to match?  Make sure you merge in master and then run test.

There is a feature in flanders that might help if you want to change the flanders schema definitions before generating plumatic schemas and clojure specs.  You can modified a flanders schema by composing the schema defintion with a declarative list of changes.  See [flanders.munge-test](https://github.com/threatgrid/flanders/blob/master/test/flanders/munge_test.clj) for examples.

src/ctia/stores/es/query.clj
```diff
@@ -2,6 +2,34 @@
...
+       {:bool {:filter [{:term {"owner" login}}
+                        {:terms {"groups" groups}}]}}
+
+       ;; or if user is listed in authorized_users or authorized_groups field
+       {:term {"authorized_users" login}}
+       {:terms {"authorized_groups" groups}}
+
+       ;; CTIM models with TLP amber that is owned by org BAR
+       {:bool {:must [{:term {"tlp" "amber"}}
+                      {:bool {:should [{:term {"owner" login}}
```
msprunck: In the ES query, the owner does not need to belong to the entity group? It does not seem to be consistent with the `owner?` function

src/ctia/stores/es/crud.clj
```diff
@@ -37,47 +47,130 @@
...
+        (seq (clojure.set/intersection
+              (set (:groups ident))
+              (set (:groups doc)))))))
+
+(s/defn owner? :- s/Bool
+  [doc ident]
+  (boolean
+   (and (= (:owner doc)
+           (:login ident))
+        (allowed-group? doc ident))))
```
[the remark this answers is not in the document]
gbuisson: I agree that this can be somewhat misleading, 
depending on the authorization module being used, we cannot guarantee user "uniqueness" across all organizations, in our current situation there might be 2 owners "foo" on 2 different orgs, thus we need to also check that the user matches a group.

I think we might also want to enforce the auth modules to prefix the owner string with their provider ids, or the origin group.

src/ctia/stores/es/crud.clj
```diff
@@ -37,47 +47,130 @@
...
+              (set (:authorized_groups doc)))))))
+
+(s/defn authorized-user? :- s/Bool
+  [doc ident]
+  (boolean
+   (and
+    (not (nil? (:login ident)))
+    (seq (:authorized_users doc))
+    (seq (clojure.set/intersection
+          #{(:login ident)}
```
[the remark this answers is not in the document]
gbuisson: I'm not sure I fully get your question but:

```clojure
ctia.main=> (set "msprunck@cisco.com")
#{\@ \c \i \k \m \. \n \o \p \r \s \u}
ctia.main=> (set ["msprunck@vigiglobe.com"])
#{"msprunck@vigiglobe.com"}
```

src/ctia/stores/es/query.clj
```diff
@@ -2,6 +2,34 @@
   (:require [clj-momo.lib.es.query :as q]
             [clojure.string :as str]))
 
+(def public-tlps
```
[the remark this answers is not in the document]
gbuisson: perhaps making a TLP dedicated ns, will try that.

src/ctia/schemas/core.clj
gbuisson: Actually we started the private-ctia branch long before those improvements, after we merge that, there's going to be a new phase were we merge back with CTIA main branch.
We should fix that at this step because everything will be broken anyway.

project.clj
```diff
@@ -156,7 +156,7 @@
                                    [org.clojure/test.check "0.9.0"]
                                    [com.gfredericks/test.chuck "0.2.7"]
                                    [prismatic/schema-generators "0.1.0"]]
-                    :pedantic? :abort
+                    :pedantic? :warn
```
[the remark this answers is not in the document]
gbuisson: Yes I mistakenly committed that.
I have some difficulties with this because of my lein profile...

src/ctia/domain/access_control.clj
```diff
@@ -0,0 +1,73 @@
...
+     (owner? current-doc ident) true
+
+     ;;or if user is listed in authorized_users field
+     (authorized-user? current-doc ident) true
+
+     ;;or if one of the users groups are in the authorized_groups field
+     (authorized-group? current-doc ident) true
+
+     ;; CTIM models with TLP green that is owned by org BAR
+     (and (some #{(:tlp current-doc)} public-tlps)
```
yogsototh: why not `(contains? public-tlps (:tlp current-doc))`

src/ctia/http/routes/coa.clj
```diff
@@ -87,7 +111,9 @@
                 :path-params [id :- s/Str]
                 :header-params [{Authorization :- (s/maybe s/Str) nil}]
                 :capabilities :read-coa
-                (if-let [coa (read-store :coa (fn [s] (read-coa s id)))]
+                :identity identity
```
yogsototh: I would prefer not to override a core fn name here. I would suggest `id-record`.

edit: I saw that `identity` is used in other routes. So that is not something to do in this PR. But I believe it would be nice to keep in mind we should make that change one day.

src/ctia/domain/access_control.clj
```diff
@@ -0,0 +1,73 @@
...
+
+(s/defn authorized-user? :- s/Bool
+  [doc ident]
+  (boolean
+   (and
+    (not (nil? (:login ident)))
+    (seq (:authorized_users doc))
+    (seq (clojure.set/intersection
+          #{(:login ident)}
+          (set (:authorized_users doc)))))))
```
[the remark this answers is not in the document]
gbuisson: see: https://stackoverflow.com/questions/3249334/test-whether-a-list-contains-a-specific-value-in-clojure

src/ctia/http/routes/coa.clj
gbuisson: I'm not going to fix coding style issues from previous feature PRs into this one, I keep that in mind but I won't make the change now.

PR #588, opened 2017-09-06
Kafka

quoll:
Adds a new Event Hook to publish events to a Kafka topic


src/ctia/flows/hooks/kafka.clj
```diff
@@ -0,0 +1,42 @@
...
+(defrecord KafkaPublisher [producer topic partition opts]
+  Hook
+  (init [_]
+    :nothing)
+  (destroy [_]
+    (.close producer))
+  (handle [_ event _]
+    (let [event-json (json/encode event)]
+      (c/retry-exp
+       "send to Kafka"
```
msprunck: The `send-async` function will return immediately once the record has been stored in the buffer. So I don't think that the `retry-exp` is useful here.

src/ctia/flows/hooks/kafka.clj
```diff
@@ -0,0 +1,42 @@
...
+       (kafka/send-async! producer topic partition production-key event-json opts)))))
+
+(defn new-publisher
+  ([]
+   (let [{:keys [enabled host port topic partition security truststore keystore password]}
+         (get-in @properties [:ctia :hook :kafka])]
+     (when enabled
+       (log/info "Configuring Kafka publishing")
+       (let [config (cond-> {:bootstrap.servers (str host ":" port)}
```
msprunck: In an HA env, you often have at least 3 brokers. You should provide a configuration with a list of broker. Another option would be to retrieve the brokers automatically via zookeeper using [franzy-admin](https://github.com/ymilky/franzy-admin/blob/master/src/franzy/admin/cluster.clj#L19).

src/ctia/flows/hooks/kafka.clj
quoll: Good point. It came from having differing ideas at different times, and not merging them correctly.
I don't want to block, since I don't know what this effect this has on the rest of CTIA. But I'm also loathe to fire-and-forget with a `send-async!`.
What I'd like to do here is keep the retry, change to a blocking `send`, and wrap it all in a future. Are you aware of problems with that?

msprunck: No it should work.

src/ctia/flows/hooks/kafka.clj
```diff
@@ -0,0 +1,42 @@
...
+  Hook
+  (init [_]
+    :nothing)
+  (destroy [_]
+    (.close producer))
+  (handle [_ event _]
+    (let [event-json (json/encode event)]
+      (c/retry-exp
+       "send to Kafka"
+       (kafka/send-async! producer topic partition production-key event-json opts)))))
```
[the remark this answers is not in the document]
quoll: I'll note for anyone else who looks here that Franzy has conflicting docs for the necessity of specifying the partition.

gbuisson:
This PR looks outdated as the latest commit dates from Sep 2017, I think it would be better to start a new one than keep it as is, let me know if you want to reopen it.

PR #589, opened 2017-09-08
Validating specs on all POST/PUTs, including new string field length limits

### LINKED ISSUE #578: String fields need reasonable size limit
For example, 1024 chars for a title field, 5k for descriptions.  Use HTTP errors to enforce limits.

polygloton:
Closes #578 
See also threatgrid/ctim#160


saintx:
I read through the CTIM PR and this one makes much more sense.   Looks good!

PR #590, opened 2017-09-19
JWT autofill (SwaggerUI, GraphiQL, GraphQL Voyager)

msprunck:
- [x] SwaggerUI: backport the JWT autofill magic from the IROH repo
- [x] GraphiQL: add JWT autofill
- [x] GraphQL Voyager: add JWT autofill

The key used to get the JWT token from the browser local storage is configurable using this property:
```properties
ctia.http.jwt.local-storage-key=:iroh-auth-token
```

I changed the way the API is built by using a function and `api` instead of `defapi` to be able to get conf properties when building API routes (See `ctia.http.handler` and `ctia.http.server`).

### DIFF
--- a/src/ctia/http/server.clj
+++ b/src/ctia/http/server.clj
@@ -42,7 +42,7 @@
     :or {access-control-allow-methods "get,post,put,delete"}}]
   (doto
       (jetty/run-jetty
-       (cond-> #'handler/api-handler
+       (cond-> (handler/api-handler)
 
          access-control-allow-origin
          (wrap-cors :access-control-allow-origin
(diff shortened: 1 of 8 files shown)


src/ctia/http/server.clj
```diff
@@ -42,7 +42,7 @@
     :or {access-control-allow-methods "get,post,put,delete"}}]
   (doto
       (jetty/run-jetty
-       (cond-> #'handler/api-handler
+       (cond-> (handler/api-handler)
```
yogsototh: 👍

PR #603, opened 2017-10-03
Register a shutdown hook at init that invokes shutdown-agents

gbuisson:


saintx:
LGTM

src/ctia/shutdown.clj
```diff
@@ -31,4 +37,4 @@
              (fn [_]
                (.addShutdownHook
                 (Runtime/getRuntime)
-                (Thread. shutdown-ctia!))))))))
+                (Thread. shutdown-ctia-and-agents!))))))))
```
saintx: This makes sense.

polygloton:
I would add `(shutdown-agents)` to the end of `shutdown-ctia!`.  Since we will always want to do that, we might as well hard-code it (and not make it a hook).  Also, that guarantees it goes last, just incase some fool tries to use the pool in a shutdown hook.  :)

gbuisson:
@polygloton I tried that at first, the problem is that it makes the JVM unusable for the next tests.

Issue #605, opened 2017-10-04
E1331: CRUD routes for Attack pattern, Malware and Tool entities


msprunck:
Add new routes for these new CTIM entities:
- Attack pattern
- Malware
- Tool

PR #606, opened 2017-10-04
Attack pattern, Malware and Tool support

msprunck:
Connected to #605 

CTIA API update to support CRUD operations on Attack pattern, malware and tool CTIM entities (depends on threatgrid/ctim#171).

Tasks:

- [x] Add a namespace in `src/ctia/domain/entities` for managing IDs
- [x] Add entity specific capabilities in `ctia.auth`
- [x] Add plumatic schemas for the new entity in `ctia.schemas.core`
- [x] Update the `Bulk` schema in `ctia.schemas.bulk`
- [x] Add a new store Protocol in `ctia.store` 
- [x] Define store operations in a new namespace in `src/ctia/stores/es`
- [x] Add a `realize-<entity>` function in `ctia.domain.entities`
- [x] Update the `bulk` route in `ctia.http.routes.bulk`
- [x] Define HTTP routes for the entity in a new namespace in `src/ctia/http/routes`
- [x] Add new routes to the compojure api handler in `ctia.http.handler`
- [x] Define which default store should be used in the default properties `resources/ctia-default.properties`
- [x] Add a Record that implement the store Protocol in `ctia.stores.es.store`
- [x] Define search params in `ctia.http.routes.common`
- [x] Define sorting fields in `ctia.schemas.sorting`
- [x] Define an ES mapping in `ctia.stores.es.mapping`
- [x] Instanciate the ES store in `ctia.init`
- [x] Update the bulk route test in `ctia.http.routes.bulk-test` (update the bulk size)
- [x] Test entity routes in a new namespace in `test/ctia/http/routes`
- [x] Add entity properties for generative tests in `ctia.http.generative.properties`
- [x] Update generative tests for this entity in `ctia.http.generative.es-store-spec`
- [x] Define which store should be used for tests in `ctia.test-helpers.es/fixture-properties:es-store`


resources/ctia-default.properties
```diff
@@ -49,11 +50,13 @@ ctia.store.identity=es
...
 ctia.store.indicator=es
 ctia.store.judgement=es
+ctia.store.malware=es
 ctia.store.relationship=es
 ctia.store.sighting=es
-ctia.store.ttp=es
+ctia.store.tool=es
 
 ctia.store.es.actor.indexname=ctia_actor
+ctia.store.es.attack-pattern.indexname=ctia_atttack-pattern
```
saintx: I'd use an underscore, and only 2 `t`'s here.

PR #609, opened 2017-10-11
initial CTIA ES migration task

gbuisson:
This PR introduces a new task to execute an ES index migration, along with migration operations for switching current deployed instances to private CTIA.

- the migration task duplicates indexes instead of modifying documents in place
- transforms are done in fixed batches
- uses ES bulk inserts for performance considerations
- migration operations are composed using clojure transducers

### DIFF
--- a/src/ctia/task/migrate_es_stores.clj
+++ b/src/ctia/task/migrate_es_stores.clj
@@ -1,75 +1,50 @@
 (ns ctia.task.migrate-es-stores
-  (:require
-   [clj-momo.lib.clj-time.core :as time-core]
-   [clj-momo.lib.clj-time.coerce :as time-coerce]
-   [clj-momo.lib.es
-    [document :as es-doc]
-    [index :as es-index]]
-   [clojure.string :as string]
-   [clojure.tools.logging :as log]
-   [ctia
-    [init :refer [init-store-service! log-properties]]
-    [properties :as p]
-    [store :refer [stores]]]))
+  (:require [clj-momo.lib.es
+             [document :as es-doc]
+             [index :as es-index]]
+            [clojure.string :as string]
+            [clojure.tools.logging :as log]
+            [ctia
+             [init :refer [init-store-service! log-properties]]
+             [properties :as p]
+             [store :refer [stores]]]
+            [ctia.task.migrations :refer [available-migrations]]))
 
 (def default-batch-size 100)
 
-(def add-groups
-  "set a document group to [\"tenzin\"] if unset"
-  (map (fn [{:keys [groups]
-            :as doc}]
-         (if-not (seq groups)
-           (assoc doc :groups ["tenzin"])
-           doc))))
+(defn compose-migrations
+  "compose migrations from a list of keywords"
+  [migration-keys]
+  (let [migrations
+        (vals (select-keys available-migrations
+                           migration-keys))]
+    (if (seq migrations)
+      (apply comp migrations)
+      (do (log/warn "target migration not found, copying data")
+          (map identity)))))
 
-(def fix-end-time
-  "fix end_time to 2535"
-  (map
-   (fn [{:keys [valid_time]
-        :as doc}]
-     (if (:end_time valid_time)
-       (update-in doc
-                  [:valid_time
-                   :end_time]
-                  #(let [max-end-time (time-core/internal-date 2525 01 01)
-                         end-time (time-coerce/to-internal-date %)]
-                     (if (time-core/after? end-time max-end-time)
-                       max-end-time
-                       end-time)))
-       doc))))
-
-(def available-operations
-  "define all migration operations here"
-  {:__test (map #(assoc % :groups ["migration-test"]))
-   :default (map identity)
-   :add-groups add-groups
-   :fix-end-time fix-end-time})
-
-(defn compose-operations
-  "compose operations from a list of keywords"
-  [operation-keys]
-  (let [operations (vals (select-keys available-operations
-                                      operation-keys))]
-    (apply comp
-           (:default available-operations)
-           operations)))
+(defn prefixed-index [index prefix]
+  (let [version-trimmed (string/replace index #"^v\w+.\w+.\w_" "")]
+    (str "v" prefix "_" version-trimmed)))
 
 (defn source-store-map->target-store-map
   "transform a source store map into a target map,
   essentially updating indexname"
   [store prefix]
-  (update store :indexname #(str "v" prefix "_" %)))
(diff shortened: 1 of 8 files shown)


src/ctia/task/migrate_es_stores.clj
```diff
@@ -0,0 +1,204 @@
...
+(defn -main
+  "invoke with lein run -m ctia.task.migrate-es-stores <suffix> <operations> <batch-size> <confirm?>"
+  [suffix ops batch-size confirm?]
+
+  (assert suffix
+          "Please provide an indexname suffix for target store creation")
+  (assert ops "Please provide a csv operation list")
+  (assert batch-size "Please specify a batch size")
+
+  (log/warn "migrating all ES Stores")
```
msprunck: Why use the warn log level everywhere? It's a manual operation, info should be sufficient.

test/ctia/task/migrate_es_stores_test.clj
```diff
@@ -0,0 +1,103 @@
...
+    (testing "simulate migrate es indexes"
+      (sut/migrate-store-indexes "foo"
+                                 "add-groups,fix-end-time"
+                                 100
+                                 false))
+    (testing "migrate es indexes"
+      (sut/migrate-store-indexes "foo"
+                                 "add-groups,fix-end-time"
+                                 100
+                                 true))))
```
msprunck: Could you check that a new index has been created and that all docs have been fixed?

src/ctia/task/migrate_es_stores.clj
```diff
@@ -0,0 +1,229 @@
...
+  (map
+   (fn [{:keys [valid_time]
+        :as doc}]
+     (if (:end_time valid_time)
+       (update-in doc
+                  [:valid_time
+                   :end_time]
+                  #(string/replace %
+                                   #"2535"
+                                   "2525"))
```
saintx: This will make it compliant with the song lyrics.

src/ctia/task/migrate_es_stores.clj
```diff
@@ -0,0 +1,229 @@
...
+           operations)))
+
+(defn source-store-map->target-store-map
+  "transform a source store map into a target map,
+  essentially updating indexname"
+  [store prefix]
+  (update store :indexname #(str prefix "_" %)))
+
+(defn store->map
+  "transfrom a store record
```
saintx: s/transfrom/transform

src/ctia/task/migrate_es_stores.clj
```diff
@@ -0,0 +1,229 @@
...
+     (if (:end_time valid_time)
+       (update-in doc
+                  [:valid_time
+                   :end_time]
+                  #(string/replace %
+                                   #"2535"
+                                   "2525"))
+       doc))))
+
+(def available-operations
```
saintx: I think we should be able to add more "jobs" or "operations" like this over time, perhaps in their own namespace, and when we successfully apply those jobs over a data store in a given environment (int/test/prod) we can make a record somewhere that we applied that job/operation on that environment at a given datetime.  This would allow us to very easily run the data migration "orchestrator/runner" (for lack of a better term off the top of my head) on a given environment, and it would retroactively apply *all* of the jobs/operations that had not yet been applied in that environment.

For example, I need to do a migration soon to deduplicate indicators.  It's important for when we start making mappings between indicators and Tool, Malware, Vulnerability, and Attack Pattern objects that we don't have lots of duplicate indicators.  This means finding all duplicates, designating one as the original, updating all relationship mappings that point at the duplicates so they point to the originals, and then remove the duplicates.  This is just *one* example of migration jobs we're going to need to add over time, and the jobs will need to be applied cumulatively across multiple environments.  It would be great to have a structure we could extend over time to define these jobs and to record their application in a given environment.

test/ctia/task/migrate_es_stores_test.clj
```diff
@@ -0,0 +1,210 @@
...
+
+(deftest fix-end-time-test
+  (is (= (transduce sut/fix-end-time conj [{}]) [{}]))
+  (is (= (transduce sut/fix-end-time conj [{:valid_time
+                                            {:start_time "foo"}}])
+         [{:valid_time
+           {:start_time "foo"}}]))
+  (is (= (transduce sut/fix-end-time conj
+                    [{:valid_time
+                      {:end_time "2535"}}])
```
yogsototh: I think the test should use a more realistic `#inst` here instead of the string "2535".

src/ctia/task/migrate_es_stores.clj
```diff
@@ -0,0 +1,229 @@
...
+
+(defn -main
+  "invoke with lein run -m ctia.task.migrate-es-stores <prefix> <operations> <batch-size> <confirm?>"
+  [prefix ops batch-size confirm?]
+
+  (assert prefix "Please provide an indexname suffix for target store creation")
+  (assert ops "Please provide a csv operation list")
+  (assert batch-size "Please specify a batch size")
+  (log/info "migrating all ES Stores")
+  (setup)
```
yogsototh: I would parse the "ops" to a list of keyword before providing them directly to `migrate-store-indexes`.

src/ctia/task/migrate_es_stores.clj
```diff
@@ -0,0 +1,229 @@
...
+  {:__test (map #(assoc % :groups ["migration-test"]))
+   :default (map identity)
+   :add-groups add-groups
+   :fix-end-time fix-end-time})
+
+(defn compose-operations
+  "compose operations from a list of keywords"
+  [ops]
+  (let [operation-keys (map keyword
+                            (clojure.string/split ops #","))
```
yogsototh: I wouldn't "parse" a string here and `ops` should be a list of keywords and parsed before (see other comment in `-main`).

src/ctia/task/migrate_es_stores.clj
```diff
@@ -0,0 +1,206 @@
...
+  (let [migrations
+        (vals (select-keys available-migrations
+                           migration-keys))]
+    (if (seq migrations)
+      (apply comp migrations)
+      (do (log/warn "target migration not found, copying data")
+          (map identity)))))
+
+(defn prefixed-index [index prefix]
+  (let [version-trimmed (string/replace index #"^v\w+.\w+.\w_" "")]
```
yogsototh: I think this regex is slightly too risky. Why not: `#"^v[^_]*_"` to remove everything up to the first underscore that shouldn't be part of any semantic version number.

In your regex the `.` are considered as "any char" and not as dot.

PR #618, opened 2017-10-24
IROH 905: Investigation persistence

polygloton:
This is just a beginning of an effort to add an `Investigation` entity to CTIA.  We are still defining the schema, so it is intentionally open-ended in this PR.


polygloton:
Need changes in threatgrid/ctim#177 to be deployed, and CTIM version updated, before merging

test/ctia/http/routes/investigation_test.clj
```diff
@@ -0,0 +1,138 @@
...
+  (helpers/set-capabilities! "foouser"
+                             ["foogroup"]
+                             "user"
+                             all-capabilities)
+  (whoami-helpers/set-whoami-response "45c1f5e3f05d0"
+                                      "foouser"
+                                      "foogroup"
+                                      "user")
+
+  (testing "POST /ctia/attack-pattern"
```
polygloton: This string is wrong

src/ctia/stores/es/store.clj
```diff
@@ -294,4 +294,4 @@
     (inv/handle-list state filtermap ident params))
   IQueryStringSearchableStore
   (query-string-search [_ query filtermap ident params]
-    (tool/handle-query-string-search state query filtermap ident params)))
+    (inv/handle-query-string-search state query filtermap ident params)))
```
polygloton: @gbuisson Gets credit for finding this stupid bug.  I did a bad cut-and-paste.  :(

src/ctia/stores/es/mapping.clj
```diff
@@ -635,6 +635,16 @@
       :tokenizer "keyword"
       :type "custom"}}}})
 
+(def investigation-mapping
```
polygloton: I ask that reviewers pay special attention to this mapping.  The goals is that we can add anything to an Investigation so that we can postpone defining a schema.  I think this should work, but I'm not 100% sure.

polygloton:
A lot of boiler-plate code is exposed by this change.  We should make an effort to clean this up.

src/ctia/stores/es/mapping.clj
gbuisson: perhaps you just want to store and not make dynamic data available for search.
you want to make common entity fields like `:description` searchable.
In this case I suggest putting all that data on a top level field and disable indexing for it:

https://www.elastic.co/guide/en/elasticsearch/reference/current/enabled.html

project.clj
```diff
@@ -42,7 +42,7 @@
                  ;; Schemas
                  [prismatic/schema ~schema-version]
                  [metosin/schema-tools ~schema-tools-version]
-                 [threatgrid/ctim "0.4.21"]
+                 [threatgrid/ctim "0.4.22-SNAPSHOT"]
```
gbuisson: I thought that `Investigation` was not supposed to make it on CTIM  and only defined on CTIA?

polygloton: We will need to use it on the UI too, so I think CTIM makes sense.

src/ctia/stores/es/mapping.clj
polygloton: This is a good suggestion.  I talked it over with Craig, and we decided to just set `:dynamic = false`.  The [doc](https://www.elastic.co/guide/en/elasticsearch/reference/current/dynamic.html) says that, "Newly detected fields are ignored. These fields will not be indexed so will not be searchable but will still appear in the _source field of returned hits."  This seems like a reasonable compromise.

project.clj
polygloton: Though it would be easy to move to CTIA if that's what we really want. I'm not sure how "pure" CTIM really needs to be.  I'll go out on a limb and say I think it's not important.

PR #625, opened 2017-11-22
Let pagination headers pass throug CORS

yogsototh:
Relates to: https://github.com/threatgrid/iroh/issues/1536

### DIFF
--- a/resources/ctia-default.properties
+++ b/resources/ctia-default.properties
@@ -13,6 +13,9 @@ ctia.http.enabled=true
 ctia.http.port=3000
 ctia.http.access-control-allow-origin=http://(localhost|127.0.0.1)(:\\d+)?
 ctia.http.access-control-allow-methods=get,post,put,delete
+ctia.http.access-control-allow-headers=x-total-hits,x-next,x-previous,x-sort,etag
+ctia.http.access-control-expose-headers=x-total-hits,x-next,x-previous,x-sort,etag
+
 ctia.http.jwt.enabled=true
 ctia.http.jwt.public-key-path=resources/cert/ctia-jwt.pub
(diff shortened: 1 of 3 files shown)


src/ctia/http/server.clj
```diff
@@ -26,11 +26,17 @@
             (split origins-str #","))))
 
 
+(defn- str->vec-of-keywords
+  "take a string with words separated with commas, returns a vec of keywords"
+  [str]
+  (vec (map keyword (split methods-str #","))))
```
msprunck: You could use `mapv` instead of `(vec (map`

resources/ctia-default.properties
```diff
@@ -13,6 +13,8 @@ ctia.http.enabled=true
 ctia.http.port=3000
 ctia.http.access-control-allow-origin=http://(localhost|127.0.0.1)(:\\d+)?
 ctia.http.access-control-allow-methods=get,post,put,delete
+ctia.http.access-control-allow-headers=server,date,content-type,connection,x-total-hits,x-next,x-sort,etag
```
msprunck: CORS are from the client POV. Why we need to add server, date and connection to this list?

gbuisson: also add x-previous

gbuisson:
the build fails with:

Exception in thread "main" java.lang.RuntimeException: Unable to resolve symbol: methods-str in this context, compiling:(ctia/http/server.clj:32:21)

likely due to a missing require

gbuisson:
perhaps Expose-Headers is more suitable to our case:

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers

PR #631, opened 2017-12-11
Investigation GraphQL endpoint

gbuisson:
As Investigations could turn to be very large records, I Added an Investigation GraphQL endpoint.
This should provide more efficient Investigation listing capabilities and significantly reduce Investigation API list load times.


gbuisson:
As that change would provide mostly a network based impact I'm not certain benchmarking that on my computer would be relevant, I think I should instead work on adding source filtering support.

Issue #628, opened 2017-12-11
Epic: CTIA Bundle API Endpoint


saintx:
Build a new bundle API endpoint in CTIA that performs a "graph merge" of objects and relationships into the existing object and relationship graph in CTIA.

### [Issue 629: CTIA Bundle API Endpoint](https://github.com/threatgrid/ctia/issues/629)
1. CTIA has a bundle API endpoint, which accepts a [bundle object](https://github.com/threatgrid/ctim/blob/master/src/ctim/schemas/bundle.cljc).
2. Don't permit anyone other than us to use the bundle API endpoint. This is an experimental feature not ready to be shared with the public.
3. Objects in the bundle may use transient UUIDs, which CTIA converts to real UUIDs before adding them to storage. (This relates to [Bug #610: Transient IDs should not be stored.](https://github.com/threatgrid/ctia/issues/610))
4. The response of the bundle API endpoint contains a map of the transient UUIDs to the real UUIDs along with a brief status message for each item in the bundle. Status messages indicate whether a new object was created, an existing object was updated, or the system failed to store the object.

### [Issue 630: Bundle Upserts](https://github.com/threatgrid/ctia/issues/630)
CTIA handles upserts seamlessly.
   - This mechanism enforces prefix key constraints on specified external ID prefixes.
   - The external ID prefixes that enforce key constraints can be defined in the CTIA configuration.
   - Key constraint patterns are tied to specific organizations, so users outside of that org cannot use those key prefix patterns to update documents.
   - XID constraints update if the XID exists, or create a new one if the XID doesn't already exist, if the XID prefix enforces a uniqueness constraint.

PR #634, opened 2017-12-18
Bundle import

### LINKED ISSUE #629: CTIA bundle API endpoint
CTIA supports a bundle API endpoint, which allows users with the correct permissions to POST a CTIM [bundle object](https://github.com/threatgrid/ctim/blob/master/src/ctim/schemas/bundle.cljc).

## Key Results

- [x] By posting a bundle in this way, CTIA performs a "graph merge" on the input bundle.
- [x] The ability to post bundles must be controlled via permissions, while we test this feature, and updates to the deployment configuration should be submitted. This is an experimental beta feature not ready to be enabled for all users by default.
- [x] Objects in the bundle may use transient UUIDs, which CTIA converts to real UUIDs before adding them to storage. (This relates to [Bug #610: Transient IDs should not be stored.](https://github.com/threatgrid/ctia/issues/610))
- [x] The response of the bundle API endpoint contains a map of the transient UUIDs to the real UUIDs along with a brief status message for each item in the bundle. Status messages indicate whether a new object was created, an existing object was updated, or the system failed to store the object.

msprunck:
Epic #628

Closes #629 

This PR add a bundle API endpoint, which allows users with the correct permissions to POST a CTIM bundle object. It's a beta feature which is not visible in the default Swagger UI. See the [documentation](https://github.com/msprunck/ctia/blob/b320ff5c0244238c2bdac3622f529edc877cee71/README.md#bundle-import) for more details.

ctia-issue-634

- [x] [Documentation update](https://github.com/msprunck/ctia/blob/b320ff5c0244238c2bdac3622f529edc877cee71/README.md#bundle-import)
- [x] Existing entities lookup (based on external_ids)
- [x] Bulk entities creation with no duplicates 
- [x] Error handling with partial results


src/ctia/http/routes/bundle.clj
```diff
@@ -0,0 +1,326 @@
...
+  "Index entities by external_id
+
+   Ex:
+   {{:external_id \"ctia-1\"} {:external_id \"ctia-1\"
+                               :entity {...}}
+    {:external_id \"ctia-2\"} {:external_id \"ctia-2\"
+                               :entity {...}}}"
+  [entities]
+  (let [entity-with-external-id
+        (reduce (fn [acc {:keys [external_ids] :as entity}]
```
gbuisson: I think you could get rid of the `reduce` and replace it with `map`

src/ctia/stores/es/crud.clj
```diff
@@ -40,6 +41,44 @@
...
+  [{:index {:_id \"1\"}}
+   {:index {:_id \"2\"}}]
+
+  ->
+
+  [{:_id \"1\"}
+   {:_id \"2\"}]
+  "
+  [items]
+  (map (comp val first) items))
```
yogsototh: I found it hard to read because I'm really not used to `val` and also I generally never user `first` directly on a map to transform a single element hash-map as a list of two elements [k v].

Would you mind to use `(map (comp first vals) items)` instead?

src/ctia/flows/crud.clj
```diff
@@ -182,7 +183,8 @@
     (assoc fm
            :events
            (doall
-            (for [entity entities]
+            (for [{:keys [error] :as entity} entities
```
gbuisson: perhaps it's a good time to replace that with `doseq`?

msprunck: I can not use `doseq` here, it returns nil

src/ctia/http/routes/bundle.clj
```diff
@@ -0,0 +1,326 @@
...
+
+(defn transient-id?
+  [id]
+  (and id (re-matches id/transient-id-re id)))
+
+(def default-external-key-prefixes "ctia-")
+
+(defn debug [msg v]
+  (log/info msg v)
+  v)
```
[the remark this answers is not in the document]
msprunck: Yes it should by `log/debug`

saintx:
Is the comma separated list of external ID key prefixes required, or optional?

msprunck:
The HTTP param `external-key-prefixes` is optional. The bundle importer will use the configured prefixes (`:external-key-prefixes` property). If nothing matches, new entities are created.

PR #637, opened 2018-01-05
Updated dependencies with minor version updates

quoll:
Updated all minor dependency releases.

Composure-api also has a minor update, but it breaks the build.


src/ctia/http/middleware/unknown.clj
```diff
@@ -0,0 +1,18 @@
...
+      "<body> <h2>HTTP ERROR: 404</h2> <p>Problem accessing "
+       uri
+       ". Reason: <pre>    Not Found</pre></p> </body> </html>"))
+
+(defn wrap-unknowns
+  "When a request results in a nil response, then this indicates that no route handled it."
+  [handler]
+  (fn [request]
+    (or (handler request)
+        (not-found (err-html (:uri request))))))
```
msprunck: Compojure API provides a ["not-found" handler](https://github.com/metosin/compojure-api/wiki/Routing#not-found-at-runtime). I think it's a better option that respect response format. I don't know if it applies to the dependency that requires this change.

quoll: I removed the middleware, and added a `compojure.route/not-found` path as suggested.

I have kept the unknown.clj file (along with the now-unused `wrap-unknowns` wrapper), as this contains the HTML to be returned. A reason to keep the `wrap-unknowns` function is if we ever want to have HTML that looks like the previous version returned by ring/compojure. The old 404 response included the requested URI, but the `compojure.route/not-found` path does not get access to this.

quoll:
If this passes review, then it can be merged without modification, just as soon as CTIM, Flanders, and clj-momo are all merged.

src/ctia/http/middleware/unknown.clj
```diff
@@ -0,0 +1,22 @@
+(ns ctia.http.middleware.unknown
+  (:require [ring.util.response :refer [not-found]]))
+
```
msprunck: Could you add comment to explain why this middleware is added?

quoll: As stated in the comment above:

"I have kept the unknown.clj file (along with the now-unused wrap-unknowns wrapper), as this contains the HTML to be returned. A reason to keep the wrap-unknowns function is if we ever want to have HTML that looks like the previous version returned by ring/compojure. The old 404 response included the requested URI, but the compojure.route/not-found path does not get access to this."

gbuisson:
LGTM!

PR #644, opened 2018-01-31
Use the right authentication identity data structure

### LINKED ISSUE #643: Bundle imports creating duplicate documents in ES
Repeatedly posting the same bundle to my local `/ctia/bundle/import` API endpoint is causing duplicate entities to be created in ES.

The `bundle` object has the exact same information in it each time I import it.  Identical transient IDs,  timestamps, and external IDs.  The documents created in ElasticSearch are identical in every way, *except* for their `:modified_time`.

My understanding is that repeatedly importing the same object should not create duplicates.  Is there an additional configuration step required to prevent my local instance of CTIA from creating duplicate entries?

msprunck:
Closes #643 

Two data structure are used for authentication identity in CTIA:
- An `IIdentity` protocol (used by the create and update flows)
- A map `{:login login :groups groups}` (used by stores)

The bundle importers was only using the map form.

Changes:
- The `ident->map` helper has been moved to `ctia.auth`
- Usage of the appropriate form in the bulk and bundle importer routes

### DIFF
--- a/src/ctia/auth.clj
+++ b/src/ctia/auth.clj
@@ -1,4 +1,5 @@
-(ns ctia.auth)
+(ns ctia.auth
+  (:require [schema.core :as s]))
 
 (defprotocol IIdentity
   (authenticated? [this])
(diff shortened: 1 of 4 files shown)


hshahman:
Verified in rel-20180312 v0.2

PR #646, opened 2018-02-06
Fix duplicates when submitting bundles

### LINKED ISSUE #645: CTIA ignoring external-key-prefixes query parameter
CTIA appears to be ignoring my external-key-prefixes query param and creating duplicate objects from identical data, ignoring the external ID.  In some cases, it is giving the objects their own UUID.  In other cases, it is creating records in the store for the transient IDs.

See notes below for detail.

msprunck:
Closes #645 

- [x] Use the ElasticSearch wait_for param for the bundle import endpoint (See https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-refresh.html)
When a bundle is imported there are three steps:
  * All entities except relationships are created and stored
  * The source_ref and the target_ref in the relationships are set with entities final IDs
  * Relationships are created and stored (The wait_for param is only used at this step)

- [x] Paginate results when searching entities by external_ids
- [x] Display Swagger UI for the bundle routes

### DIFF
--- a/src/ctia/http/routes/bundle.clj
+++ b/src/ctia/http/routes/bundle.clj
@@ -298,6 +298,10 @@
              #(dissoc % :new-entity :old-entity)
              (apply concat (vals bundle-import-data)))})
 
+(defn bulk-params []
+  {:refresh
+   (get-in @properties [:ctia :store :bundle-refresh] false)})
+
 (s/defn import-bundle :- BundleImportResult
   [bundle :- NewBundle
    external-key-prefixes :- (s/maybe s/Str)
@@ -312,7 +316,7 @@
                             (entities-import-data->tempids entities-import-data)))
                      (apply merge {}))]
     (debug "Import bundle response"
-           (->> (bulk/create-bulk bulk tempids auth-identity)
+           (->> (bulk/create-bulk bulk tempids auth-identity (bulk-params))
                 (with-bulk-result bundle-import-data)
                 build-response))))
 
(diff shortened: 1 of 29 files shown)


src/ctia/http/routes/bundle.clj
```diff
@@ -113,17 +113,29 @@
...
+               (let [{results :data
+                      {next-page :next} :paging}
+                     (read-store entity-type (list-fn entity-type)
+                                 {:external_ids external-ids}
+                                 (auth/ident->map auth-identity)
+                                 paging)
+                     acc-entities (into entities results)]
+                 (if next-page
+                   (recur next-page acc-entities)
+                   acc-entities))))
```
yogsototh: As loop are not easy to read, would you mind to write an helper?

Something similar to:

```clojure
(defn all-pages [fn]
   (loop [paging {:offset 0 :limit find-by-external-ids-limt}
              entities []]
              (let [{results :data {next-page :next} :paging} (fn paging) acc-entities (into entities results]
                  (if next-page
                         (recur next-page acc-entities)
                         acc-entities)))
```

saintx:
Still testing this locally.  I can't get anywhere near 1000 objects per bundle to actually import.  I think the request body might be truncated long before that, causing schema validation failures.  But it seems to work if I limit myself to ~100 objects per bundle, or thereabouts.

hshahman:
Verified in rel-20180312 v0.2

PR #647, opened 2018-02-08
Scratchpad Entity

gbuisson:
Add CTIM Scratchpad entity support as well as a new `PATCH` flow intended for partial updates
- [x] Store
- [x] Basic CRUD routes
- [x] Atomic operation routes
- [x] Graphql
- [x] migration support

Misc: 
- added support to migrate `Investigation`entities
- added `abstraction_level` to `AttackPattern`entity mapping 
- Fixed Swagger page title

### DIFF
--- a/src/ctia/flows/crud.clj
+++ b/src/ctia/flows/crud.clj
@@ -300,20 +300,28 @@
     :delete (first results)
     :update (first entities)))
 
-(defn recast [orig-coll new-coll]
+(defn recast
+  "given a source collection and the target one,
+   cast target the same as source"
+  [orig-coll new-coll]
   (cond
     (vector? orig-coll) (vec new-coll)
     (set? orig-coll) (set new-coll)
     :else new-coll))
 
 (defn add-colls [& args]
+  "given many collections as argument
+   concat them keeping the first argument type"
   (let [new-coll
         (->> args
              (map #(or % []))
              (apply into))]
     (recast (first args) new-coll)))
 
-(defn remove-colls [& args]
+(defn remove-colls
+  "given many collections as argument
+   remove items on a from b successively"
+  [& args]
   (let [new-coll
         (reduce
          (fn [a b]
@@ -321,7 +329,10 @@
                    (or a []))) args)]
     (recast (first args) new-coll)))
 
-(defn replace-colls [& args]
+(defn replace-colls
+  "given many collections as argument
+   replace a from b successively"
+  [& args]
   (let [new-coll (last args)]
     (recast (first args) new-coll)))
 
@@ -394,7 +405,6 @@
              realize-fn
              update-fn
              entity-id
-             partial-entity
              identity
              entity
              long-id-fn
@@ -404,7 +414,6 @@
          :entity-type entity-type
          :entities [entity]
          :prev-entity prev-entity
-         :partial-entity partial-entity
          :identity identity
          :long-id-fn long-id-fn
          :realize-fn realize-fn
--- a/src/ctia/flows/crud.clj
+++ b/src/ctia/flows/crud.clj
@@ -1,7 +1,7 @@
 (ns ctia.flows.crud
   "This namespace handle all necessary flows for creating, updating
   and deleting entities."
-  (:require [clj-momo.lib.map :refer [deep-merge]]
+  (:require [clj-momo.lib.map :refer [deep-merge-with]]
             [clojure.spec.alpha :as cs]
             [clojure.tools.logging :as log]
             [ctia
@@ -376,7 +376,6 @@
 (defn patch-flow
   "This function centralizes the patch workflow.
   It is helpful to easily add new hooks name
-
   To be noted:
     - `:before-update` hooks can modify the entity stored.
     - `:after-update` hooks are read only"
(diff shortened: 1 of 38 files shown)


test/ctia/http/routes/scratchpad_test.clj
```diff
@@ -0,0 +1,287 @@
...
+      (testing "POST /ctia/scratchpad/:id/observables :add"
+        (let [new-observables [{:type "ip" :value "42.42.42.42"}]
+              response (post (str "ctia/scratchpad/" (:short-id scratchpad-id) "/observables")
+                             :body {:operation :add
+                                    :observables new-observables}
+                             :headers {"Authorization" "45c1f5e3f05d0"})
+              updated-scratchpad (:parsed-body response)]
+          (is (= 200 (:status response)))
+          (is (subset? (set new-observables)
+                       (set (:observables updated-scratchpad))))))
```
msprunck: I suggest adding a test to check that `new-observables` is not equal to `(:observables updated-scratchpad)`.

Same comment for all `:add` operations on bundle entities and texts.

src/ctia/flows/crud.clj
```diff
@@ -296,6 +300,49 @@
...
+                   :remove remove-colls
+                   :replace replace-colls
+                   replace-colls)
+        entity (-> (deep-merge-with patch-fn
+                                    prev-entity
+                                    partial-entity)
+                   un-store
+                   (dissoc :id))]
+    (assoc fm :entities [entity])))
+
```
yogsototh: All those functions are really "data manipulation" about things without types or form.
Its hard to follow. Also I couldn't find any test showing how to use them and what to expect from them individually.

Would you mind to add a docstring with example and if possible unit test for each of these functions with simple examples? Typically for `add-colls`, `remove-colls` and `replace-colls`.

src/ctia/http/routes/common.clj
```diff
@@ -398,6 +401,28 @@
...
+  {(s/optional-key :fields) [scratchpad-sort-fields]})
+
+(s/defschema ScratchpadSearchParams
+  (st/merge
+   PagingParams
+   BaseEntityFilterParams
+   SourcableEntityFilterParams
+   ScratchpadFieldsParam
+   {:query s/Str
+    (s/optional-key :texts.text) s/Str
```
yogsototh: I know that certainly "works™" right now. But this is "dangerous" from:

https://clojure.org/reference/reader

> Keywords - Keywords are like symbols, except:
> ... they cannot contain '.' or name classes.

So the solution might be to use strings instead of keywords for query if possible and easy to do. Certainly something to point in another technical issue. I imagine the problem is the same for all other entities.

test/ctia/http/generative/es_store_spec.clj
```diff
@@ -56,3 +56,6 @@
 
 (defspec ^:generative api-for-tool-routes-es-store
   prop/api-for-tool-routes)
+
+#_(defspec ^:disabled api-for-scratchpad-routes-es-store
+    prop/api-for-scratchpad-routes)
```
yogsototh: You should either remove it before merging or use a `;; TODO: fix it` comment + file an issue not to forget it.

test/ctia/http/generative/properties.clj
```diff
@@ -136,3 +142,7 @@
 (def api-for-tool-routes
   (api-for-route 'tool
                  (spec-gen "max-new-tool")))
+
+#_(def api-for-scratchpad-routes
```
yogsototh: same here, see previous comment.

test/ctia/http/routes/actor_test.clj
```diff
@@ -92,20 +76,7 @@
...
-                :actor_type "Hacker",
-                :title "actor",
-                :confidence "High",
-                :source "a source"
-                :valid_time {:start_time #inst "2016-02-11T00:40:48.212-00:00"
-                             :end_time #inst "2016-07-11T00:40:48.212-00:00"}
-                :schema_version schema-version
-                :tlp "green"}
-               actor))))
+               (assoc new-actor :id (id/long-id actor-id)) actor))))
```
yogsototh: so much better!

test/ctia/http/routes/scratchpad_test.clj
```diff
@@ -0,0 +1,287 @@
...
+                                            :name "TEST"
+                                            :labels ["malware"]}}}
+              response (post (str "ctia/scratchpad/" (:short-id scratchpad-id) "/bundle")
+                             :body {:operation :replace
+                                    :bundle bundle-entities}
+                             :headers {"Authorization" "45c1f5e3f05d0"})
+              updated-scratchpad (:parsed-body response)]
+          (is (= 200 (:status response)))
+          (is (= (:malwares bundle-entities)
+                 (-> updated-scratchpad :bundle :malwares)))))
```
yogsototh: I suggest to add a test that check that only the `:malwares` value was updated.
For example using a GET just before.

~~~
(is (= old-bundle (into old-bundle bundle-entities)))
~~~

I'm not sure how :replace could take advantage of `deep-merging` instead of a simple `into` there. It would be nice to have an example for that specific case that will also help to test that we only replace a specific sub element of the scratchpad.

gbuisson:
I adressed all your feedback, also released `clj-momo` (I forgot about it ;-))

gbuisson:
@msprunck you do that with any of the dedicated routes:

- adding a CTIM Object is about patching the bundle to add an entity so you should use POST /scratchpad/scratchpad-foo/bundle with the entity you wish to add

- a CTIM object id is in fact a Bundle reference, like `actor_refs` for instance you can use the same route is above.

hshahman:
verified in rel-v0.3

PR #653, opened 2018-02-22
CTIM 0.4.28

gbuisson:
⚠️ migration to CTIM 0.4.28 needed after deploying.

depends https://github.com/threatgrid/ctim/pull/200


gbuisson:
@msprunck why would that change?

msprunck:
To sort on `:severity`, `:resolution` or `:internal`. `:severity` is part of the sort fields of the Judgement entity.

gbuisson:
I'm not certain it is needed ATM, at least I have no demand for that, we should add that later if someone need it, else I prefer not to add unneeded features.

project.clj
```diff
@@ -18,7 +18,7 @@
 ;;   valid, and update the exclusions/comments accordingly
 ;; - Maybe you can just delete the dep! (doesn't hurt to check)
 
-(defproject ctia "1.0.0-rc1"
+(defproject ctia "1.1.0"
```
yogsototh: You can't depend on a SNAPSHOT version while having a non SNAPSHOT version here.

test/ctia/task/migrations_test.clj
```diff
@@ -28,3 +28,21 @@
                       {:end_time #inst "2524-01-01T00:00:00.000-00:00"}}])
          [{:valid_time
            {:end_time #inst "2524-01-01T00:00:00.000-00:00"}}])))
+
+(deftest pluralize-target-test
+  (is (= (transduce sut/pluralize-target conj
+                    [{:type "sighting"
+                      :target []}]) [{:type "sighting"
```
yogsototh: wrong indentation here.

PR #655, opened 2018-02-24
Relationship update route and store methods

### LINKED ISSUE #654: Support updating Relationships using a PUT route
In order to support imports avoiding data duplication, we need to let Relationships be updated, add a new PUT route for that entity similar to the others.

gbuisson:
closes #654 
Also updated tests to use ctim.examples


gbuisson:
PR issued from wrong branch

hshahman:
Verified in rel-20180312 v0.2

PR #656, opened 2018-02-26
Relationship update route & store methods

### LINKED ISSUE #654: Support updating Relationships using a PUT route

gbuisson:
closes #654
Also updated tests to use ctim.examples

### DIFF
--- a/src/ctia/domain/entities.clj
+++ b/src/ctia/domain/entities.clj
@@ -135,23 +135,29 @@
   (default-realize-fn "scratchpad" NewScratchpad StoredScratchpad))
 
 (s/defn realize-relationship :- StoredRelationship
-  [{:keys [source_ref target_ref]
-    :as new-relationship} :- NewRelationship
-   id :- s/Str
-   tempids :- (s/maybe TempIDs)
-   owner :- s/Str
-   groups :- [s/Str]]
-
-  (assoc new-relationship
-         :id id
-         :type "relationship"
-         :source_ref (get tempids source_ref source_ref)
-         :target_ref (get tempids target_ref target_ref)
-         :created (time/now)
-         :owner owner
-         :groups groups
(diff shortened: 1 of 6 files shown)


src/ctia/domain/entities.clj
```diff
@@ -135,23 +135,29 @@
...
+   (assoc new-relationship
+          :id id
+          :type "relationship"
+          :source_ref (get tempids source_ref source_ref)
+          :target_ref (get tempids target_ref target_ref)
+          :created (time/now)
+          :owner (or (:owner prev-object) owner)
+          :groups (or (:groups prev-object) groups)
+          :tlp (:tlp new-relationship (properties-default-tlp))
+          :schema_version schema-version))
```
msprunck: I suggest to use the `default-realize-fn` function and adding `:source_ref` and `:target_ref`.

src/ctia/domain/entities.clj
```diff
@@ -135,23 +135,29 @@
...
+    tempids :- (s/maybe TempIDs)
+    owner :- s/Str
+    groups :- [s/Str]
+    prev-object]
+   (assoc new-relationship
+          :id id
+          :type "relationship"
+          :source_ref (get tempids source_ref source_ref)
+          :target_ref (get tempids target_ref target_ref)
+          :created (time/now)
```
msprunck: `:created` should not be set for an update

src/ctia/domain/entities.clj
```diff
@@ -135,23 +135,29 @@
...
+          :owner (or (:owner prev-object) owner)
+          :groups (or (:groups prev-object) groups)
+          :tlp (:tlp new-relationship (properties-default-tlp))
```
msprunck: The `:tlp` field is not editable for an update

PR #661, opened 2018-02-28
disable indexing for the event :entity index field

### LINKED ISSUE #660: Disable Event Store :entity indexing
Alter the mapping to disable indexing on that field, as it makes the index hit the max fields limit.

Propose the following changes:
 - [ ] dynamic: false
 - [ ] remove dynamic_templates
 - [ ] map entitty.id, entity.type, entity.source to token

gbuisson:
closes #660 

https://www.elastic.co/guide/en/elasticsearch/reference/current/enabled.html

### DIFF
--- a/src/ctia/stores/es/mapping.clj
+++ b/src/ctia/stores/es/mapping.clj
@@ -621,7 +621,8 @@
     {:owner token
      :groups token
      :timestamp ts
-     :entity {:type "object"}
+     :entity {:enabled false
+              :type "object"}
      :id token
      :http-params {:type "object"}
      :type token


src/ctia/stores/es/mapping.clj
```diff
@@ -621,7 +621,8 @@
     {:owner token
      :groups token
      :timestamp ts
-     :entity {:type "object"}
+     :entity {:enabled false
```
saintx: Looks good.

hshahman:
No QA required.

PR #668, opened 2018-03-06
Remove Cider dependency

msprunck:
We don't need the nrepl Cider handler on DEV, INT or PROD.

Put Cider in project.clj produces a lot of conflicts with emacs


hshahman:
No QA required.

PR #670, opened 2018-03-13
Allow to DELETE indicators, update entities valid_time field

gbuisson:
- Add a Swagger undocumented `DELETE` route for `indicators`
- Break down `valid_time` computation logic from `default-realize`
- Change `valid_time` computation so that one may edit the dates

### DIFF
--- a/src/ctia/domain/entities.clj
+++ b/src/ctia/domain/entities.clj
@@ -74,6 +74,19 @@
       (contains? schema (s/required-key k))
       (contains? schema k)))
 
+(defn make-valid-time
+  "make a valid-time object from either in this order:
+  the newest value, the previous value or the default value"
+  [prev-valid-time new-valid-time now]
+  {:valid_time {:start_time
+                (or (:start_time new-valid-time)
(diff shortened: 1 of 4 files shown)


src/ctia/domain/entities.clj
```diff
@@ -74,6 +74,19 @@
...
       (contains? schema k)))
 
+(defn make-valid-time
+  "make a valid-time object from either in this order:
+  the newest value, the previous value or the default value"
+  [prev-valid-time new-valid-time now]
+  {:valid_time {:start_time
+                (or (:start_time new-valid-time)
+                    (:start_time prev-valid-time)
+                    now)
```
msprunck: `:start_time` and `:end_time` are optional in [CTIM](https://github.com/threatgrid/ctim/blob/master/src/ctim/schemas/common.cljc#L336). Why set a default value?

gbuisson: That's what that piece of code used to do, but we can make it a little smarter, set a default value only if the key is required.

PR #677, opened 2018-03-21
Ability to remove readonly access for the static auth

msprunck:
Closes threatgrid/iroh#1811

This PR adds the property `ctia.auth.static.readonly-for-anonymous`. When the static authentication is used `ctia.auth.type=static`, read/write access are handled as follows:

|ctia.auth.static.readonly-for-anonymous|password OK|read access|write access|
|----------------------------------------|-------------|------------|------------|
|true                                                             |yes                |yes               |yes                |
|true                                                             |no                  |yes              |no                |
|false                                                            |yes                |yes              |yes                |
|false                                                            |no                |no                 |no                  |

QA: Check the access described in the table above


hshahman:
Verified in rel-v0.3

PR #684, opened 2018-04-03
Malware abstraction level support

gbuisson:
Also set entity mappings to use `dynamic: false` to avoid useless migrations in the future


gbuisson:
for QA:

On CTIA, malware objects now support an abstraction_level field,
when posting/updating a malware one can set this field.

hshahman:
Verified in rel-v1.1

PR #685, opened 2018-04-04
Support JWT Scopes

yogsototh:
Support `ctia` and `casebook` scopes.

I changed the cert to sync and make it easier to both generate JWT and also test the access locally.


gbuisson:
Very clever way of doing it in the meantime. LGTM

PR #688, opened 2018-04-06
HTTP CRUD tests refactor

gbuisson:
Adds a new `crud` test helper to test common behaviors across our entity routes.
This PR allows to remove a lot of duplicated test code.

Also:
- put generative tests under a selector disabled by default.
- Enable to set the default refresh behavior for bulk inserts using properties

### DIFF
--- a/src/ctia/http/routes/bundle.clj
+++ b/src/ctia/http/routes/bundle.clj
@@ -318,7 +318,7 @@
 
 (defn bulk-params []
   {:refresh
-   (get-in @properties [:ctia :store :bundle-refresh] false)})
+   (get-in @properties [:ctia :store :bundle-refresh] "false")})
 
 (s/defn import-bundle :- BundleImportResult
   [bundle :- NewBundle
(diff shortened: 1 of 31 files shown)


src/ctia/http/routes/bundle.clj
```diff
@@ -318,7 +318,7 @@
 
 (defn bulk-params []
   {:refresh
-   (get-in @properties [:ctia :store :bundle-refresh] false)})
+   (get-in @properties [:ctia :store :bundle-refresh] "false")})
```
yogsototh: hmmmm.... that feel bad to use a string instead of a bool. I though properties were coerced.

edit: ok remember now that it is not a boolean but a 3 possible values, perhaps it would be best to use Refresh as an enum of keywords instead of strings.

gbuisson: yeah that hurts I agree :-)

hshahman:
No QA required.

PR #690, opened 2018-04-11
Fix the function that count the max number of entities in a bundle

msprunck:
Closes threatgrid/iroh#1825

The fix is tested using the updated `bundle-maximal` example (which contains a `:timestamp` field)

### DIFF
--- a/src/ctia/http/routes/bundle.clj
+++ b/src/ctia/http/routes/bundle.clj
@@ -372,7 +372,8 @@
                                  :create-sighting
                                  :create-tool
                                  :import-bundle}
-                 (if (> (bundle-size bundle)
-                        (bundle-max-size))
-                   (bad-request (str "Bundle max nb of entities: " (bundle-max-size)))
-                   (ok (import-bundle bundle external-key-prefixes auth-identity))))))
+                 (let [max-size (bundle-max-size)]
+                   (if (> (bundle-size bundle)
+                          max-size)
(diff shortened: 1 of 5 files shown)


src/ctia/http/routes/bundle.clj
```diff
@@ -366,7 +372,7 @@
                                  :create-sighting
                                  :create-tool
                                  :import-bundle}
-                 (if (> (bulk/bulk-size bundle)
-                        (bulk/get-bulk-max-size))
-                   (bad-request (str "Bundle max nb of entities: " (bulk/get-bulk-max-size)))
+                 (if (> (bundle-size bundle)
+                        (bundle-max-size))
+                   (bad-request (str "Bundle max nb of entities: " (bundle-max-size)))
```
yogsototh: I would a temporary variable here not to call `bundle-max-size` twice.

test/ctia/http/routes/bundle_test.clj
```diff
@@ -187,7 +187,8 @@
                            indicators
                            sightings)]
     (testing "Import bundle with all entity types"
-      (let [new-bundle (deep-dissoc bundle-maximal :id)
+      (let [new-bundle (-> bundle-maximal
+                           deep-dissoc-entity-ids)
```
yogsototh: I think you can remove the thread here and juste use `(deep-dissoc-entity-ids bundle-maximal)`.

PR #691, opened 2018-04-11
Project Structure refinements

gbuisson:
Slight changes to make life easier when changing entity behaviors

- Simplified Store protocols, there's now a generic `IStore` protocol most entities use it
- added an `entity-crud-routes` function that generates api routes for 90% of our needs
- Try to split entity definition code from implementation, entity ns is top level for easy access
- simplified id manipulation code
- some kibit here and there


gbuisson:
I got everything working except when I rebased latest commits:

Caused by: java.lang.RuntimeException: Binding form is not vector: login

If someone has a clue you can commit directly to this branch!

yogsototh:
Moved to https://github.com/threatgrid/ctia/pull/696

hshahman:
No QA required.

PR #696, opened 2018-04-24
Finish the PR from GB

yogsototh:
See https://github.com/threatgrid/ctia/pull/691

Mainly fixes badly/fogotten merged git conflicts
But also renamed the `:identity` by `:auth-identity` in the compojure-api routes to prevent defmethod collision with the use of `ring-jwt-middleware`.

Also Fix a pb with the :import-bundle capability that wasn't included with scopes. Now there is `private-intel` scope that is supported as well as a de facto inclusion if the user has `private-intel` scope. (notice `private-intel` scope can be different depending on config).


yogsototh:
Thanks for that great work @gbuisson !

PR #699, opened 2018-05-03
Fixed missing casebook patch route

gbuisson:
Fix leftover Casebook patch route from the last refactor.


hshahman:
Verified in rel-v1.1

PR #703, opened 2018-05-04
Fix an exception when trying to DELETE an unexisting entity record

gbuisson:


src/ctia/flows/crud.clj
```diff
@@ -486,7 +486,7 @@
   (let [entity (get-fn entity-id)]
     (-> {:flow-type :delete
          :entity-type entity-type
-         :entities [entity]
+         :entities (remove nil? [entity])
```
msprunck: Not easier to test if the entity is nil?

```clojure
(if entity [entity] [])
```

gbuisson: I'm not sure, I prefer `remove nil?`so I avoid to repeat `entity`, anyway the root of the problem is a flaw in our flow abstraction with a mismatch between collections/single items, we should address that later.

hshahman:
Verified in rel-v1.1

PR #704, opened 2018-05-07
Fix migration from 0.4.28 to 1.0.0

### LINKED ISSUE #702: Fix migration from schema 0.4.28 to 1.0.0
Log message in Kibana https://kibana-n.int.iroh.site/goto/3718e2ef885eb2a339c0cd5185b9bacf

The error seems related to the renaming of `pki-serial` to `pki_serial`:

```clojure
{:relations [{:related {:type (not (#{"file_path" "mac_address" "device" "hostname" "url" "user" "ipv6" "email" "sha256" "sha1" "md5" "ip" "domain" "imei" "amp_computer_guid" "pki_serial" "imsi" "file_name"} "pki-serial"))}}]}
```

msprunck:
Closes #702

- Rename pki-serial observable type to pki_serial
- Update simplified incident model
- Add a default catch clause for the whole migration process
- Improve the display of coercion errors

When a validation error occurs, the whole process is not stopped. Is this the expected behaviour?

### DIFF
--- a/src/ctia/task/migrations.clj
+++ b/src/ctia/task/migrations.clj
@@ -123,7 +123,7 @@
            ("judgement"
             "verdict") (rename-judgement-observable-type doc old new)
            "casebook" (rename-casebook-observable-types doc old new)
-           identity))))
+           doc))))
 
 ;;--- Simplify incident model
 
--- a/src/ctia/task/migrations.clj
+++ b/src/ctia/task/migrations.clj
@@ -1,7 +1,8 @@
 (ns ctia.task.migrations
   (:require [clj-momo.lib.clj-time
              [coerce :as time-coerce]
-             [core :as time-core]]))
+             [core :as time-core]]
+            [clojure.set :as set]))
 
 (def add-groups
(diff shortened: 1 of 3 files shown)


gbuisson:
The validation step is done after the migration, I don't think we should stop, IMO it's better to get knowledge about all inconsistencies at once

src/ctia/task/migrations.clj
```diff
@@ -55,6 +56,98 @@
...
+                          (mapv #(rename-judgement-observable-type % old new)
+                                verdicts)))))
+
+(defn rename-casebook-observable-types
+  [{:keys [observables bundle] :as casebook} old new]
+  (cond-> casebook
+    (seq observables) (update :observables with-renamed-observable-types old new)
+    bundle (update :bundle rename-bundle-observable-types old new)))
+
+(defn rename-observable-type
```
gbuisson: I would have made a piece of code that given any document traverses it and changes the observable_type key,
but it's a matter of taste.

hshahman:
No QA required.

PR #709, opened 2018-05-09
Longer timeouts for migrations

gbuisson:
Allow to specify an ES timeout per store, longer timeouts for migrations.

I cannot add a test easily for the timeout setting because clj-http-fake overrides the connection manager ignoring its timeout setting.


hshahman:
No QA required.

PR #710, opened 2018-05-11
Report all schema errors at the end of the migration process

msprunck:
The current migration only reports schema errors with logs during the migration process. Errors are not reported at the end.

This PR changes this behaviour to report all schema errors at the end of the migration process and return the -1 status code if an error occurs.

### DIFF
--- a/src/ctia/task/migrate_es_stores.clj
+++ b/src/ctia/task/migrate_es_stores.clj
@@ -257,6 +257,21 @@
                      (:type target-store)
                      checked-count))))))
 
+(defn check-store-index
+  [[sk sr :as store] batch-size]
+  (try
+    (log/infof "checking store: %s" sk)
+    (check-store sr batch-size)
+    (catch Exception e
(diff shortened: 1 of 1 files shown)


src/ctia/task/migrate_es_stores.clj
```diff
@@ -257,6 +257,21 @@
...
                      checked-count))))))
 
+(defn check-store-index
+  [[sk sr :as store] batch-size]
+  (try
+    (log/infof "checking store: %s" sk)
+    (check-store sr batch-size)
+    (catch Exception e
+      (if-let [errors (some->> (ex-data e) :error (remove nil?))]
+        (let [message (format (str "The store %s is invalid, value cannot be coerced "
```
yogsototh: I wouldn't be so definitive in that sentence. Perhaps, use "certainly a coercion issue". Because I'm not sure having a `:error` field in the ex-data would be enough to be certain this is a coercion issue.

PR #714, opened 2018-05-16
fix bad root-scopes initialisation

### LINKED ISSUE #713: Write into Public CTIA by a JWT token
Currently, we can POST any types of entity (a sighting for example) into public CTIA using a JWT token, like a token from the local storage in the browser.
I thought we can write into CTIA using API_KEY only. 
Is it an issue or by design?

@folbricht @craigbro @gbuisson @saintx

yogsototh:
Close #713 
Close https://github.com/threatgrid/iroh/issues/1949

Root scopes properties wasn't correctly set due to bad init ordering.

_We should think about using TK or a similar project to handle init order and prevent future errors like these._

### DIFF
--- a/src/ctia/auth/jwt.clj
+++ b/src/ctia/auth/jwt.clj
@@ -10,11 +10,11 @@
     [auth :as auth :refer [IIdentity]]
     [properties :as prop]]))
 
-(def entity-root-scope
+(defn entity-root-scope []
   (get-in @prop/properties [:ctia :auth :entities :scope]
           "private-intel"))
 
-(def casebook-root-scope
+(defn casebook-root-scope []
   (get-in @prop/properties [:ctia :auth :casebook :scope]
           "casebook"))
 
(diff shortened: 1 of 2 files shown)


test/ctia/auth/jwt_test.clj
```diff
@@ -33,45 +40,45 @@
 
 
 (deftest scopes-to-capapbilities-test
-  (is (= "private-intel" sut/entity-root-scope)
+  (is (= "private-intel" @sut/entity-root-scope)
```
gbuisson: setting default properties configuration should be done on `ctia-default.properties` and not hardcoded, doing this will make this test useless.

src/ctia/auth/jwt.clj
```diff
@@ -10,13 +10,19 @@
     [auth :as auth :refer [IIdentity]]
     [properties :as prop]]))
 
-(def entity-root-scope
-  (get-in @prop/properties [:ctia :auth :entities :scope]
-          "private-intel"))
+(def entity-root-scope (atom nil))
```
gbuisson: I would much prefer if we change the `def` to `defn` instead of creating an atom just for that

PR #715, opened 2018-05-18
Open Schema Version

gbuisson:
closes https://github.com/threatgrid/iroh/issues/1952

Completes the `schema_version` ignore overriding the schemas exposed to the route handlers.

### DIFF
--- a/src/ctia/schemas/core.clj
+++ b/src/ctia/schemas/core.clj
@@ -1,23 +1,31 @@
 (ns ctia.schemas.core
-  (:require [ctim.schemas
-             [verdict :as vs]
-             [bundle :as bundle]
-             [common :as cos]
-             [vocabularies :as vocs]]
-            [flanders
-             [schema :as f-schema]
-             [spec :as f-spec]
-             [utils :as fu]]
-            [schema-tools.core :as st]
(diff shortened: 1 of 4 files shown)


src/ctia/schemas/core.clj
```diff
@@ -1,23 +1,31 @@
...
+
+(sc/defschema OpenCTIMSchemaVersion
+  {(sc/optional-key :schema_version) s/Str})
+
+(sc/defschema CTIAEntity
+  (st/merge
+   OpenCTIMSchemaVersion
+   (st/optional-keys
+    {:authorized_users [Str]
+     :authorized_groups [Str]})))
```
yogsototh: Why use two `defschema` declaration here? Knowing that `OpenCTIMSchemaVersion` is only used here? Why not just

~~~clojure
(sc/defschema CTIAEntity
  (st/optional-keys
    {:authorized_users [Str]
     :authorized_groups [Str]
     :schema_version s/Str}))
~~~

gbuisson: I wanted to make it clear and obvious if we decide otherwise later.

hshahman:
Verified in rel-v1.2

PR #716, opened 2018-05-21
Incident PATCH

gbuisson:
close https://github.com/threatgrid/iroh/issues/1962

Make the `PATCH`api feature available to any entity, enable it only for Casebook and Incident.


hshahman:
Verified in rel-v1.2

PR #717, opened 2018-05-22
Incident Status update route

gbuisson:
close https://github.com/threatgrid/iroh/issues/1963

also return a proper 404 if trying to update an unknown record.

### DIFF
--- a/src/ctia/entity/incident.clj
+++ b/src/ctia/entity/incident.clj
@@ -1,5 +1,10 @@
 (ns ctia.entity.incident
   (:require
+   [ctia.domain.entities :refer [default-realize-fn un-store with-long-id]]
+   [ctia.flows.crud :as flows]
+   [ring.util.http-response :refer [ok not-found]]
+   [clj-momo.lib.clj-time.core :as time]
+   [compojure.api.sweet :refer [context POST routes]]
    [ctia.domain.entities :refer [default-realize-fn]]
    [ctia.store :refer :all]
    [ctia.http.routes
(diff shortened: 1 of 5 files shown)


test/ctia/http/routes/incident_test.clj
```diff
@@ -22,6 +25,52 @@
...
+                         :body new-status
+                         :headers {"Authorization" "45c1f5e3f05d0"})
+          updated-incident (:parsed-body response)]
+      (is (= 200 (:status response)))
+      (is (= "Open" (:status updated-incident)))
+      (is (get-in updated-incident [:incident_time :opened]))
+
+      (is (t/within?
+           (t/interval (t/minus (t/now) (t/minutes 10))
+                       (t/plus (t/now) (t/minutes 10)))
```
yogsototh: Could you consider using a `with-redef` for `now` instead of using magic numbers here? I think there should be a fixture for that.

src/ctia/entity/incident.clj
```diff
@@ -46,6 +53,65 @@
...
+         :identity-map identity-map
+         (let [status-update (make-status-update update)]
+           (if-let [updated
+                    (-> (flows/patch-flow
+                         :get-fn #(read-store :incident
+                                              read-record
+                                              %
+                                              identity-map
+                                              {})
+                         :realize-fn realize-incident
```
msprunck: I think that the `status-update` should be done by the `realize-fn`. You could compose the default `realize-incident` with the `status-update`

gbuisson: good idea, I might change my mind but I prefer to have it explicitly defined that way, will change it if that becomes an issue.

hshahman:
Verified in rel-v1.2

PR #718, opened 2018-05-23
Split Bundle and Bulk ns, rename entity tests

gbuisson:
connected https://github.com/threatgrid/iroh/issues/1965

In order to implement Incident related entities retrieval, 
we need to access the bulk/bundle code.

This PR decouples the HTTP layer from the core layer, 
also rename all entity http tests so they match the related entity ns


hshahman:
No QA required.

PR #719, opened 2018-05-30
Entity Bundle export route

gbuisson:
closes https://github.com/threatgrid/iroh/issues/1965

This adresses what is required for #1965 except that I made it generic for all CTIA entities, one can export a Bundle with an entity with its relationships and all the target objects as well.


src/ctia/bundle/routes.clj
```diff
@@ -3,17 +3,53 @@
...
+                                :list-indicators
+                                :read-judgement
+                                :list-judgements
+                                :read-malware
+                                :read-relationship
+                                :list-relationships
+                                :read-sighting
+                                :list-sightings
+                                :read-tool
+                                :read-verdict}
```
msprunck: Why the `:list-` capability is only required for indicators, judgements, relationships and sightings? All entities can have a relationship between them.

src/ctia/auth/capabilities.clj
```diff
@@ -1,17 +1,52 @@
 (ns ctia.auth.capabilities
-  (:require
-   [ctia.entity.entities :refer [entities]]))
+  (:require [clojure.set :as set]))
+
+(def entities
+  #{:actor
+    :attack-pattern
+    :campaign
+    :casebook
```
yogsototh: casebook was not present on purpose in the entities list in the jwt ns.
So with this code this will make people with private-intel scope be able to access the casebook route which is incorrect.

src/ctia/bundle/core.clj
```diff
@@ -316,4 +324,121 @@
...
+                                 source_ref]}]
+                      #{target_ref source_ref})
+                    relationships))
+        by-type (dissoc (group-by
+                         #(ent/long-id->entity-type %)
+                         all-ids) nil)]
+    (bulk-keys->bundle-keys
+     (bulk/fetch-bulk by-type identity-map))))
+
+(def list-limit 100)
```
yogsototh: I believe this number should be configurable.

src/ctia/bulk/core.clj
```diff
@@ -1,35 +1,34 @@
...
+   [ctia.bulk.core :as bulk]))
+
+(def bulk-entity-mapping
+  (into {}
+        (map (fn [{:keys [entity plural]}]
+               {entity plural}) (vals entities))))
 
 (defn bulk-key
-  "Returns the bulk key for a given entity type"
+  "Converts a bulk entity key to an entity type
```
msprunck: It's the opposite -> "Converts an entity type to a bulk key"

src/ctia/bulk/core.clj
```diff
@@ -1,35 +1,34 @@
...
+   [ctia.entity.entities :refer [entities]]
+   [ctia.flows.crud :as flows]
+   [ctia.lib.keyword :refer [singular]]
+   [ring.util.http-response :refer :all]
+   [ctia.bulk.core :as bulk]))
+
+(def bulk-entity-mapping
+  (into {}
+        (map (fn [{:keys [entity plural]}]
+               {entity plural}) (vals entities))))
```
msprunck: Nice!

src/ctia/bulk/core.clj
```diff
@@ -1,35 +1,34 @@
...
 (defn entity-type-from-bulk-key
   "Converts a bulk entity key to an entity type
    Ex: :attack_patterns -> :attack-pattern"
   [k]
-  (case k
-    :attack_patterns :attack-pattern
-    :exploit_targets :exploit-target
-    :data_tables :data-table
-    (singular k)))
+  (get (set/map-invert bulk-entity-mapping) k))
```
msprunck: It's not better to set the inverted map in a def?
```clojure
(def inverted-bulk-entity-mapping (set/map-invert bulk-entity-mapping))
```

src/ctia/bundle/core.clj
```diff
@@ -316,4 +302,114 @@
...
+                read-fn
+                id
+                identity-map
+                {})))
+
+(def empty-bundle
+  {:id (id/make-transient-id "bundle")
+   :type "bundle"
+   :source "ctia"
+   :valid_time {}})
```
msprunck: A bundle is not an entity, it should not contain an `id` and a `valid_time`

src/ctia/bulk/core.clj
```diff
@@ -81,9 +80,12 @@
   "
   [func bulk & args]
   (->> bulk
-       (pmap (fn [[entity-type entities]]
-               [entity-type
-                (apply func entities (entity-type-from-bulk-key entity-type) args)]))
+       (pmap (fn [[bulk-key entities]]
```
msprunck: Would you mind changing `bulk-key` by `bulk-k` to avoid confusion with the function?

src/ctia/entity/feedback.clj
```diff
@@ -78,21 +78,23 @@
...
+     :new-schema fs/NewFeedback
+     :entity-schema fs/Feedback
+     :get-schema fs/PartialFeedback
+     :get-params FeedbackGetParams
+     :list-schema fs/PartialFeedbackList
+     :external-id-q-params FeedbackByExternalIdQueryParams
+     :realize-fn fs/realize-feedback
+     :get-capabilities :read-feedback
+     :post-capabilities :create-feedback
+     :put-capabilities :create-feedbacl
```
msprunck: typo `:create-feedback`

gbuisson: nice catch!

test/ctia/auth/jwt_test.clj
```diff
@@ -41,12 +41,18 @@
...
+         (set/difference caps/all-capabilities
+                         (sut/scopes-to-capabilities #{(sut/entity-root-scope)
+                                                       (sut/casebook-root-scope)})))
+      "with all scopes you should have most capabilities except some very
        specific ones")
+
+  (is (= #{:developer :specify-id :external-id}
+         (set/difference caps/all-capabilities-no-casebook
+                         (sut/scopes-to-capabilities #{(sut/entity-root-scope)})))
+      "entity scope don't provide casebook")
```
yogsototh: Would you mind to add another test to see the difference between `caps/all-capabilities` when the scope doesn't contain casebook so we could verify the difference between `all-capabilities` and `all-capabilities-no-casebook`.

hshahman:
Verified in rel-v1.4

PR #720, opened 2018-06-08
Incident casebook link route

gbuisson:
epic https://github.com/threatgrid/iroh/issues/1940

closes  https://github.com/threatgrid/iroh/issues/1964

Also, moved bulk and bundle test ns to their correct location.


test/ctia/link/routes_test.clj
```diff
@@ -0,0 +1,69 @@
...
+         (is (= 201 incident-status))
+         (is (= 201 link-status))
+
+         (is (= (:id casebook-body)
+                (:source_ref link-response))
+             "The New Relationship targets the incident")
+
+         (is (= (:id incident-body)
+                (:target_ref link-response))
+             "The New Relationship sources the casebook"))))))
```
msprunck: Could you check that the relationship exists in the store? I'm not sure that the result comes from the store.

src/ctia/link/routes.clj
```diff
@@ -0,0 +1,75 @@
+(ns ctia.link.routes
```
msprunck: Why not add this route to the `ctia.entity.incident` or `ctia.entity.relationship` ns? For me a link between a casebook and a incident is a `Relationship`.

gbuisson:
I think I addressed all feedback

src/ctia/entity/relationship.clj
```diff
@@ -62,6 +71,69 @@
...
+
+(def incident-casebook-link-route
+  (POST "/:id/link" []
+        :return rs/Relationship
+        :body [link-req IncidentCasebookLinkRequest
+               {:description "an Incident Link request"}]
+        :header-params [{Authorization :- (s/maybe s/Str) nil}]
+        :summary "Link an Incident to a Casebook"
+        :path-params [id :- s/Str]
+        :capabilities #{:create-incident
```
msprunck: Why the `:create-incident` capability is required? Only a Relationship is created.

src/ctia/entity/relationship.clj
```diff
@@ -62,6 +71,69 @@
...
+                                       long-id->id
+                                       :short-id)
+                                   identity-map
+                                   {})]
+
+          (cond
+            (not incident)
+            (not-found)
+            (not casebook)
+            (not-found)
```
msprunck: IMO, if the casebook is not found, it's a bad request. The `not-found` response is only related to the request URI.

hshahman:
Verified in rel-v1.4

PR #727, opened 2018-06-18
ES migration Tweak Index Settings pre/post migration

gbuisson:
Change the migraton behavior so that:
- new indices are first created using 0 replicas and indexing disabled
- then when the migration is done, the correct values are set and manual indexing is triggered.

- this new behavior can be turned on/off using properties, (off by default)
- also exposed the ES refresh_interval configuration on properties https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-update-settings.html
- moved previous benchmarks to their correct ns

On my machine, the perforate bench shows on averagea 2s improvement over a migration of 100k records.

```
Goal:  Run a migration
-----
Case:  :optimizations_enabled
Evaluation count : 12 in 12 samples of 1 calls.
             Execution time mean : 1.274855 min
    Execution time std-deviation : 939.727463 ms
   Execution time lower quantile : 1.253696 min ( 2.5%)
   Execution time upper quantile : 1.293570 min (97.5%)

Case:  :base
Evaluation count : 12 in 12 samples of 1 calls.
             Execution time mean : 1.297877 min
    Execution time std-deviation : 1.253525 sec
   Execution time lower quantile : 1.271381 min ( 2.5%)
   Execution time upper quantile : 1.327619 min (97.5%)
```

### DIFF
--- a/benchmarks/ctia/entity/actor_bench.clj
+++ b/benchmarks/ctia/entity/actor_bench.clj
@@ -1,0 +1,31 @@
+(ns ctia.entity.actor-bench
+  (:require [ctia.test-helpers
+             [benchmark :refer [cleanup-ctia!
+                                setup-ctia-es-store!]]
+             [core :as helpers :refer [delete post]]]
+            [ctim.examples.actors
+             :refer [new-actor-minimal
+                     new-actor-maximal]]
+            [perforate.core :refer :all]))
+
+(def small-actor new-actor-minimal)
+(def big-actor (dissoc new-actor-maximal :id))
+
+(defgoal create-actor "Create Actor"
+  :setup (fn [] [(setup-ctia-es-store!)])
+  :cleanup (fn [_] cleanup-ctia!))
+
+(defn play [fixture]
+  (let [{:keys [status parsed_body]}
+        (post "ctia/actor"
+              :body fixture
+              :headers {"Authorization" "45c1f5e3f05d0"})]
+    (if (= 201 status)
+      (delete (str "ctia/actor" (:id parsed_body)))
+      (prn "play-big: " status))))
+
+(defcase create-actor :big-actor-es-store
+  [_] (play big-actor))
+
+(defcase create-actor :small-actor-es-store
+  [_] (play small-actor))
--- a/benchmarks/ctia/entity/actor_bench.clj
+++ b/benchmarks/ctia/entity/actor_bench.clj
@@ -21,7 +21,6 @@
               :body fixture
               :headers {"Authorization" "45c1f5e3f05d0"})]
     (if (= 201 status)
-      (delete (str "ctia/actor" (:id parsed_body)))
       (prn "play-big: " status))))
 
 (defcase create-actor :big-actor-es-store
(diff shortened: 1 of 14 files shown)


benchmarks/ctia/entity/actor_bench.clj
```diff
@@ -0,0 +1,31 @@
...
+  (let [{:keys [status parsed_body]}
+        (post "ctia/actor"
+              :body fixture
+              :headers {"Authorization" "45c1f5e3f05d0"})]
+    (if (= 201 status)
+      (delete (str "ctia/actor" (:id parsed_body)))
+      (prn "play-big: " status))))
+
+(defcase create-actor :big-actor-es-store
+  [_] (play big-actor))
```
yogsototh: `play` does not only create but also either delete or print to stdout so I think it doesn't really measure only the time it takes to call `create-actor`. Why is the `delete` necessary, isn't it handled by the `cleanup-ctia!`?

I would either rename `create-actor` as `create-and-delete-actor` or remove completely the `if` body to remove the `delete` and `prn` statements.

benchmarks/ctia/entity/campaign_bench.clj
```diff
@@ -0,0 +1,34 @@
...
+  :cleanup (fn [_] cleanup-ctia!))
+
+(defn play [fixture]
+  (let [{:keys [status parsed_body]}
+        (post "ctia/campaign"
+              :body fixture
+              :headers {"Authorization" "45c1f5e3f05d0"})]
+    (if (= 201 status)
+      (delete (str "ctia/campaign" (:id parsed_body)))
+      (prn "play-big: " status))))
```
yogsototh: same remark

src/ctia/task/migrate_es_stores.clj
```diff
@@ -130,32 +136,53 @@
      "false"
      bulk-max-size)))
 
+(def optimized-settings
+  {:index
+   {:number_of_replicas 0
+    :refresh_interval -1}})
```
yogsototh: If this is an optimisation why would someone would want not to use it and make it configurable?

I think you should add a documentation in some file about the advantages and drawbacks of the two different approaches.

benchmarks/ctia/entity/actor_bench.clj
```diff
@@ -0,0 +1,30 @@
...
+  :setup (fn [] [(setup-ctia-es-store!)])
+  :cleanup (fn [_] cleanup-ctia!))
+
+(defn play [fixture]
+  (let [{:keys [status parsed_body]}
+        (post "ctia/actor"
+              :body fixture
+              :headers {"Authorization" "45c1f5e3f05d0"})]
+    (if (= 201 status)
+      (prn "play-big: " status))))
```
yogsototh: I think we should get rid of the complete block here. It will always generate a `(prn "play-big: 201")` and it takes a bit of time for IOs and as the only operation is a post, that time might not be totally marginal.

benchmarks/ctia/entity/campaign_bench.clj
```diff
@@ -0,0 +1,33 @@
...
+  :setup (fn [] [(setup-ctia-es-store!)])
+  :cleanup (fn [_] cleanup-ctia!))
+
+(defn play [fixture]
+  (let [{:keys [status parsed_body]}
+        (post "ctia/campaign"
+              :body fixture
+              :headers {"Authorization" "45c1f5e3f05d0"})]
+    (if (= 201 status)
+      (prn "play-big: " status))))
```
yogsototh: same remark here.

PR #728, opened 2018-06-25
Incident link: handle the case where a bad casebook_id is issued

gbuisson:
closes https://github.com/threatgrid/iroh/issues/1964


gbuisson:
I adressed it here: https://github.com/threatgrid/ctia/pull/728/files#diff-f2e6dc9f1e019cd84704e11bf057a066R110

Actually he was using an URL inside the URL to issue the link request.

hshahman:
Verified in rel-v1.4

PR #730, opened 2018-07-03
add an helper to get the full list of scopes in a comment

yogsototh:
I frequently need to retrieve the full list of scopes to provide it to "clients".
Any cleaner suggestion is welcome.

### DIFF
--- a/src/ctia/auth/capabilities.clj
+++ b/src/ctia/auth/capabilities.clj
@@ -1,7 +1,8 @@
 (ns ctia.auth.capabilities
   (:require
    [ctia.entity.entities :refer [entities]]
-   [clojure.set :as set]))
+   [clojure.set :as set]
+   [clojure.string :as string]))
 
 (def all-entities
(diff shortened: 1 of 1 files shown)


src/ctia/auth/capabilities.clj
```diff
@@ -45,11 +46,42 @@
...
+          n (if (= "list" loc)
+              (subs p 0 (dec (count p)))
+              p)]
+      (if n
+        (str "private-intel/" n suff)
+        (str "private-intel/" loc))))
+
+  (sort (set/union
+         #{"casebook:read" "casebook:write"}
+         (set (map cap-to-scope all-capabilities-no-casebook))))
```
msprunck: Why not use `all-capabilities` which contains casebook capabilities?

yogsototh: because casebook are a special case and having the root scope `private-intel` doesn't provide access to casebook.

PR #732, opened 2018-07-09
Upgrade to CTIM 1.0.1

gbuisson:
closes https://github.com/threatgrid/iroh/issues/2066

- [x] remove ExploitTarget
- [x] Weakness support
- [x] Vulnerability support


src/ctia/schemas/graphql/flanders.clj
```diff
@@ -104,8 +104,10 @@
       nil))
 
   NumberType
-  (->graphql' [{:keys [description]} _ _]
-    {:type Scalars/GraphQLString
+  (->graphql' [{:keys [description float?]} _ _]
+    {:type (if-not float?
+             Scalars/GraphQLInt
+             Scalars/GraphQLFloat)
```
msprunck: IMO we should not use the NumberType as a parent of Float, Int. We should only use specific types in flanders. It's a good workaround for now but we should define proper types in flanders in the future

PR #733, opened 2018-07-10
Fix the NPE which occurs when a relationship contains an invalid ref

msprunck:
Closes threatgrid/iroh#1829

This PR fixes the NPE which occurs when a relationship contains an invalid ID, ex:
`transient:1345`. 

The NPE was thrown by the code that resolves the entity identified by the invalid ID when a `target_entity` or `source_entity` field was requested in a GraphQL query.

Ex: 

```graphql
query TTPs {
  observable(type:"sha1", value:"8305cb9f39638dad87bbc1bfb84f8de1630e2b32") {
    type
    value
    judgements {
      nodes {
         relationships {
          nodes {
            target_entity {
              ...indicatorFields
            }
          }
        }
      }
    }
  }
}

fragment indicatorFields on Indicator {
  type
  relationships {
    nodes {
      target_entity {
        ...attackPatternFields
        ...malwareFields
        ...toolFields
      }
    }
  }
}

fragment attackPatternFields on AttackPattern {
  id
  type
  name
}

fragment malwareFields on Malware {
  id
  type
  name
}

fragment toolFields on Tool {
  id
  type
  name
}
```

QA:

1. Add a relationship with a `target_ref` set with `transient:1234` to CTIA
2. Send a GraphQL query that requests the `target_source` field

### DIFF
--- a/src/ctia/entity/relationship/graphql_schemas.clj
+++ b/src/ctia/entity/relationship/graphql_schemas.clj
@@ -39,7 +39,7 @@
     (g/new-object name description [] (into fields
                                             related-judgement-fields))))
 
-(s/defn ref->entity-type :- s/Keyword
+(s/defn ref->entity-type :- (s/maybe s/Keyword)
   "Extracts the entity type from the Reference"
   [ref :- s/Str]
   (some-> ref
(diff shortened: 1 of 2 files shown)


src/ctia/entity/relationship/graphql_schemas.clj
```diff
@@ -69,25 +69,26 @@
 
 (def relation-fields
   (merge
```
yogsototh: is the merge still needed?

hshahman:
Verified in rel-v1.6

PR #734, opened 2018-07-24
Remove Store check process from the migration task

gbuisson:
closes https://github.com/threatgrid/iroh/issues/2088

- We now only validate the bulks and check store size
- I kept the store validation process as another dedicated task


src/ctia/task/check_es_stores.clj
```diff
@@ -0,0 +1,153 @@
...
+            offset (:offset next 0)
+            search_after (:sort paging)
+            checked (coerce! data)
+            checked-count (+ checked-count
+                             (count checked))]
+        (if next
+          (recur offset search_after checked-count)
+          (log/infof "%s - finished checking %s documents"
+                     (:type target-store)
+                     checked-count))))))
```
yogsototh: To make the code slightly more modulable I would use a higher order function like: `(do-all-documents store batch-size f)` where `f` will be your function to check each doc. It appears this is the kind of thing that could be reused for further migration process.

Also the `coerce!` is somehow hidden inside the make the `loop` logic. That made a bit difficult for me to discover where the check was really made.

PR #738, opened 2018-08-31
Removing usage of Elasticsearch field data cache

msprunck:
Closes threatgrid/iroh#1955

The `fielddata` option was used to sort on text data field. This PR removes its usage by:
1- Changing the datatype of all `token` to the keyword datatype that allows the use of `doc_values` by default.
2- Using a keyword subfield when we have to sort on a `text` field.

⚠️  This PR requires to reindex all indexes ⚠️


src/ctia/stores/es/crud.clj
```diff
@@ -190,6 +191,51 @@
    {:all-of {s/Any s/Any}
     :one-of {s/Any s/Any}}))
 
+(def sort-fields-mapping
+  "Mapping table for all fields which needs to be renamed
+   for the sorting. Instead of using fielddata we can have
+   a text field for full text searches, and an unanalysed keyword
+   field with doc_values enabled for sorting"
+  {"title" "title.whole"})
```
ereteog: This assumes that any type with a `title` field will be indexed with multifield, and at least one named `whole` like [this](https://github.com/threatgrid/ctia/blob/master/src/ctia/stores/es/mapping.clj#L64):
```
(def describable-entity-mapping
  {:title (assoc all_text
                 :fields {:whole all_token})
   :short_description all_text
   :description all_text})
```
Are all these entities the only ones that should refer it?
```
⋊> ~/w/ctia on es-migration-2556 ⨯ grep -r "describable-entity-mapping" src/*                                 14:19:52
src/ctia/entity/investigation.clj:     em/describable-entity-mapping
src/ctia/entity/sighting/es_store.clj:     em/describable-entity-mapping
src/ctia/entity/actor.clj:     em/describable-entity-mapping
src/ctia/entity/judgement/es_store.clj:     em/describable-entity-mapping
src/ctia/entity/relationship.clj:     em/describable-entity-mapping
src/ctia/entity/incident.clj:     em/describable-entity-mapping
src/ctia/entity/weakness/mapping.clj:     em/describable-entity-mapping
src/ctia/entity/data_table.clj:     em/describable-entity-mapping
src/ctia/entity/campaign.clj:     em/describable-entity-mapping
src/ctia/entity/casebook.clj:     em/describable-entity-mapping
src/ctia/entity/indicator.clj:     em/describable-entity-mapping
src/ctia/entity/vulnerability/mapping.clj:     em/describable-entity-mapping
src/ctia/entity/coa.clj:     em/describable-entity-mapping
src/ctia/stores/es/mapping.clj:(def describable-entity-mapping
```
My main concern is that it's hidden here.

src/ctia/stores/es/mapping.clj
```diff
@@ -28,19 +27,12 @@
 (def token
   "A mapping for fields whose value should be treated like a symbol.
   They will not be analyzed, and they will be lowercased."
-  {:type "text"
-   ;; TODO use token and disable fielddata once token analyzer is supported
-   :fielddata true
-   :analyzer "token_analyzer"
-   :search_analyzer "token_analyzer"})
```
ereteog: Here is the summary of the conversation about the `lenght` filter that was used for tokens.
This contribution removes it. We have 3 solutions:
- keep this new behavior which is probably the best solution. Since we remove the fielddata, it will not produce any memory issue. In that case you could remove the `token_len` filter and `token_analyzer`  definitions in the index settings, which are not used any more.
- If we do want to not index too long tokens, it has to be reverted and tokens longer that 255 will not be indexed at all and thus could not be queried, which would be the case for many urls. We could significantly raise that value for users that will look for long urls for instance.
 - replace that `length` filter with a [`truncate` filter](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/analysis-truncate-tokenfilter.html) to enable to search on long urls: `{"type": "truncate", "length": 2048}` while limiting the size indexed value (but preserving its full value in `_source`). For the case of urls, they are in general distinct on first 2048 characters.

PR #740, opened 2018-09-10
CTIM 1.0.4

gbuisson:
- Added Vulnerability CVSSv2 fields on mapping and graphql definitions

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -44,7 +44,7 @@
                   :exclusions [prismatic/plumbing
                                potemkin
                                com.andrewmcveigh/cljs-time]]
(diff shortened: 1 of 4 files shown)


hshahman:
No QA required.

saintx:
Heads-up.  This broke imports on INT, and before we run it in PROD we'll need to run a data migration.

gbuisson:
@saintx please avoid mentioning our environments in the Open Source repository, I suggest we discuss that in the product repository.

This model was not officially in production nor being used by any customer, I suggest we simply delete the existing indexes.

PR #743, opened 2018-09-26
Event Model updates and API

gbuisson:
⚠️ This PR needs a reset of our CTIA event indexes ⚠️ 

closes threatgrid/iroh#2024

- Move event code from CTIM to CTIA
- Change the event model to be more usable regarding UI consumption
- Add Event API routes allowing to query records for any entity/linked entity, access control filtering
- A lot of fixes, especially regarding event creation accuracy


src/ctia/entity/event/crud.clj
```diff
@@ -0,0 +1,74 @@
...
+   (->> events
+        (map (partial attach-bulk-fields (:index state))))
+   (get-in state [:props :refresh] false)))
+
+(s/defn produce
+  ([state :- ESConnState
+    slice-props :- SliceProperties
+    events :- [Event]]
+   (index-produce state slice-props events))
+
```
yogsototh: _style remark_: I think this is an unnecessary newline, while I think it is tolerated by https://github.com/bbatsov/clojure-style-guide#optional-new-line-after-fn-name all example given don't use any empty lines. So I would get rid of them when possible.

src/ctia/entity/event/crud.clj
```diff
@@ -0,0 +1,74 @@
...
+  ([state :- ESConnState
+    events :- [Event]]
+   (simple-produce state events)))
+
+(s/defn handle-create :- [Event]
+  "produce an event to ES"
+  [state :- ESConnState
+   events :- [Event]]
+
```
yogsototh: _style remark_: I think this is an unnecessary newline, while I think it is tolerated by https://github.com/bbatsov/clojure-style-guide#optional-new-line-after-fn-name all example given don't use any empty lines. So I would get rid of them when possible.

test/ctia/entity/event_test.clj
```diff
@@ -0,0 +1,242 @@
...
+                            :groups ["group1"],
+                            :confidence "High",
+                            :owner "user1"},
+                           :id
+                           "http://localhost:3001/ctia/event/event-00000000-0000-0000-0000-111111111121",
+                           :type "event",
+                           :event_type :record-deleted}]
+                         (:parsed-body (get (str "ctia/event/search?query=" q)
+                                            :content-type :json
+                                            :headers {"Authorization" "user1"})))))))))))))))
```
yogsototh: _remark_: I think if you could you should use Matt strategy to improve the readability and efficiency of such tests by providing a list of smaller checks instead of an equality with a very big object. Typically, check the count of element is correct. Check each element `:type` is equal to `"event"`, etc...

hshahman:
Verified in rel-v1.12

PR #746, opened 2018-10-23
Bundle import: Fix source_ref and target_ref resolution in relationships

msprunck:
Closes threatgrid/iroh#2207

- transient IDs  resolution in Relationhips has been fixed
- a Relationship entity cannot be created if the source or the target ref contains a transient ID
- all errors related to a bundle import are logged with the WARN level

QA:

1- Import a bundle with a Relationship containing transient IDs for the source and the target ref
2- Import the same bundle a second time
3- All entities should be imported once

### DIFF
--- a/src/ctia/bulk/core.clj
+++ b/src/ctia/bulk/core.clj
@@ -138,7 +138,8 @@
                        tempids
                        login
                        {:refresh refresh})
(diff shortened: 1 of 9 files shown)


hshahman:
Verified in rel-v1.13

PR #750, opened 2018-11-06
Improve Flow entity validation behavior

gbuisson:
closes https://github.com/threatgrid/iroh/issues/2237

This changes the flow validate-entities fn in 2 ways,

- validate TLP settings even if there is no spec for an entity
- don't savagely return http exceptions but properly descend them to the underlying caller

finally I added 2 custom http exception handlers, one for spec client errors and another for tlp client errors


hshahman:
Verified in rel-v1.13

PR #755, opened 2018-11-16
Recursively ignore schema_version

gbuisson:
We already ignore `schema_version` on our Individual documents,
this is replaced both at create and update flows with the latest `ctim-schema-version` at realize.

However we skipped that for sub entities, for instance in the `Bundle` schema or in `Casebooks`, this could lead to issues like #2183 or issues for Casebook when we'll deploy updates.

This PR fixes that by changing our internal CTIA `defschema` functions, walking schemas and replacing all occurrences of `schema_version` with an open string.

### DIFF
--- a/src/ctia/schemas/utils.clj
+++ b/src/ctia/schemas/utils.clj
@@ -1,0 +1,17 @@
+(ns ctia.schemas.utils
+  (:require [schema-tools.walk :as sw]
+            [schema.core :as s]))
+
+(defn recursive-open-schema-version
+  "walk a schema and replace all instances of schema version
+   with an open string schema"
+  [s]
+  (sw/walk
+   (fn [x]
+     (if (and
+          (instance? clojure.lang.IMapEntry x)
+          (= :schema_version (:k (first x))))
+       [(first x) s/Str]
+       (recursive-open-schema-version x)))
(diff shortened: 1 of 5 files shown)


src/ctia/schemas/utils.clj
```diff
@@ -0,0 +1,17 @@
...
+
+(defn recursive-open-schema-version
+  "walk a schema and replace all instances of schema version
+   with an open string schema"
+  [s]
+  (sw/walk
+   (fn [x]
+     (if (and
+          (instance? clojure.lang.IMapEntry x)
+          (= :schema_version (:k (first x))))
```
msprunck: I would replace that using `prewalk` and `map?`:

```Clojure
(defn recursive-open-schema-version
   "walk a schema and replace all instances of schema version
    with an open string schema"
   [s]
   (prewalk #(if (and (map? %)
                      (= :schema_version (:k (first x))))
              [(first x) s/Str]
              (recursive-open-schema-version %))
           m))
```

Could you add a test for the `recursive-open-schema-version` fn?

gbuisson: your version gives a stackoverflow so i keep mine ;-), I added the test

PR #758, opened 2018-11-23
Allow Regular users bundle import capability

gbuisson:
closes https://github.com/threatgrid/iroh/issues/2279
- Remove Casebook from Bunde, NewBundle
- Add the `bundle-import` capability for all users


hshahman:
Verified in rel-v1.18

PR #761, opened 2018-11-28
Use ctim 1.0.7

msprunck:
Connected to https://github.com/threatgrid/ctim/pull/223

> distinct :impact_score and :exploitability_score
data for version 2 and version 3 of the CVSS impact metrics.


msprunck:
Good catch @gbuisson, I updated the ES mapping

PR #769, opened 2019-01-15
Add a status route

gbuisson:
Just a simple route to check Auth and server status.

### DIFF
--- a/src/ctia/status/routes.clj
+++ b/src/ctia/status/routes.clj
@@ -1,0 +1,15 @@
+(ns ctia.status.routes
+  (:require
+   [compojure.api.sweet :refer :all]
+   [schema.core :as s]
+   [ctia.schemas.core :refer [StatusInfo]]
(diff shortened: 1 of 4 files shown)


src/ctia/status/routes.clj
```diff
@@ -0,0 +1,15 @@
...
+  (:require
+   [compojure.api.sweet :refer :all]
+   [schema.core :as s]
+   [ctia.schemas.core :refer [StatusInfo]]
+   [ring.util.http-response :refer [ok]]))
+
+(defroutes status-routes
+  (context "/status" []
+           :tags ["Status"]
+           (GET "/" []
```
msprunck: The purpose of this route is to test the authentication of the user. Providing a `whoami` endpoint could be better, what do you think?

gbuisson: I don't think so, the route is currently very basic but in the future we might add many more checks related to CTIA's health

PR #770, opened 2019-01-15
Display DELETE routes by default in Swagger UI

gbuisson:
This PR changes the default behaviour for our Swagger routes so DELETE calls are included in the spec.

QA Instructions:

- A DELETE route should be listed in Swagger for each entity API subset


hshahman:
Verified in rel-v1.16

PR #771, opened 2019-01-15
Lighter dev setup

gbuisson:
Disable containers we currently don't use to save system resources while coding.

### DIFF
--- a/src/ctia/http/routes/crud.clj
+++ b/src/ctia/http/routes/crud.clj
@@ -43,7 +43,7 @@
            can-patch?
            can-search?
            can-get-by-external-id?]
-    :or {hide-delete? true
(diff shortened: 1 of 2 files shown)


src/ctia/http/routes/crud.clj
```diff
@@ -43,7 +43,7 @@
            can-patch?
            can-search?
            can-get-by-external-id?]
-    :or {hide-delete? true
+    :or {hide-delete? false
```
msprunck: I don't know if this change should be part of this PR or could you add a comment in the PR's description?

gbuisson: it's inherited from my previous branch, that won't hurt though

PR #772, opened 2019-01-16
bundle/export enabled to specify relationship filters

ereteog:
connected to threatgrid/iroh/issues/2095

bundle/export: introduced an optional `related_to` parameter which enables to search relationships only by `target_ref` or `source_ref`

**QA**
WIthout precising `related_to` parameter or selecting both `target_ref` and `source_ref` the behavior should be the same as before.
If you set related_to to `source_ref` (respectively `target_ref`), the requested ids must be only in `source_ref` (respectively `target_ref`) of retrieved relationships.

### DIFF
--- a/src/ctia/bundle/core.clj
+++ b/src/ctia/bundle/core.clj
@@ -337,9 +337,8 @@
 
 (defn fetch-entity-relationships
   "given an entity id, fetch all related relationship"
-  [id identity-map]
-  (let [filters [{:source_ref id}
-                 {:target_ref id}]
+  [id identity-map related-to]
+  (let [filters (map #(hash-map % id)  related-to)
         rel-lists
         (map
          #(some->
@@ -368,10 +367,11 @@
   [id
    identity-map
    ident
-   {:keys [include_related_entities]
-    :or {include_related_entities true}}]
+   {:keys [include_related_entities related_to]
+    :or {include_related_entities true
+         related_to [:source_ref :target_ref]}}]
   (if-let [record (fetch-record id identity-map)]
-    (let [relationships (fetch-entity-relationships id identity-map)]
+    (let [relationships (fetch-entity-relationships id identity-map related_to)]
       (cond-> {}
         record
         (assoc (-> (:type record)
--- a/src/ctia/bundle/core.clj
+++ b/src/ctia/bundle/core.clj
@@ -338,7 +338,7 @@
 (defn fetch-entity-relationships
   "given an entity id, fetch all related relationship"
   [id identity-map related-to]
-  (let [filters (map #(hash-map % id)  related-to)
+  (let [filters (map #(hash-map % id) (set related-to))
         rel-lists
         (map
          #(some->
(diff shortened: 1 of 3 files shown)


src/ctia/bundle/core.clj
```diff
@@ -337,9 +337,8 @@
 
 (defn fetch-entity-relationships
   "given an entity id, fetch all related relationship"
-  [id identity-map]
-  (let [filters [{:source_ref id}
-                 {:target_ref id}]
+  [id identity-map related-to]
+  (let [filters (map #(hash-map % id)  related-to)
```
yogsototh: *minor* there are two spaces after `id)`.

src/ctia/bundle/routes.clj
```diff
@@ -14,6 +14,7 @@
 
 (s/defschema BundleExportQuery
   {:ids [s/Str]
+   (s/optional-key :related_to) [(s/enum :source_ref :target_ref)]
```
yogsototh: A simple remark, the actual schema authorize duplicate related_to values. If that isn't a problem for swagger/compojure-api, perhaps we could use a set instead of a list here.

test/ctia/bundle/routes_test.clj
```diff
@@ -505,7 +528,20 @@
...
+       (is (= 402 (count (:relationships bundle-get-res-4))))
+
+       (is (= (count (:sightings bundle-get-res-3))
+              (count (:sightings bundle-get-res-5))))
+       (is (= (count (:relationships bundle-get-res-3))
+              (count (:relationships bundle-get-res-5))))
+       (is (= (count (:indicators bundle-get-res-3))
+              (count (:indicators bundle-get-res-5))))
+
+       (is (empty? (:relationships bundle-from-source)))
```
msprunck: I think that you should check that the related entities (relationship, sighting) is returned if you export a bundle with one of the Indicator ID.

test/ctia/bundle/routes_test.clj
```diff
@@ -505,7 +528,20 @@
...
+
+       (is (every? #(contains? #{sighting-id-1 sighting-id-2}
+                               (:target_ref %))
+                   (:relationships bundle-from-target)))))))
```
msprunck: Same here, I would only export with one Sighting ID to check that only the related entities are returned.

test/ctia/bundle/routes_test.clj
```diff
@@ -505,7 +528,20 @@
...
        (is (nil? (:indicators bundle-get-res-4)))
-       (is (= 402 (count (:relationships bundle-get-res-4))))))))
+       (is (= 402 (count (:relationships bundle-get-res-4))))
+
+       (is (= (count (:sightings bundle-get-res-3))
+              (count (:sightings bundle-get-res-5))))
+       (is (= (count (:relationships bundle-get-res-3))
+              (count (:relationships bundle-get-res-5))))
+       (is (= (count (:indicators bundle-get-res-3))
+              (count (:indicators bundle-get-res-5))))
```
msprunck: It's not easier to compare the whole bundle?

test/ctia/bundle/routes_test.clj
```diff
@@ -505,7 +533,21 @@
...
+       (is (= bundle-get-res-3 bundle-get-res-5))
+
+       (is (= indicator-id (-> bundle-from-source
+                               :indicators
+                               first
+                               :id)))
+       (is (= sighting-id-2 (-> bundle-from-source
+                               :sightings
+                               first
+                               :id)))
```
msprunck: Could you check that only the expected entities are returned? The test becomes very difficult to read, could you add `testing` blocks or a message associated with some `is` assertions?

PR #780, opened 2019-01-21
Docker HUB docker-compose config (demo and client-dev)

gbuisson:
This PR changes docker-compose setup for both client-dev and demo so they use the image available at docker Hub, this make it possible to launch a demo instance of CTIA locally without setting anything besides docker.


saintx:
This is a big improvement.  Thank you!  LGTM

PR #784, opened 2019-02-04
Fix Event Modified status typo

### LINKED ISSUE #782: Event action "modified" field
Some events have the action value 'modifed' which should be 'modified'

https://github.com/threatgrid/ctia/blob/5d5e735f771181f98211d1e6eee2dfa3f331b5ec/src/ctia/entity/event/obj_to_event.clj#L32

gbuisson:
closes #782


hshahman:
Verified in rel-v1.18

PR #785, opened 2019-02-04
Incident GraphQL interface

### LINKED ISSUE #783: Graphql Incident support
We would like to be able to query for incidents via the graphql endpoint, like other types:

https://github.com/threatgrid/ctia/blob/master/src/ctia/graphql/schemas.clj

msprunck:
Closes #783


hshahman:
Verified in rel-v1.18

PR #786, opened 2019-02-04
Validate Bulk/Bundle imports with specs

gbuisson:
closes https://github.com/threatgrid/iroh/issues/2441

⚠️ This change may break existing imports as it makes Bundle/Bulk request checks stricter ⚠️ 

## Description 

This PR adds support for validating bulk/bundle entities using CTIM specs.
as a Bonus I also fixed the exception behaviour so it is uniform for all error cases, 
we now don't throw an error if one of the entities fail spec or top validation but simply return data about the failed entities.

## QA checks:

- Push a Bundle with Invalid fields such as relationship source_ref or target_ref, any entity valid_time
- Push a Bundle with Invalid TLP settings (white in Private CTIA)
- The Response should be either 201 or 200 for bulk/bundles with the incorrect entities marked as error
- Individual entity POST/PUT spec errors should still be 400 http status codes

### DIFF
--- a/src/ctia/flows/crud.clj
+++ b/src/ctia/flows/crud.clj
@@ -83,34 +83,43 @@
       (get tempids (:id entity))
       (when-let [entity-id (find-checked-id entity)]
         (when-not (auth/capable? identity :specify-id)
-          (throw (http-response/bad-request!
-                  {:error "Missing capability to specify entity ID"
-                   :entity entity})))
+          {:error "Missing capability to specify entity ID"
+           :entity entity})
         (if (id/valid-short-id? entity-id)
           entity-id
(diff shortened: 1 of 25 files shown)


src/ctia/flows/crud.clj
```diff
@@ -99,14 +99,16 @@
     (when spec
       (when-not (cs/valid? spec entity)
         (throw (ex-info (cs/explain-str spec entity)
```
msprunck: We should not throw an exception for `bulk` or `bundle` operations. Some entities could be created before the spec error occurs.

hshahman:
Verified in rel-v1.18

PR #789, opened 2019-02-06
Handle the bundle export when an external entity is referenced

### LINKED ISSUE #788: Empty warnings with a nil value are logged
Based on the log message:

```java
2019-02-05 13:41:34,538 WARN (qtp42167981-25 - /ctia/bundle/import?external-key-prefixes=hydrant-) [ctia.bundle.core] - nil
```

It seems to come from this fn

```clojure
(defn log-errors
  [response]
  (let [errors (->> response
                    :results
                    (map :error)
                    seq)]
    (doseq [error errors]
      (log/warn error)))
  response)
```

msprunck:
Closes threatgrid/iroh#2444
Closes #788 

This PR adds a test to check if an entity that is part of an exported bundle is stored in the local CTIA instance. Otherwise, it is ignored and will not be part of the exported bundle.

QA: Export a bundle containing a Relationship that references an external entity.

### DIFF
--- a/src/ctia/bundle/core.clj
+++ b/src/ctia/bundle/core.clj
@@ -267,8 +267,7 @@
   [response]
   (let [errors (->> response
                     :results
-                    (map :error)
-                    seq)]
+                    (keep :error))]
     (doseq [error errors]
       (log/warn error)))
   response)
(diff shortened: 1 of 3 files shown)


hshahman:
verified in rel-v1.18

PR #794, opened 2019-02-11
Swagger OAuth2 support

gbuisson:
This PR adds support for configuring OAuth2 authentication in the Swagger Interface.

screenshot 2019-02-11 at 17 53 48
screenshot 2019-02-09 at 00 23 13
screenshot 2019-02-11 at 17 54 09

- Setup API Key and OAuth2 Security definition
- Configurable through properties

### DIFF
--- a/src/ctia/http/handler.clj
+++ b/src/ctia/http/handler.clj
@@ -23,7 +23,8 @@
              [version :refer [wrap-version]]]
             [ctia.metrics.routes :refer [metrics-routes]]
             [ctia.observable.routes :refer [observable-routes]]
-            [ctia.properties :refer [properties]]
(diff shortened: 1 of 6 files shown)


src/ctia/http/handler.clj
```diff
@@ -77,81 +78,129 @@
...
+  [swagger-base-conf
+   {:keys [client-id
+           app-name
+           authorization-url
+           token-url
+           refresh-url
+           flow
+           realm
+           scopes
+           key]}]
```
msprunck: Could you rename it to avoid confusion with `clojure.core/key`?

hshahman:
Fantastic - it helps a lot.

PR #798, opened 2019-02-20
bundle/export: limit number of retrieved relationships

ereteog:
related to https://github.com/threatgrid/iroh/issues/2479
This PR introduced in bundle export, a limit to the number of relationships that can be retrieved per entity.
It also adds a POST bundle/export route for enabling large numbers of ids as parameters.

**QA**
- Bundle/export shall not return more than configured ctia.http.bundle.max-relationships property per retrieved entity.
- New POST bundle/export route must behave like GET bundle/export route.

### DIFF
--- a/resources/ctia-default.properties
+++ b/resources/ctia-default.properties
@@ -18,7 +18,7 @@ ctia.http.access-control-allow-origin=http://(localhost|127.0.0.1)(:\\d+)?
 ctia.http.access-control-allow-methods=get,post,put,patch,delete
 
 ctia.http.events.timeline.max-seconds=5
-ctia.http.bundle.max-relationships=1000
+ctia.http.bundle.export.max-relationships=1000
 
 ctia.http.jwt.enabled=true
 ctia.http.jwt.public-key-path=resources/cert/ctia-jwt.pub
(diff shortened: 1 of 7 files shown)


resources/ctia-default.properties
```diff
@@ -17,6 +17,9 @@ ctia.http.port=3000
 ctia.http.access-control-allow-origin=http://(localhost|127.0.0.1)(:\\d+)?
 ctia.http.access-control-allow-methods=get,post,put,patch,delete
 
+ctia.http.events.timeline.max-seconds=5
+ctia.http.bundle.max-relationships=1000
```
msprunck: Could you update the documentation https://github.com/threatgrid/ctia/blob/master/doc/properties.org with with new property?

msprunck: To be more explicit, could you rename the property to something like `ctia.http.bundle.max-exported-relationships` or `ctia.http.bundle.export.max-relationships`

src/ctia/bundle/routes.clj
```diff
@@ -25,40 +60,21 @@
...
+                (ok (export-bundle
+                     (:ids q)
+                     identity-map
+                     identity
+                     (select-keys q [:include_related_entities :related_to]))))
+
+           (POST "/export" []
+                :return NewBundle
+                :header-params [{Authorization :- (s/maybe s/Str) nil}]
+                :body [q BundleExportQuery]
```
msprunck: To show all available values in swagger UI, I would keep `:include_related_entities` and `:related_to` as query parameters.

PR #799, opened 2019-02-21
Add back the Authorization field to Swagger POST reqs

gbuisson:


hshahman:
It was part of the v1.18

PR #800, opened 2019-02-26
Make Bundle export schema sub entities open

gbuisson:
closes https://github.com/threatgrid/iroh/issues/2492


hshahman:
Will be verified in https://github.com/threatgrid/iroh/issues/2492

PR #803, opened 2019-02-27
Restrict Event DELETE route to developer capabilities

gbuisson:
closes https://github.com/threatgrid/iroh/issues/2502


hshahman:
Verified in rel-v1.21

Issue #807, opened 2019-03-11
bundle/export does not return relationship to a casebook


ereteog:
On this instance of CTIA: https://private.intel.int.iroh.site/i
the following incident: https://private.intel.int.iroh.site/ctia/incident/incident-61e49122-b66b-4a36-b640-58b9fbfc02a2 is targeted by 8 relationships with casebooks as source_ref.
However, these casebooks and relationships are not returned by bundle/export when queried with this incident id.
```clojure
[
  {
    "tlp": "amber",
    "schema_version": "1.0.8",
    "target_ref": "https://private.intel.int.iroh.site:443/ctia/incident/incident-61e49122-b66b-4a36-b640-58b9fbfc02a2",
    "type": "relationship",
    "id": "https://private.intel.int.iroh.site:443/ctia/relationship/relationship-ec491337-3493-4556-8ccd-f335b54eb8e6",
    "source_ref": "https://private.intel.int.iroh.site:443/ctia/casebook/casebook-9efdec1e-b082-4824-acaf-6774c1d97d02",
    "relationship_type": "related-to"
  },
  {
    "tlp": "amber",
    "schema_version": "1.0.9",
    "target_ref": "https://private.intel.int.iroh.site:443/ctia/incident/incident-61e49122-b66b-4a36-b640-58b9fbfc02a2",
    "type": "relationship",
    "id": "https://private.intel.int.iroh.site:443/ctia/relationship/relationship-c43339e5-a81e-48f3-b5e9-72354b49849e",
    "source_ref": "https://private.intel.int.iroh.site:443/ctia/casebook/casebook-7bf97d14-67b9-45b9-8104-189023f18444",
    "relationship_type": "related-to"
  },
  {
    "tlp": "amber",
    "schema_version": "1.0.9",
    "target_ref": "https://private.intel.int.iroh.site:443/ctia/incident/incident-61e49122-b66b-4a36-b640-58b9fbfc02a2",
    "type": "relationship",
    "id": "https://private.intel.int.iroh.site:443/ctia/relationship/relationship-483d0ac6-a345-4334-850e-b617102ff188",
    "source_ref": "https://private.intel.int.iroh.site:443/ctia/casebook/casebook-7bf97d14-67b9-45b9-8104-189023f18444",
    "relationship_type": "related-to"
  },
  {
    "tlp": "amber",
    "schema_version": "1.0.9",
[...]

PR #808, opened 2019-03-11
Bundle export documentation

ereteog:
closes https://github.com/threatgrid/ctia/issues/807

The problem was that the port was forgotten in the id of bundle/export route that works as expected.

### DIFF
--- a/src/ctia/bundle/routes.clj
+++ b/src/ctia/bundle/routes.clj
@@ -68,7 +68,7 @@
                 :return NewBundleExport
                 :header-params [{Authorization :- (s/maybe s/Str) nil}]
                 :query [q BundleExportQuery]
-                :summary "Export a record with its local relationships"
+                :summary "Export records with their local relationships. Ids must be URIs (with port if precised)."
                 :capabilities export-capabilities
                 :auth-identity identity
                 :identity-map identity-map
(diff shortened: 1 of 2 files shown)


src/ctia/bundle/routes.clj
```diff
@@ -83,7 +83,7 @@
                 :header-params [{Authorization :- (s/maybe s/Str) nil}]
                 :query [q BundleExportOptions]
                 :body [b BundleExportIds]
-                :summary "Export a record with its local relationships"
+                :summary "Export records with their local relationships. Ids must be URIs (with port if precised)."
```
msprunck:
```suggestion
                :summary "Export records with their local relationships. IDs are URIs (with port if precised)."
```

src/ctia/bundle/routes.clj
```diff
@@ -68,7 +68,7 @@
                 :return NewBundleExport
                 :header-params [{Authorization :- (s/maybe s/Str) nil}]
                 :query [q BundleExportQuery]
-                :summary "Export a record with its local relationships"
+                :summary "Export records with their local relationships. Ids must be URIs (with port if precised)."
```
msprunck:
```suggestion
                :summary "Export records with their local relationships. IDs are URIs (with port if precised)."
```

PR #809, opened 2019-03-12
Empty request es body

ereteog:
closes https://github.com/threatgrid/iroh/issues/2511
closes https://github.com/threatgrid/iroh/issues/2544

- Removes dead code
- More verbose logging for investigation
- fix empty bodies sent to ES
- fix bundle/export when some relationships reference deleted entities

**QA**
- import a bundle with an incident, a sighting and a relationship linking them. 
- delete imported incident
- request bundle/export with the previously imported sighting, it shall return this sighting and the relationship.


src/ctia/http/exceptions.clj
```diff
@@ -34,10 +34,24 @@
   (logging/log! :error e (ex-message e))
   (ex/response-validation-handler e data request))
 
+(defn es-ex-message
+  [e data request]
+  (str
```
gbuisson: I would prefer that we log a map instead of concatenating a string like this, this will not show up great on Kibana

src/ctia/http/exceptions.clj
```diff
@@ -34,10 +34,21 @@
...
+(defn es-ex-data
+  [e data request]
+  (into
+   {:message (.getMessage e)
+    :stacktrace (->> (.getStackTrace e)
+                     (mapv str))
+    :request request
+    :data data}
+   (when (instance? ExceptionInfo e)
+     {:ex-data (ex-data e)})))
```
ereteog: I have finally solved the bug, but shouldn't we let this log enrichment?

msprunck: The message and the stacktrace are already logged when you put the exception as the 1st argument.  I would only add the `request`, `data` and `ex-data`

```Clojure
(defn es-ex-data
  [e data request]
  (let [exception-data (ex-data e)]
    (cond-> {:request request
             :data data}
      exception-data
      (assoc :ex-data exception-data))))
```

hshahman:
Verified in rel-v1.20

PR #810, opened 2019-03-19
Es migration 2556

ereteog:
closes https://github.com/threatgrid/iroh/issues/2556

Enable to restart migration where it stops and handle writings (update and deletes) during migration.
The number of parameter has augmented so it now uses tool.cli for command line parameters.
```
  -i, --id ID                  id of the migration state to create or restart
  -p, --prefix PREFIX          prefix of the newly created indices 
  -m, --migrations MIGRATIONS  a comma separated list of migration ids to apply
  -b, --batch-size SIZE        migration batch size
      , --buffer-size SIZE       max number of batches in buffer between source and target
  -s, --stores STORES          comma separated list of stores to migrate
  -c, --confirm                really do the migration?
  -r, --restart                restart ongoing migration?
  -h, --help  
```

- apply migration `1.0.0` with prefix `1.0.1` to stores `tool` and `malware` with batches of size `1000`, and assign `migration-1`as id:
`lein run -m ctia.task.migration.migrate-es-stores -p 1.0.1 -m 1.0.0 -c -i migration-1 -s tool,malware -b 1000`
- add `-r`to restart that migration, behavior should be slightly modified to enable to restart when only migration-id, batch-size and restart parameters
`lein run -m ctia.task.migration.migrate-es-stores -m 1.0.0 -c -i migration-1 -b 1000 -r`

<a name="qa">[§](#qa)</a> QA
============================

No QA specifically needed on this PR. It is only about the migration process, not the transformations of documents and indices' mappings. However, when the migration will be applied on test, CTIA should be fully retested.

This PR includes a migration.md document, you should read it and validate its clarity.

- migration.md describes how to migrate CTIA stores.

### DIFF
--- a/src/ctia/task/migration/migrate_es_stores.clj
+++ b/src/ctia/task/migration/migrate_es_stores.clj
@@ -135,8 +135,7 @@
    confirm? :- s/Bool
    restart? :- s/Bool]
   (let [migration-state (if restart?
-                          (mst/restart-migration migration-id
-                                                 @mst/migration-es-conn)
+                          (mst/get-migration migration-id @mst/migration-es-conn)
                           (mst/init-migration migration-id prefix store-keys confirm?))
         migrations (compose-migrations migrations)
         batch-size (or batch-size default-batch-size)]
--- a/src/ctia/task/migration/migrate_es_stores.clj
+++ b/src/ctia/task/migration/migrate_es_stores.clj
@@ -42,7 +42,7 @@
 (s/defn handle-deletes
   [{:keys [created stores]} :- mst/MigrationSchema
    store-keys :- [s/Keyword]
-   confirm? :- [s/Bool]]
+   confirm? :- s/Bool]
   (let [deletes (mst/fetch-deletes store-keys created)]
     (doseq [[entity-type entities] deletes]
       (log/infof "Handling %s deleted %s during migration"
--- a/src/ctia/task/migration/migrate_es_stores.clj
+++ b/src/ctia/task/migration/migrate_es_stores.clj
@@ -44,15 +44,20 @@
 (s/defn handle-deletes
   [{:keys [created stores]} :- mst/MigrationSchema
    store-keys :- [s/Keyword]
+   batch-size :- s/Int
    confirm? :- s/Bool]
-  (let [deletes (mst/fetch-deletes store-keys created)]
-    (doseq [[entity-type entities] deletes]
-      (log/infof "Handling %s deleted %s during migration"
-                 (count entities)
-                 (name entity-type))
-      (when confirm?
-        (mst/batch-delete (get-in stores [entity-type :target :store])
-                          (map :id entities))))))
+  (loop [search_after nil]
+    (let [{:keys [data paging]} (mst/fetch-deletes store-keys created batch-size search_after)
+          {new-search-after :sort next :next} paging]
+      (doseq [[entity-type entities] data]
+        (log/infof "Handling %s deleted %s during migration"
+                   (count entities)
+                   (name entity-type))
+        (when confirm?
+          (mst/batch-delete (get-in stores [entity-type :target :store])
+                            (map :id entities))))
+      (when next
+        (recur new-search-after)))))
 
 (defn append-coerced
   [coercer coerced entity]
@@ -88,6 +93,10 @@
                (:type source-store)
                source-store-size)
 
+    (when (and confirm? (not started))
+      (mst/update-migration-store migration-id
+                                  entity-type
+                                  {:started (time/now)}))
     (loop [current-search-after search_after
            migrated-count migrated-count]
       (let [{:keys [data paging]} (mst/fetch-batch source-store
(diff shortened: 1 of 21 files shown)


test/ctia/task/migration/store_test.clj
```diff
@@ -0,0 +1,304 @@
...
+
+  (es-index/delete! es-conn "test_*"))
+
+(deftest fetch-deletes-test
+  (whoami-helpers/set-whoami-response "45c1f5e3f05d0"
+                                      "foouser"
+                                      "foogroup"
+                                      "user")
+  (post-bulk examples)
+  (Thread/sleep 1000) ;; ensure indices refresh
```
yogsototh: It would be safer with a loop looking at the data that cool fail after too much attempt (after 10s for example).

yogsototh:
Please slash manually locally and use the following template in your commit message (not in GitHub):

https://raw.githubusercontent.com/threatgrid/iroh/master/.github/pull_request_template.md?token=AAAW5S6LFTDWGXXNOBKQXNK4YF4MS

In particular I think you should really add an Ops block and if copy any documentation in a .org file inside the repository.

src/ctia/task/migration/migrate_es_stores.clj
```diff
@@ -0,0 +1,289 @@
...
+      (exit true))
+    (log/info "migration complete")
+    (catch Exception e
+      (log/error e "Unexpected error during migration")
+      (exit true)))
+  (exit false))
+
+(def cli-options
+  ;; An option with a required argument
+  [["-i" "--id ID" "The idID of the migration state to create or restar"
```
msprunck:
```suggestion
  [["-i" "--id ID" "The ID of the migration state to create or restart"
```

src/ctia/task/migration/migrate_es_stores.clj
```diff
@@ -0,0 +1,289 @@
...
+      (log/infof "checking store: %s" sk)
+      (check-store sr batch-size))))
+
+(defn exit [error?]
+  (if error?
+    (System/exit -1)
+    (System/exit 0)))
+
+(defn run-migration
+  [migration-id prefix migrations store-keys batch-size confirm? restart?]
```
msprunck: IMO the `restart?` option is risky because there is no way to stop the migration process gently. You could register a shutdown hook to the java runtime to stop the process properly.

test/ctia/task/migration/store_test.clj
```diff
@@ -0,0 +1,304 @@
...
+        "fetch-deletes shall return tool1 and tool2 that were deleted after since")
+    (is (= #{(:id sighting2)} (set (map :id sighting)))
+        "fetch-deletes shall return only entites that were deleted after since")
+    (is (nil? malware) "fetch-deletes shall only retrieve entity types given as parameter"))
+  (es-index/delete! es-conn "ctia_*"))
+
+
+(deftest init-get-migration-test
+  (post-bulk examples)
+  (Thread/sleep 1000) ; ensure indices refresh
```
msprunck: It could be better to use the `refresh?` param instead of a `Thread/sleep`

test/ctia/task/migration/store_test.clj
msprunck: It could be better to use the refresh? param instead of a Thread/sleep

src/ctia/task/migration/migrate_es_stores.clj
```diff
@@ -0,0 +1,289 @@
...
+  "migrate all es store indexes"
+  [migration-id :- s/Str
+   prefix :- s/Str
+   migrations :- [s/Any]
+   store-keys :- [s/Keyword]
+   batch-size :- s/Int
+   confirm? :- s/Bool
+   restart? :- s/Bool]
+  (let [migration-state (if restart?
+                          (mst/get-migration migration-id @mst/migration-es-conn)
```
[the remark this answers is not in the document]
ereteog: Even if a store was fully migrated during previous execution, writes could have occurred before this restart and thus you must handle it anyway.

test/ctia/task/migration/store_test.clj
ereteog: there is no `refresh?` parameter on that function that directly request a local CTIA bulk route. I finally triggered manually the refresh.

src/ctia/task/migration/store.clj
```diff
@@ -0,0 +1,392 @@
...
+                                      (name entity)
+                                      migration-id
+                                      nil)
+        coerce (coerce-to-fn MigrationSchema)]
+    (when-not migration-raw
+      (log/errorf "migration not found: %s" migration-id)
+      (throw (ex-info "migration not found" {:id migration-id})))
+    (-> (coerce migration-raw)
+        (update :stores enrich-stores)
+        (store-migration es-conn))))
```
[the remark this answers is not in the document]
ereteog: I store it because, the size of the source size could have changed before restart. It is updated in enrich-stores, thus I store it here.

src/ctia/task/migration/migrate_es_stores.clj
ereteog: The restart of the migration of a store is mainly based on the search_after field in the source store. In the target store, the ES ids of documents are preserved from source to target store. Thus in the worst case, document will be migrated twice but it will only erased previously written documents.

PR #811, opened 2019-03-25
Fail capability red test.

yogsototh:
Close https://github.com/threatgrid/iroh/issues/2581

The ability to set the identity was also granted for normal users. This PR fix that.


hshahman:
Verified in rel-v1.21

PR #812, opened 2019-03-29
issue-2599: fix wrong error status code

yogsototh:
Related to https://github.com/threatgrid/iroh/issues/2599

Reverse 401 / 403. 401 should be used when authentication is required and has failed or has not yet been provided and 403 when user might not have the necessary permissions.

QA
==

Check that authentication errors (Expired JWT, invalid JWT) return a 401 and that permission errors (missing capability, access control) return 403.

Release Notes
=============

```
intern: Fix usage of 401 and 403 statuses
```


hshahman:
Verified in rel-v1.22

PR #814, opened 2019-04-15
Sighting sensor coordinates

jyoverma:
Connected to threatgrid/ctim#237 and threatgrid/ctim#238

Changes to es stores mapping for adding Sensor Coordinates in Sighting. This change is to support the data model change in CTIM

Upgrade to CTIM 1.0.11 (sensor coordinates and data table in Sighting)

QA:
1. Post a new Sighting with sensor coordinates and a data table
2. Retrieve the same entity and check that the data is the same
3. Retrieve an old entity without error

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -44,7 +44,7 @@
                   :exclusions [prismatic/plumbing
                                potemkin
                                com.andrewmcveigh/cljs-time]]
-                 [threatgrid/ctim "1.0.9"
+                 [threatgrid/ctim "1.0.11"
                   :exclusions [threatgrid/flanders
                                metosin/ring-swagger
                                com.google.guava/guava]]
--- a/src/ctia/entity/sighting/es_store.clj
+++ b/src/ctia/entity/sighting/es_store.clj
@@ -25,7 +25,6 @@
      {:observed_time em/valid-time
       :count {:type "long"}
       :sensor em/token
-      :sensor_coordinates em/sighting-sensor
       :targets em/sighting-target
       :reference em/token
       :confidence em/token
--- a/src/ctia/entity/sighting/es_store.clj
+++ b/src/ctia/entity/sighting/es_store.clj
@@ -31,6 +31,7 @@
       :confidence em/token
       :severity em/token
       :resolution em/token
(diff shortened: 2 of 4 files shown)


yogsototh:
@jyoverma you need to sync all repository.

Here is the classical workflow:

1. You first need to merge https://github.com/threatgrid/ctim/pull/238 (note that we are waiting some answer from SSE team)
2. once merged, a new release version must be created and published to clojars
3. You need to change the dependency version of CTIM in the `project.clj` file in this repository
4. The dependency of CTIM should certainly be updated in the `iroh` repository too.

Once everything is done, the CI shall fail because some code need to be updated. I'm thinking about the GraphQL part and maybe other part of the code.
I think we could take care of it ourselves, I'm not sure you'll be allowed to publish to clojars.

Unfortunately as we are waiting to merge that PR on CTIM. In order not to be blocked, here is a development workflow, which is a bit more difficult.

Dev workflow:

1. on your local computer, go in your git feature branch. Check the version of CTIM is a SNAPSHOT one (currently it is "1.0.11-SNAPSHOT"). You do not need to push that SNAPSHOT version to clojars unless you want to make the tests in the CI, furthermore, SNAPSHOT version can be overwritten, so it should be preferable to only install that version locally during the development phase.
2. install locally the new package: in the `ctim` local repository do a `lein install`.
3. Then go to your local clone of CTIA. Change the version in `project.clj` of `ctim` to use that SNAPSHOT version.
4. Launch the tests locally:
    - `docker-compose -f containers/dev/docker-compose.yml up -d`
    - `lein test :all`
5. Now some tests might fail, you shall then fix them locally.

Once all the tests passe, we shall simply wait for the CTIM to be merged, then use a release version (not suffixed by `-SNAPSHOT`) and publish it to clojars. Then we shall only replace, in our branch the `ctim` version from the SNAPSHOT one to the release one. And we'll then be able to merge that branch.

Note that a similar workflow should be done for the `iroh` repository. But in IROH it is almost certain that no code change will be needed.

hshahman:
Verified in rel-v1.23

PR #815, opened 2019-04-17
Add the timestamp field to all entities

msprunck:
> Closes threatgrid/iroh#2607

This PR uses the `default-realise-fn` for all entities. It ensures that all common fields are added when the entity is realized. Only the addition of specific fields are defined in the `realize-judgement` and `realize-sighting` fns.

<a name="qa">[§](#qa)</a> QA
============================

1. Post a Judgement and a Sighting without the `timestamp` field
2. Retrieve the stored entities and check that the `timestamp` has been added automatically

### DIFF
--- a/src/ctia/entity/judgement/schemas.clj
+++ b/src/ctia/entity/judgement/schemas.clj
@@ -2,7 +2,7 @@
   (:require [clj-momo.lib.time :as time]
             [ctia.domain
              [access-control :refer [properties-default-tlp]]
-             [entities :refer [schema-version make-valid-time]]]
+             [entities :refer [default-realize-fn]]]
             [ctia.schemas
              [utils :as csu]
              [core :refer [def-acl-schema def-stored-schema TempIDs]]
@@ -14,7 +14,8 @@
             [ring.util.http-response :as http-response]
             [schema-tools.core :as st]
             [schema.core :as s]
-            [ctim.schemas.incident :as is]))
+            [ctim.schemas.incident :as is]
+            [ctia.flows.schemas :refer [with-error]]))
 
 (def-acl-schema Judgement
   js/Judgement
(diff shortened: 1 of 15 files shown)


src/ctia/entity/sighting/schemas.clj
```diff
@@ -32,34 +32,21 @@
...
+(def sighting-default-realize
+  (default-realize-fn "sighting" NewSighting StoredSighting))
+
 (s/defn realize-sighting :- StoredSighting
-  ([new-sighting :- NewSighting
-    id :- s/Str
-    tempids :- (s/maybe TempIDs)
-    owner :- s/Str
-    groups :- [s/Str]]
+  ([new-sighting id tempids owner groups]
```
ereteog: same remark on parameters' schemas?

src/ctia/entity/sighting/schemas.clj
```diff
@@ -32,34 +32,21 @@
...
    (realize-sighting new-sighting id tempids owner groups nil))
-  ([new-sighting :- NewSighting
-    id :- s/Str
-    tempids :- (s/maybe TempIDs)
-    owner :- s/Str
-    groups :- [s/Str]
-    prev-sighting :- (s/maybe StoredSighting)]
+  ([new-sighting id tempids owner groups prev-sighting]
    (let [now (time/now)]
```
ereteog: now is not used anymore, you can remove the `let`

src/ctia/entity/judgement/schemas.clj
```diff
@@ -37,42 +38,34 @@
...
-                        (:tlp prev-object (properties-default-tlp)))
-             :schema_version schema-version}
-            (make-valid-time (:valid_time prev-object)
-                             (:valid_time new-judgement)
-                             now)))))
+    prev-judgement :- (s/maybe StoredJudgement)]
+   (let [{:keys [error] :as disposition}
+         (try
+           (determine-disposition-id new-judgement)
+           (catch clojure.lang.ExceptionInfo e
```
yogsototh: Wouldn't it be safe to check the `ex-data` of the `ExceptionInfo` here to be certain this is a Mismatching disposition problem at the source of the exception?

src/ctia/http/exceptions.clj
```diff
@@ -94,6 +94,14 @@
       :entity entity
       :class (.getName (class e))})))
 
+(defn realize-entity-error-handler
+  "Handle error at the realize step"
+  [^Exception e data request]
+  (logging/log! :info e (ex-message e))
```
yogsototh: `ex-message` was apparently added in Clojure version 1.10: https://clojure.github.io/clojure/clojure.core-api.html but we are using Clojure 1.9: https://github.com/threatgrid/ctia/blob/master/project.clj#L31

I suggest to use the old `(.getMessage e)`.

msprunck: `ex-message` is already used in other error handlers. We will probably have to rename it when we move to Clojure 1.10

src/ctia/entity/judgement/schemas.clj
msprunck: Yes it would be better but it was not in the scope of this PR

hshahman:
Verified in rel-v1.23

PR #820, opened 2019-05-17
improve textual search

ereteog:
> / Close https://github.com/threatgrid/iroh/issues/2286

* `AND` as default search operator, this behavior is configurable and could be set to `OR` to have previous behavior. If needed, it can be configured per store as for other es properties.
* fixed existing tests on search which could not be reached.
* fixed mapping of vulnerability `description` field.
* fixed default analyzer that was not properly set (which is used for _all field).
* removed dead code.
* english possessive stemmer to match `browser's` when loooking for `browser`.
* better tokenizer, to enable to search on domain components (like `cisco` matches `www.cisco.com`).
* plural / singular management, `word` matches both `word` and `words`.

search in CTIA now use `AND` as default behavior when entering multiple keywords like in popular search engines.

<a name="qa">[§](#qa)</a> QA
============================

for each describable entities (entities with `title` and `description` fields), like actors, malwares, vulnerabilities:

-  NOW: `AND` is now the default operator for search queries, all words in queries must be matched in entities like sightings or judgments.

- AFTER MIGRATION:
1. insert instances with texts in `description` including english possessive and capitalized letters "this Chrome's vulnerability allow an attacker to EXECUTE arbitrary code, http://www.talos.com".
2. entity search route MUST match these instances with following queries:
- `chrome vulnerability allow`
- `Chrome's vulnerability code`
- `chrome execute`
- `Chrome`
- `attacker attackers`
3. entity search route MUST NOT match these instances with following queries:
- `chrome missingWord`
- `cchrome`
- `"chrome  allow an attacker"`
- `talos`
- `www.talos.com`

- Managing the english possessive require an 'identity' migration likde for https://github.com/threatgrid/ctia/pull/738

- Public documentation update needed

```
public: Improved textual search. In particular, `AND` is now the default search operator, textual search must match all query keywords.
```


resources/ctia-default.properties
```diff
@@ -51,8 +51,10 @@ ctia.store.es.default.port=9200
 ctia.store.es.default.indexname=ctia
 ctia.store.es.default.replicas=1
 ctia.store.es.default.refresh_interval=1s
+ctia.store.es.default.default_operator=AND
 ctia.store.es.default.shards=5
 ctia.store.es.default.refresh=false
+ctia.store.es.default.refresh=false
```
msprunck:
```suggestion
```

src/ctia/properties.clj
```diff
@@ -33,6 +33,7 @@
    (str "ctia.store.es." store ".refresh_interval")  s/Str
    (str "ctia.store.es." store ".replicas") s/Num
    (str "ctia.store.es." store ".shards") s/Num
+   (str "ctia.store.es." store ".default_operator") (s/enum "OR" "AND")
```
msprunck: I don't think we want a different behaviour for each store. IMO, a global is sufficient

test/ctia/test_helpers/search.clj
```diff
@@ -2,29 +2,142 @@
...
+(defn search-ids
+  ([search-uri query] (search-ids search-uri
+                                  query
+                                  {"Authorization" "45c1f5e3f05d0"}))
+  ([search-uri query headers]
+   (->> (search search-uri query headers)
+        :parsed-body
+        (map :id)
+        set)))
+
```
msprunck: Could you remove one carriage return? (@gbuisson is very strict about that)

hshahman:
There is one more thing to check as mentioned here:

image

PR #821, opened 2019-06-03
Entity and lookup sort Improvements

gbuisson:
This PR Adds 
- new sort options for Sighting and Judgements, allowing to easily fetch latest objects
- Changes Bundle export so it includes objects ordered from latest Relationships first

> **Epic** https://github.com/threatgrid/iroh/issues/2696

<a name="qa">[§](#qa)</a> QA
============================

1. POST multiple `Judgements` and `Sightings` with different`valid_time`, `observed_time` and `timestamps`
2. query using the new sort choices
3. your responses should be correctly ordered based on the sort choice

4. POST an entity and link it to another one multiple times, exceeding the Bundle export limit
5. Retrieved entities should be the most recently linked

```
public: New sort choices to get the latest Sightings and Judgements
```

### DIFF
--- a/src/ctia/entity/sighting.clj
+++ b/src/ctia/entity/sighting.clj
@@ -9,10 +9,13 @@
             [schema.core :as s]))
 
 (def sighting-sort-fields
-  (apply s/enum ss/sighting-fields))
+  (apply s/enum (map name ss/sighting-sort-fields)))
+
+(def sighting-fields
+  (apply s/enum (map name ss/sighting-fields)))
 
 (s/defschema SightingFieldsParam
-  {(s/optional-key :fields) [sighting-sort-fields]})
+  {(s/optional-key :fields) [sighting-fields]})
 
 (s/defschema SightingSearchParams
   (st/merge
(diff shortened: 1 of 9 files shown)


src/ctia/entity/sighting.clj
```diff
@@ -24,18 +27,15 @@
...
    PagingParams
    SightingFieldsParam
    {(s/optional-key :sort_by)
-    (s/enum
-     :id
-     :timestamp
-     :confidence
-     :observed_time.start_time)}))
+    (apply s/enum
+           (map name ss/sighting-sort-fields))}))
```
ereteog: you can directly use `sighting-sort-fields` which is defined above.

folbricht:
Verified in 1.28 with sightings using this method:

- Generate 50 random sightings for a sha256, all with random timestamps (start and end time)
- Post all sightings to PCTIA
- Sort the sightings by timestamp locally
- Search for sightings with sort_by=observed_time.start_time&sort_order=desc
- Compare the locally sorted list with the list from the API. All should match

PR #824, opened 2019-06-18
Identity Assertion CTIA changes

jyoverma:
> **Epic** threatgrid/iroh#2760
> Close threatgrid/iroh#2763

Creating a Store, CRUD & Search routes to support the new IdentityAssertion CTIM Entity

Put all informations that need to be communicated to IROH Services Clients.
Typically IROH-UI, ATS Integration, Orbital, etc...

<a name="qa">[§](#qa)</a> QA
============================

1. Identity Assertion CRUD operations can be tested from the swagger UI (if available)
An example identity assertion object is as below:
```
{
  "authorized_groups": [
    "string"
  ],
  "valid_time": {
    "start_time": "2019-06-19T16:50:21.940Z",
    "end_time": "2019-06-19T16:50:21.940Z"
  },
  "identity": {
    "type": "endpoint",
    "observables": [
      {
        "value": "test.com",
        "type": "domain"
      }
    ],
    "observed_time": {
      "start_time": "2019-06-19T16:50:21.940Z",
      "end_time": "2019-06-19T16:50:21.940Z"
    },
    "os": "Mac OS"
  },
  "schema_version": "string",
  "revision": 0,
  "authorized_users": [
    "string"
  ],
  "type": "identity-assertion",
  "source": "jyoverma",
  "external_ids": [
    "string"
  ],
  "external_references": [
    {
      "source_name": "string",
      "description": "string",
      "url": "string",
      "hashes": [
        "string"
      ],
      "external_id": "string"
    }
  ],
  "source_uri": "string",
  "language": "string",
  "tlp": "amber",
  "timestamp": "2019-06-19T16:50:21.940Z",
  "assertions": [
    {
      "name": "group",
      "value": "foo"
    }
  ]
}
```
2. Run a search for the the assertions.key, assertions.value, identity.observable.type and identity.observable.value if identity assertion is available in Swagger UI
3. if identity assertion is not available in Swagger UI, run tests via CLI

- Config change needed: threatgrid/tenzin# NO
- Migration needed: threatgrid/tenzin# NO
- How to enable/disable that feature: (ex remove service from `bootstrap.cfg`, add scope to org)

- Public documentation updated needed: threatgrid/iroh-ui#
  See internal [doc file](./services/iroh-auth/doc/public-doc.org)

```
intern: Short description for internal eyes
public: Short description for public release notes.
```


project.clj
```diff
@@ -44,7 +44,7 @@
                   :exclusions [prismatic/plumbing
                                potemkin
                                com.andrewmcveigh/cljs-time]]
-                 [threatgrid/ctim "1.0.11"
+                 [threatgrid/ctim "1.0.12-SNAPSHOT"
```
msprunck: This version of CTIM is not deployed on clojars with `lein deploy clojars`

jyoverma: Done.

src/ctia/stores/es/mapping.clj
```diff
@@ -118,6 +118,11 @@
    {:type token
     :value all_token}})
 
+(def assertion
+  {:properties
+   {:name all_text
+    :value all_text}})
```
ereteog: when you use the mapping `all_text`in CTIA it enables 2 things:
1. search on _all fields meaning that you can search without specifying the field that it should match.
2. search on words in the corresponding text.

As a consequence, for the name `cisco:ctr:group`from that [example](https://github.com/threatgrid/ctim/blob/master/src/ctim/examples/identity_assertions.cljc#L31), the following queries should match that text: 
* assertions.name:"cisco:ctr:group"
* assertions.name:"cisco ctr"
* "cisco:ctr:group"
* "cisco ctr"
* ctr
* cisco
* group
note that `cisco:ctr:group`will generate an error because it's a reserved word for lucene queries.

If you use `all_token`, it should only match:
* assertions.name:"cisco:ctr:group"
* "cisco:ctr:group"

If you use the versions without the `all_`prefix it will force to specify the searched field: `assertions.name:...`

Thus, IMHO we should have:
``` clojure
(def assertion
  {:properties
   {:name all_token
    :value all_text}})
```

test/ctia/bundle/routes_test.clj
```diff
@@ -232,6 +243,7 @@
                        :source "source"
                        :indicators (set indicators)
                        :sightings (set sightings)
+                       :identity_assertions (set identity_assertions)
```
ereteog: shouldn't it be `:identity-assertions`?

jyoverma:
Issues addressed by https://github.com/threatgrid/ctia/pull/872

PR #828, opened 2019-06-25
Kafka Event publisher

gbuisson:
This PR adds an event handler that pushes Events to a Kafka topic.
Target is to support Amazon MSK.

> **Epic** https://github.com/threatgrid/iroh/issues/2745

Highlights: 

- Can be enabled/setup using CTIA properties
- SSL Support
- Compression configuration
- The Kafka topic is automatically created at startup using the supplied configuration
- The docker-compose setup has been updated to include zookeeper and Kafka images from confluent
- Removed unused docker-compose test setup

No QA needed

No ops needed

```
intern: Added a CTIA kafka event publisher
```


src/ctia/flows/hooks/event_hooks.clj
```diff
@@ -31,6 +52,20 @@
     (->RedisEventPublisher (lr/server-connection redis-config)
                            channel-name)))
 
+(defn kafka-event-publisher []
+  (let [kafka-topic-config
+        (get-in @properties [:ctia :hook :kafka :topic])]
+
+    (log/warn "Setting up Kafka topic")
+    (try
+      (lk/create-topic)
```
msprunck: You could use `kafka.admin.AdminUtils/listTopics` to check if the topic already exists

src/ctia/flows/hooks/event_hooks.clj
```diff
@@ -31,6 +52,20 @@
...
+      (catch org.apache.kafka.common.errors.TopicExistsException e
+        (log/warn "Kafka Topic already exists")))
```
msprunck:
```suggestion
        (log/info "Kafka Topic already exists")))
```

src/ctia/flows/hooks/event_hooks.clj
```diff
@@ -31,6 +52,20 @@
...
+
+    (->KafkaEventPublisher
+     (lk/build-producer)
+     kafka-topic-config)))
```
msprunck: The publisher doesn't need the topic config, the topic name is sufficient.

src/ctia/lib/kafka.clj
```diff
@@ -0,0 +1,99 @@
...
+  (let [producer-opts {}
+        {:keys [request-size]}
+        (get-in @properties [:ctia :hook :kafka])
+        {:keys [session-timeout
+                connection-timeout
+                operation-retry-timeout
+                address] :as zk-config}
+        (get-in @properties [:ctia :hook :kafka :zk])
+        brokers (opk/find-brokers {:kafka/zookeeper address})
+        kafka-config (merge {"bootstrap.servers" brokers
```
msprunck: Should we configure a `gzip` compression with the `compression.type` property?

doc/properties.org
```diff
@@ -148,6 +148,22 @@
...
+|--------------------------------------------+-------------------------------------+-----------------|
+| ctia.hook.kafa.request-size                | set the Kafa max request size       | number          |
+| ctia.hook.kafa.zk.session-timeout          | Zookeeper session timeout           | number          |
+| ctia.hook.kafa.zk.connection-timeout       | Zookeeper connection timeout        | number          |
+| ctia.hook.kafa.zk.connection-retry-timeout | Zookeeper connection retry timeout  | number          |
+| ctia.hook.kafa.zk.address                  | Zookeeper address                   | string          |
+| ctia.hook.kafa.topic.name                  | The topic name to push messages to  | string          |
+| ctia.hook.kafa.topic.num-partitions        | Setup the topic partition countDown | string          |
+| ctia.hook.kafa.topic.replication-factor    | Setup the topic replication-factor  | string          |
+
```
msprunck: We should probably allow `ssl.` configuration for in-transit encryption.

src/ctia/flows/hooks/event_hooks.clj
gbuisson: I do both now, so if another process tries to create the topic after the check it's still caught.

src/ctia/lib/kafka.clj
```diff
@@ -0,0 +1,123 @@
...
+        {:keys [session-timeout
+                connection-timeout
+                operation-retry-timeout
+                address] :as zk-config}
+        (get-in @properties [:ctia :hook :kafka :zk])
+        _ (clojure.pprint/pprint compression-type)
+        brokers (opk/find-brokers {:kafka/zookeeper address})
+        kafka-config (cond-> {"bootstrap.servers" brokers
+                              "max.request.size" request-size}
+                       (ssl-enabled?) (merge (make-ssl-opts))
```
msprunck:
```suggestion
                       (ssl-enabled?) (into (make-ssl-opts))
```

When we only have two maps, we usually use `into` instead of `merge`

src/ctia/lib/kafka.clj
```diff
@@ -0,0 +1,122 @@
...
+    (let [records (.poll consumer timeout)]
+      (doseq [record (.records records name)]
+        (f (okh/consumer-record->message decompress
+                                         record)))
+      (.commitSync consumer)
+      (recur))))
+
+(defn subscribe [f timeout]
+  (let [{:keys [name]}
+        (get-in @properties [:ctia :hook :kafka :topic])
```
ereteog:
```clojure
name (get-in @properties [:ctia :hook :kafka :topic :name])
```

containers/dev/docker-compose.yml
```diff
@@ -11,26 +11,41 @@ services:
...
+      ZOOKEEPER_SERVER_ID: 1
+      ZOOKEEPER_CLIENT_PORT: 2181
+      ZOOKEEPER_TICK_TIME: 2000
+      ZOOKEEPER_INIT_LIMIT: 5
+      ZOOKEEPER_SYNC_LIMIT: 2
+      ZOOKEEPER_SERVERS: localhost:2181:2181
+    ports:
+      - "2181:2181"
+
+  kafka:
```
yogsototh: The current configuration does not seem to be able to work with a `scale` via docker-compose. I think we should create an issue to enhance that to be able to find potential bug only present with multiple brokers.

src/ctia/flows/hooks/event_hooks.clj
yogsototh: I think using the try / create is safer in a potentially concurrent system. I'm not against using the listTopics but I'm not sure it will be much nicer.

src/ctia/lib/kafka.clj
```diff
@@ -0,0 +1,119 @@
...
+        (get-in @properties [:ctia :hook :kafka :zk])
+        brokers (opk/find-brokers {:kafka/zookeeper address})
+        kafka-config (cond-> {"bootstrap.servers" brokers
+                              "group.id" "ctia"}
+                       (ssl-enabled?) (merge (make-ssl-opts)))]
+    (okh/build-consumer kafka-config
+                        (okh/byte-array-deserializer)
+                        (okh/byte-array-deserializer))))
+
+(defn poll [consumer name f timeout]
```
yogsototh: It would be nice to have a docstring here. It is not perfectly clear to me what `consumer`, `name`, `f` and `timeout` are here.

In particular it seems that `f` is a handler so it should be preferable to rename it `handler` instead or `handle-event`.

src/ctia/lib/kafka.clj
```diff
@@ -0,0 +1,119 @@
...
+     :consumer-thread consumer-thread}))
+
+(defn stop-consumer
+  [{:keys [consumer
+           consumer-thread]}]
+  (.stop consumer-thread))
+
+(defn create-topic []
+  (let [{:keys [address]}
+        (get-in @properties [:ctia :hook :kafka :zk])
```
yogsototh: `address (get-in @properties [:ctia :hook :kafka :zk :address])`

src/ctia/lib/kafka.clj
```diff
@@ -0,0 +1,119 @@
...
+
+(defn kafka-delete-topic [zk-addr
+                          topic-name]
+  (with-open [zk-utils (okh/make-zk-utils {:servers zk-addr} false)]
+    (AdminUtils/deleteTopic zk-utils
+                            topic-name)))
+
+(defn delete-topic []
+  (let [{:keys [address]}
+        (get-in @properties [:ctia :hook :kafka :zk])
```
yogsototh: same remark about using `:address` instead of `:keys` for a single key.

test/ctia/flows/hooks/kafka_event_hook_test.clj
```diff
@@ -0,0 +1,167 @@
...
+          (get-in @properties [:ctia :hook :kafka])
+          consumer-map
+          (lk/subscribe
+           (fn test-events-kafka-topic-fn [ev]
+             (let [v (:value ev)]
+               (swap! results conj (parse-string v true))
+               (.countDown finish-signal)))
+           30000)
+          ;; await rebalance
+          _ (Thread/sleep 5000)]
```
yogsototh: We've been bitten multiple time by those waiting systems inside Travis CI, I would suggest to try to find a better way if possible.

test/ctia/flows/hooks/kafka_event_hook_test.clj
```diff
@@ -0,0 +1,167 @@
...
+                         :valid_time {:start_time "2016-02-11T00:40:48.212-00:00"}})
+
+            judgement-3-id
+            (id/long-id->id judgement-3-long-id)]
+
+        (is (= 201 judgement-1-status))
+        (is (= 201 judgement-2-status))
+        (is (= 201 judgement-3-status))
+
+        (is (.await finish-signal 5 TimeUnit/SECONDS)
```
yogsototh: kind of the same remark about waiting here, travis CI can be tricky about that, perhaps you would be safer by increasing that value a lot.

PR #829, opened 2019-07-02
Fix a NPE while attempting to convert an ID on Bulk Export

gbuisson:
This PR fixes an NPE occurring on bulk export

> / Close https://github.com/threatgrid/iroh/issues/2810

The Bulk export process read entities and attempt to convert retrieved entity IDs to URLs, previous behaviour didn't check that the entity was existing before attempting to convert the ID, thus leading to a NPE, this PR fixes that and add a test to check the behaviour.

<a name="qa">[§](#qa)</a> QA
============================

1. Attempt to call Bundle/Export with an unexisting entity ID
2. You shouldn't get a 500

Nothing

Nothing

```
intern: Fix a Bug in CTIA Bulk export 
```

### DIFF
--- a/src/ctia/bulk/core.clj
+++ b/src/ctia/bulk/core.clj
@@ -69,11 +69,14 @@
   "Retrieve many entities of the same type provided their ids and common type"
   [ids entity-type auth-identity]
   (let [read-entity (read-fn entity-type auth-identity {})]
-    (map (fn [id] (try (with-long-id
-                         (read-entity id))
-                       (catch Exception e
-                         (do (log/error (pr-str e))
-                             nil)))) ids)))
+    (map (fn [id]
+           (try
+             (if-let [entity (read-entity id)]
+               (with-long-id entity)
+               nil)
+             (catch Exception e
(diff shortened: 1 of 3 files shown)


src/ctia/bulk/core.clj
```diff
@@ -69,11 +69,15 @@
...
-    (map (fn [id] (try (with-long-id
-                         (read-entity id))
-                       (catch Exception e
-                         (do (log/error (pr-str e))
-                             nil)))) ids)))
+    (map (fn [id]
+           (try
+             (if-let [entity (read-entity id)]
+               (with-long-id entity)
+               (do (log/warnf "bulk/read-entities: Entity not found: %s" id)
```
msprunck: Please don't log this case. It happens for all target entities of Relationships that are not in the same CTIA instance.

hshahman:
To verify this issue:
- Import one sighting which is linked to some indicators via relationship
- Delete one of the indicators
- Call bundle export endpoint for the sighting
==> should not return 500

hshahman:
Verified in rel-v1.28

PR #846, opened 2019-09-13
Added more security headers

yogsototh:
> Close https://github.com/threatgrid/iroh/issues/2942

- took the liberty to replace the full list of CORS headers by just '*'.
  C.f. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Expose-Headers
- took the liberty to be more restrictive than we are for headers for the UI

<a name="qa">[§](#qa)</a> QA
============================

Check that calls for the index.html page for swagger UI contains the following headers:

`Content-Type-Security`, `X-Frame-Options`, and `X-Content-Type-Options` and that everything work as expected (you can use swagger-ui to make requests mostly.

```
intern: More security headers
```

### DIFF
--- a/src/ctia/http/server.clj
+++ b/src/ctia/http/server.clj
@@ -112,6 +112,7 @@
            max-threads
            min-threads
            port
+           trusted-urls
            access-control-allow-origin
            access-control-allow-methods
            jwt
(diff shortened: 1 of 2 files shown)


src/ctia/http/server.clj
```diff
@@ -102,16 +102,28 @@
...
+                           ;; the handler can take precedence on the value
+                           ;; of the headers
+                           (into headers response-headers)))))))
 
 (defn- new-jetty-instance
   [{:keys [dev-reload
            max-threads
            min-threads
            port
+           trusted-urls
```
msprunck: Not used

hshahman:
Verified in rel-v1.34

PR #853, opened 2019-09-27
Migration by time interval

ereteog:
* The sort is too slow on very large indices like `relationships`, `sightings` or `events`, this PR enables to slice per increasing time intervals in order to sort only on smaller datasets.
  * in `ctia.task.migration.store` the `sliced-queries` and `format-buckets` function creates a list of filtering queries that will be used to perform the migration per batch.
  * in `ctia.task.migration.migrate-es-store`, this PR isolate the reading and writing in source and target into two seperated functions.
  * Dedicated tests were added for each introduced function, using real sample data.
  * a new global tests has been added to tests with real data that the slicing is well performed and that no data are lost with restart.
* remove `core.async`

# QA

No QA needed


project.clj
```diff
@@ -50,7 +51,7 @@
                                com.google.guava/guava
                                org.clojure/tools.reader
                                org.clojure/clojurescript]]
-                 [threatgrid/clj-momo "0.3.0"]
+                 [threatgrid/clj-momo "0.3.1-SNAPSHOT"]
```
gbuisson: I think you know but just to make sure you have to release it before merging

project.clj
```diff
@@ -31,6 +31,7 @@
   :dependencies [[org.clojure/clojure "1.9.0"]
                  [clj-time "0.15.1"]
                  [org.clojure/core.async "0.3.465" :exclusions [org.clojure/tools.reader]]
+                 [clj-http-fake "1.0.3"]
```
gbuisson: this should be deported to test dependencies

src/ctia/task/migration/migrate_es_stores.clj
```diff
@@ -75,6 +75,79 @@
   (let [coercer (coerce-to-fn Model)]
     #(reduce (partial append-coerced coercer) {} %)))
 
+(defn read-source
+  "This function retrieves in `source-store` all the documents that match the given `query`
+   by batch of maximum size `batch-size`. When not nil the `search_after` parameter is used
+   to skip previously retrieved data. Then it pushes the retrieved data in the given channel,
+  `data-chan`, along with next search_after value for eventual further restart. Once all the
```
gbuisson:
```suggestion
  `data-chan`, along with next search_after value for any further restart. Once all the
```

src/ctia/task/migration/migrate_es_stores.clj
```diff
@@ -75,6 +75,79 @@
...
+           migrated-count
+           migration-es-conn
+           confirm?]}]
+  (loop [current-count migrated-count]
+    ;; sync loop that consumes batches in chan and migrates them to target
+    (if-let [{:keys [documents search_after]} (<!! data-chan)]
+      (let [migrated (transduce migrations conj documents)
+           {:keys [data errors]} (list-coerce migrated)
+           new-current-count (+ current-count (count data))]
+        (doseq [entity errors]
```
gbuisson: I would test the errors list for emptiness first, it feels weird in the code, like you should always have errors

src/ctia/task/migration/migrate_es_stores.clj
```diff
@@ -75,6 +75,79 @@
...
+                (format "%s - Cannot migrate entity: %s"
+                        (name entity-type)
+                        (pr-str entity))]
+            (log/error message)))
+        (when confirm?
+          (when (seq data) (mst/store-batch target-store data))
+          (mst/rollover target-store batch-size new-current-count)
+          (mst/update-migration-store migration-id
+                                      entity-type
+                                      (into {:target {:migrated new-current-count}}
```
gbuisson: it would look better if you use `cond->` start with `{:target {:migrated new-current-count}}` and assoc only if `search_after` is non nil

src/ctia/task/migration/migrate_es_stores.clj
```diff
@@ -104,59 +185,28 @@
...
+           migrated-count migrated-count-state]
+      (if-let [query (first queries)]
+        (let [migration-params (assoc base-params
+                                      :query query
+                                      :migrated-count migrated-count
+                                      :data-chan (chan buffer-size))]
+          (log/infof "%s - handling sliced query %s"
+                     (name entity-type)
+                     (pr-str query))
+          (thread (read-source migration-params))
```
gbuisson: here if I'm not mistaken you treat each query sequentially,
you first `read-source` then finally  you `write-target` on `recur`.
there is no need for the `thread` since you do it one by one and in this regard you could totally pass an atom instead of a chan or even use `write source` passing the result of read-source as a var.

after investigating the process I believe your approach is valid though but if we could remove an unnecessary threading logic from the loop that would help understand what's happening at a glance.

test/ctia/test_helpers/es.clj
```diff
@@ -134,3 +155,28 @@
 
 (defn post-all-to-es [objects]
   (run! post-to-es objects))
+
+(defn str->edn
```
gbuisson: the naming of the function is not totally accurate, the final result is not an edn string

src/ctia/task/migration/store.clj
```diff
@@ -260,6 +262,119 @@ Rollover requires refresh so we cannot just call ES with condition since refresh
...
+  Last range does not contains end date limit in order to handle documents
+  that are created or modified during the migration. Note that new entities are now
+  created with creation value as first modified value"
+  [{:keys [conn indexname mapping]} :- StoreMap
+   search_after :- [s/Any]
+   interval :- Interval]
+  (let [agg-field (case mapping
+                    "event" :timestamp
+                    :modified)
+        query (when (some->> (first search_after)
```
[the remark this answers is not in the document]
msprunck: This query will be used each time the migration is restarted? Is it right?

src/ctia/task/migration/migrate_es_stores.clj
```diff
@@ -75,6 +77,109 @@
...
+        new-migrated-count (+ migrated-count (count data))]
+    (doseq [entity errors]
+      (let [message
+            (format "%s - Cannot migrate entity: %s"
+                    (name entity-type)
+                    (pr-str entity))]
+        (log/error message)))
+    (when confirm?
+      (when (seq data) (mst/store-batch target-store data))
+      (mst/rollover target-store batch-size new-migrated-count)
```
msprunck: I don't see the usage of `batch-size` and `new-migrated-count` in the `rollover` fn

test/ctia/task/migration/store_test.clj
```diff
@@ -123,6 +294,154 @@
...
 
+(deftest sliced-queries-test
+  (let [storemap {:conn es-conn
+                  :indexname "ctia_relationship"
+                  :mapping "relationship"
+                  :props {:write-index "ctia_relationship"}
+                  :type "relationship"
+                  :settings {}
+                  :config {}}
+        data (es-helpers/load-file-bulk es-conn "./test/data/indices/sample-relationships-1000.json")
```
msprunck: What is the difference between adding 1000 relationships or 19 relationships to test this function?

src/ctia/task/migration/migrate_es_stores.clj
ereteog: blessed catch!

src/ctia/task/migration/migrate_es_stores.clj
```diff
@@ -75,6 +77,109 @@
...
+    :as batch-params} :- (s/maybe BatchParams)]
+    ;; loop that reads batches from source and produces them in chan
+  (when batch-params
+    (let [{:keys [data paging]} (mst/query-fetch-batch query
+                                                       source-store
+                                                       batch-size
+                                                       0
+                                                       "asc"
+                                                       search_after)
+          next-search-after (:sort paging)]
```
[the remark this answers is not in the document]
ereteog: It's copy / paste inheritance from former code.
I have renamed this funciton in `read-source-batch` and added a `read-source` functon that generates the lazy sequence. This last function is simpler using `sort` which is always returned by clj-momo pagination function. I tried to use `(-> paging :next :search_after)` but when the last batch is smaller than `batch-size`, [`search_after` is nil](https://github.com/threatgrid/clj-momo/blob/master/src/clj_momo/lib/es/pagination.clj#L30), and if I pass `nil` in the next batch it will restarts from the beginning for given query. 


by `(-> paging :next :search_after)`, but it doesn't work, I think that this is a bug here: https://github.com/threatgrid/clj-momo/blob/master/src/clj_momo/lib/es/pagination.clj#L30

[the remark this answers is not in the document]
msprunck: `:sort` and `:next :search_after` are set with the value of `sort` in `clj-momo`. I do not understand how they can have a different value.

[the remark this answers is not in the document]
ereteog: here https://github.com/threatgrid/clj-momo/blob/master/src/clj_momo/lib/es/pagination.clj#L45

src/ctia/task/migration/store.clj
```diff
@@ -260,6 +262,119 @@ Rollover requires refresh so we cannot just call ES with condition since refresh
...
+(def missing-date (time-coerce/to-date-time missing-date-str))
+
+(defn missing-bucket?
+  "Returns true if the date is related to missing date.
+   Depending on date interval the date of missing bucket returned
+   by ES could be either equal or before missing date."
+  [date-str]
+  (or (= date-str missing-date-str)
+      (-> (time-coerce/to-date-time date-str)
+          (time/before? missing-date))))
```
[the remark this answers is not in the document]
ereteog: I extended the doc-string, is it better?

src/ctia/task/migration/store.clj
```diff
@@ -260,6 +262,119 @@ Rollover requires refresh so we cannot just call ES with condition since refresh
...
+           queries []]
+      (if-let [date (first dates)]
+        (->>
+         (cond
+           (missing-bucket? date) (missing-query field)
+           (next dates) (range-query field date unit)
+           :else (last-range-query field date false))
+         (conj queries)
+         (recur (next dates)))
+        queries))))
```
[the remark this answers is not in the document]
ereteog: There is different exception cases to handle.
I have done a new version without loop-recur, but it's hard to handle these exception with a clearer code. Doing so I discovered some edge cases that were not considered. I have added corresponding tests.
Let me now if it's ok this way.

test/ctia/task/migration/store_test.clj
ereteog: I have selected these relationships among 4 millions to be sufficiently random and representative (including entities with missing `modified`field). 
I checked with`time`, it takes 300ms to load these relationships with bulk, and the whole test takes 400ms.

msprunck: What is the purpose of this test? How I know that that each case is tested? If the format the Relationship entity change how can I generate a new file? IMO we should not use data like this for a unit test.

src/ctia/task/migration/store.clj
```diff
@@ -256,10 +259,126 @@ Rollover requires refresh so we cannot just call ES with condition since refresh
...
+               "week" "w"
+               "day" "d")
+        filtered (->> raw-buckets
+                      (filter #(< 0 (:doc_count %)))
+                      (map :key_as_string))
+        [first-date & rest-dates] filtered
+        queries (map #(range-query % field unit)
+                     (drop-last filtered))
+        last-query (some-> (last filtered)
+                           (last-range-query field false))]
```
msprunck: You could use this function to simplify

```clojure
(defn update-last
  [v f]
  (conj (pop v) (f (peek v))))

(update-last [1 2] inc)
```

src/ctia/task/migration/store.clj
```diff
@@ -256,10 +259,126 @@ Rollover requires refresh so we cannot just call ES with condition since refresh
...
+    (cond-> queries
+      last-query (concat [last-query])
+      (and first-date
+           (missing-bucket? first-date)) (->> rest
+                                              (cons (missing-query field))))))
```
msprunck: If the vector contains only one date corresponding to the missing bucket, I think that the result will contain two query. Is it expected?

src/ctia/task/migration/migrate_es_stores.clj
```diff
@@ -75,6 +77,109 @@
...
+        (log/error message)))
+    (when confirm?
+      (when (seq data) (mst/store-batch target-store data))
+      (mst/rollover target-store batch-size new-migrated-count)
+      (mst/update-migration-store migration-id
+                                  entity-type
+                                  (cond-> {:target {:migrated new-migrated-count}}
+                                    search_after (assoc :source
+                                                        {:search_after search_after}))
+                                  migration-es-conn))
```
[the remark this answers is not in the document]
ereteog: I will monitor any problem about that in INT. 
If I detect something, I would use `refresh="true"` to avoid too slow requests, and since the migration index is very small (1 document after the first migration :-)), it will be fast.

src/ctia/task/migration/store.clj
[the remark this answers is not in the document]
ereteog: I extended doc-string and added a comment. Let me know if it's ok.

src/ctia/task/migration/store.clj
ereteog: it will not.
I added a tests about that case: https://github.com/threatgrid/ctia/pull/853/files#diff-3c037a41bf474dc42f251773926680bfR171

src/ctia/task/migration/store.clj
```diff
@@ -256,10 +259,135 @@ Rollover requires refresh so we cannot just call ES with condition since refresh
...
+        filtered (->> raw-buckets
+                      (filter #(< 0 (:doc_count %)))
+                      (map :key_as_string))
+        [first-date & rest-dates] filtered
+        queries (map #(range-query % field unit)
+                     (drop-last filtered))
+        last-query (some-> (last filtered)
+                           (last-range-query field false))]
+    (cond-> queries
+      last-query (concat [last-query])
```
msprunck: Why do you use `concat` instead of `(conj last-query)`?

test/ctia/task/migration/store_test.clj
ereteog: the only thing it depends on is the modified field. whatever will be future format of relationship, if it keeps `modified` metadata, which is controlled by CTIA, it is it's OK.

src/ctia/task/migration/store.clj
ereteog:
```clojure
ctia.main=> (update-last [] inc)
IllegalStateException Can't pop empty vector  clojure.lang.PersistentVector.pop (PersistentVector.java:468)
```

src/ctia/task/migration/store.clj
ereteog: queries is a `list`.
``` clojure
;; notice conjoining to a list is done at the beginning
(conj '(1 2 3) 4)
;;=> (4 1 2 3)
```
I could use `mapv` for building queries to then use `conj`.

src/ctia/task/migration/migrate_es_stores.clj
```diff
@@ -75,6 +77,115 @@
...
+               :search_after next-search-after)))))
+
+(s/defn read-source ;; WARNING: defining schema output breaks lazyness
+  "returns a lazy-seq of batch from source store"
+  [read-params :- (s/maybe BatchParams)]
+  (->>
+   (lazy-seq
+    (when-let [batch (read-source-batch read-params)]
+      (cons batch (read-source batch))))
+   (take-while seq)))
```
msprunck: What is the purpose of the `take-while`?

PR #854, opened 2019-10-08
Add OAuth2 URLs to the connect-src CSP

msprunck:
Fix OAuth2 code grant in swagger UI

> Close threatgrid/iroh#2981

This PR fixes the OAuth2 code flow in swagger UI by adding the `token-url` and the `refresh-url` to the Content Security Policy header.

<a name="qa">[§](#qa)</a> QA
============================

1. Log into private/public CTIA with the OAuth2 authentication method
2. Execute the query `/ctia/status`
3. The result status should be 200

```
intern: Fix OAuth2 authentication in Swagger UI
```

### DIFF
--- a/src/ctia/http/server.clj
+++ b/src/ctia/http/server.clj
@@ -126,6 +126,20 @@
         (update :headers (fn [response-headers]
                            (into headers response-headers)))))))
 
+(defn build-csp
+  "Build the Content Security Policy header from the http configuration"
+  [{:keys [swagger] :as http-config}]" connect-src 'self';"
+  (str "default-src 'self';"
+       " style-src 'self' 'unsafe-inline';"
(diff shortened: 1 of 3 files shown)


src/ctia/http/server.clj
```diff
@@ -126,6 +126,20 @@
         (update :headers (fn [response-headers]
                            (into headers response-headers)))))))
 
+(defn build-csp
+  "Build the Content Security Policy header from the http configuration"
+  [{:keys [swagger] :as http-config}]" connect-src 'self';"
```
ereteog:
```suggestion
  [{:keys [swagger] :as http-config}]
```

hshahman:
Verified in rel-v1.35

PR #855, opened 2019-10-14
Wait_for on update / patch / delete

ereteog:
This PR enables to use the `wait_for` query parameter for not only Creation of resources, but also update, patch, delete, bulk.

> Close #https://github.com/threatgrid/iroh/issues/2957
> Related #https://github.com/threatgrid/iroh/issues/2672

In order to enable to pass the wait_for param from routes to store, this PR modifies the protocal IStore with a new `params` parameter for `update-record` and `delete-record` functions. 

<a name="qa">[§](#qa)</a> QA
============================

- Check UPDATE / PATCH
  - Perform a  loop that repeats the following test 100 times: 
  1. Update/Patch any entity with `wait_for=true` as query param
  2. Retrieve the response ID and generate the short version of that ID (last part of the the url)
  3. Retrieve that resource using the short ID
  4. Check that the resource is well updated/patched

  - Perform a loop that repeats the following test 100 times: 
  1. Update/Patch any entity with`wait_for=false` as query param
  2. Retrieve the response ID and generate the short version of that ID (last part of the the url)
  3. Retrieve that resource using the short ID
  4. Check the returned entity: most should not be updated/patched (in some lucky cases some can be)

  - Perform a loop that repeats the following test 100 times: 
  1. Update/Patch any entity without the `wait_for` query param
  2. Retrieve the response ID and generate the short version of that ID (last part of the the url)
  3. Retrieve that resource using the short ID
  4. Check the returned entity: most should not be updated/patched (in some lucky cases some can be)

- Check DELETE
  - Perform a  loop that repeats the following test 100 times: 
  1. Delete any entity with `wait_for=true` as query param
  2. Retrieve the response ID and generate the short version of that ID (last part of the the url)
  3. Retrieve that resource using the short ID
  4. Check that the response status is 404

  - Perform a loop that repeats the following test 100 times: 
  1. Delete any entity with`wait_for=false` as query param
  2. Retrieve the response ID and generate the short version of that ID (last part of the the url)
  3. Retrieve that resource using the short ID
  4. Check the returned status: most of the status of the GET should be 200 (in some lucky cases some can be 404)

  - Perform a loop that repeats the following test 100 times: 
  1. Delete any entity without the `wait_for` query param
  2. Retrieve the response ID and generate the short version of that ID (last part of the the url)
  3. Retrieve that resource using the short ID
  4.  Check that the response status is 404

- Check Bulk
  - Perform a  loop that repeats the following test 100 times: 
  1. Bulk import entities with `wait_for=true` as query param
  2. Retrieve reponse Ids and generate the short version of these IDs (last part of the the urls)
  3. Retrieve the resources using these short IDs
  4. Check that the response statuses are 200

  - Perform a loop that repeats the following test 100 times: 
  1. Bulk import entities with `wait_for=true` as query param
  2. Retrieve reponse Ids and generate the short version of these IDs (last part of the the urls)
  3. Retrieve the resources using these short IDs
  4. Check the returned status: most of the status of the GET should be 404 (in some lucky cases some can be 200)

```
intern: When updating / deleting a resource or bulk / bundle importing, the query param `wait_for=true` could be used to deal with inconsistencies.
```


src/ctia/bundle/routes.clj
```diff
@@ -97,7 +98,8 @@
                  :query-params
                  [{external-key-prefixes
                    :- (describe s/Str "Comma separated list of external key prefixes")
-                   nil}]
+                   nil}
+                  {wait_for :- (describe s/Bool "wait for created entities to be available for search") nil}]
```
msprunck: I would not expose this query param for the bundle import route. It must be set to true due to the usage of a `pmap` in the `ctia.bulk.core` ns.

test/ctia/bulk/routes_test.clj
```diff
@@ -182,6 +182,59 @@
                               (core/get bulk-ids type))))
              (keys bulk-ids))))
 
+
+(deftest test-bulk-wait_for-test
```
msprunck: I would only use `cli-http-fake` to check if the refresh param is correctly set in the ElasticSearch request

test/ctia/test_helpers/crud.clj
```diff
@@ -117,25 +122,101 @@
...
+          (is (some #(= 200 %)
+                    (repeatedly 3 #(get-status false)))
+             "Delete queries should not wait for index refresh when wait_for is false")
+          (testing "Configured ctia.store.es.default.refresh value is applied when wait_for is not specified"
+            (if (= "false" (get-in @properties [:ctia :store :es :default :refresh]))
+
+              (is (some #(= 200 %)
+                        (repeatedly 3 #(get-status nil))))
+              (is (= 404 (get-status nil)))))))
+
```
msprunck: I would only use `cli-http-fake` to check if the refresh param is correctly set in the ElasticSearch request

hshahman:
Verified in earlier releases

PR #857, opened 2019-10-16
Remove Travis AWS IAM creds from .travis.yml

danlsgiga:
This PR removes the Travis AWS IAM credentials from the .travis.yml file. The credentials will now be set in the Repository Settings via Travis Web UI [here](https://travis-ci.org/threatgrid/ctia/settings).

This will help OPs in future credential changes with:

- Not having to push PRs to change the credentials
- Not having developers having to cherry-pick the master commit to their own branches to use the new credentials
- Not having Travis running builds and tests just for a credential change
- Potentially recycling credentials at a faster cadence without impacting any of the teams

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.


danlsgiga:
Please don't merge it yet... I'm considering changing the way these creds are injected in Travis. Instead of it being in .travis.yml, set it in the Travis WebUI Settings. this way OPs can change the creds at will without the need to push PRs to all projects and also, avoid branches that are still using the old creds to fail in future runs.

PR #859, opened 2019-11-18
update setting task

ereteog:
> Closes https://github.com/threatgrid/iroh/issues/3030

This PR introduces a new task to update indices settings while also updating indices settings at startup. It also fixes a bug at index creation, `refresh_interval` is no more ignored.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

CTIA proposes a new task to update dynamic settings like `refresh_interval` and `number_of_replicas`. For that purpose, you must first update the property file with the desired values for these settings (by default or per store) and then launch the task with:
`java -cp ctia.jar:resources:. clojure.main -m ctia.task.settings`
or from source:
`lein run -m ctia.task.settings` 
You can also select some stores with the option `--stores` and a comma separated list of store names:
`lein run -m ctia.task.settings --stores relaitonship,malware,indicator`

### DIFF
--- a/src/ctia/task/settings.clj
+++ b/src/ctia/task/settings.clj
@@ -1,12 +1,12 @@
 (ns ctia.task.settings
+  (:import clojure.lang.ExceptionInfo)
   (:require [clojure.tools.cli :refer [parse-opts]]
             [clojure.tools.logging :as log]
-
             [schema.core :as s]
             [clj-momo.lib.es
              [index :as es-index]
              [schemas :refer [ESConnState]]]
-            [ctia.stores.es.init :refer [upsert-template!]]
+            [ctia.stores.es.init :refer [StoreProperties init-es-conn! get-store-properties]]
             [ctia
              [init :refer [init-store-service! log-properties]]
              [properties :refer [properties init!]]
@@ -14,16 +14,14 @@
 
 (s/defn update-store!
   "read store properties of given stores and update indices settings."
-  [{:keys [conn index config]
-    {:keys [settings]} :config
-    :as state} :- ESConnState]
+  [props :- StoreProperties]
   (try
-    (upsert-template! conn index config)
-    (log/info "updated template: " index)
-    (->> {:index (select-keys settings [:refresh_interval :number_of_replicas])}
-         (es-index/update-settings! conn index))
-    (log/info "updated settings: " index)
-    (catch clojure.lang.ExceptionInfo e
+    (let [{:keys [conn index]
+           {:keys [settings]} :config} (init-es-conn! props)]
+      (->> {:index (select-keys settings [:refresh_interval :number_of_replicas])}
+           (es-index/update-settings! conn index))
+      (log/info "updated settings:" index))
+    (catch ExceptionInfo e
       (log/warn "could not update settings on that store"
                 (pr-str (ex-data e))))))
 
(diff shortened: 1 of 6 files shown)


src/ctia/task/settings.clj
```diff
@@ -0,0 +1,57 @@
...
+             [properties :refer [properties init!]]
+             [store :refer [stores]]]))
+
+(s/defn update-store!
+  "read store properties of given stores and update indices settings."
+  [{:keys [conn index config]
+    {:keys [settings]} :config
+    :as state} :- ESConnState]
+  (try
+    (upsert-template! conn index config)
```
msprunck: The template is already updated when the store is initialised in the `init-es-conn!` fn.

src/ctia/task/settings.clj
```diff
@@ -0,0 +1,57 @@
+(ns ctia.task.settings
+  (:require [clojure.tools.cli :refer [parse-opts]]
+            [clojure.tools.logging :as log]
+
```
msprunck: Could you remove this trailing carriage return?

src/ctia/task/settings.clj
```diff
@@ -0,0 +1,57 @@
...
+    :as state} :- ESConnState]
+  (try
+    (upsert-template! conn index config)
+    (log/info "updated template: " index)
+    (->> {:index (select-keys settings [:refresh_interval :number_of_replicas])}
+         (es-index/update-settings! conn index))
+    (log/info "updated settings: " index)
+    (catch clojure.lang.ExceptionInfo e
+      (log/warn "could not update settings on that store"
+                (pr-str (ex-data e))))))
```
msprunck:
```suggestion
               (log/error e "could not update settings on that store")
```

It's an more an error than a warn IMO

src/ctia/task/settings.clj
```diff
@@ -0,0 +1,57 @@
...
+
+(defn update-stores!
+  [store-keys]
+  (doseq [kw store-keys]
+    (log/infof "updating settings for store: %s" (name kw))
+    (update-store! (-> @stores kw first :state))))
+
```
msprunck: Could you remove this trailing carriage return?

src/ctia/task/settings.clj
```diff
@@ -0,0 +1,57 @@
...
+  [{:keys [conn index config]
+    {:keys [settings]} :config
+    :as state} :- ESConnState]
+  (try
+    (upsert-template! conn index config)
+    (log/info "updated template: " index)
+    (->> {:index (select-keys settings [:refresh_interval :number_of_replicas])}
+         (es-index/update-settings! conn index))
+    (log/info "updated settings: " index)
+    (catch clojure.lang.ExceptionInfo e
```
msprunck: Could you import `clojure.lang.ExceptionInfo` instead of using the full classname?

test/ctia/task/settings_test.clj
```diff
@@ -0,0 +1,38 @@
+(ns ctia.task.settings-test
+  (:require [clojure.test :refer [deftest is testing join-fixtures use-fixtures]]
+            [ctia.task.settings :as sut]
+            [ctia.stores.es.init :as init]
+            [clj-momo.lib.es.index :as es-index]))
+
```
msprunck: Could you remove this trailing carriage return?

PR #860, opened 2019-11-20
Owner and Group fields as part of API responses

gbuisson:
> **Epic** #
> Close https://github.com/threatgrid/iroh/issues/3040
> Related #

This PR modifies all our API responses, including also the graphQL types to return the datastore fields detailing ownership (owner and groups) so to allow customers to get ownership info.

=============================

All CTIA routes now include `owner` and `group` fields in each record as part of API responses,
be careful to not supply them in the case of an Update as they are still not accepted for Update/Create requests.

<a name="qa">[§](#qa)</a> QA
============================

1. Query a document or search any document
2. All documents should contain the `owner` and `group` fields

```
intern: Private and Public Intel now return ownership data as part of the records.
```


src/ctia/domain/entities.clj
```diff
@@ -88,11 +88,8 @@
     (long-id->entity-type id-str)))
 
 (defn un-store [record]
-  (let [entity-fields [:created :modified :owner :groups]
-        event-fields [:created :modified]]
-    (case (:type record)
-      "event" (apply dissoc record event-fields)
-      (apply dissoc record entity-fields))))
+  (let [entity-fields [:created :modified]]
+    (apply dissoc record entity-fields)))
```
ereteog: that list is short and you now have only one case, you should directly shorten it to: 
``` clojure
(apply dissoc record [:created :modified])
```

src/ctia/entity/tool.clj
```diff
@@ -1,5 +1,6 @@
 (ns ctia.entity.tool
   (:require [ctia.entity.tool.schemas :as ts]
+            [ctia.entity.tool.graphql-schemas :as tgs]
```
ereteog: That ns is not used

src/ctia/graphql/schemas.clj
```diff
@@ -160,3 +160,5 @@
 
 (defn execute [query operation-name variables context]
   (g/execute graphql query operation-name variables context))
+
+
```
ereteog: could you remove this trailing carriage return?

test/ctia/test_helpers/graphql.clj
```diff
@@ -1,9 +1,9 @@
 (ns ctia.test-helpers.graphql
-  (:require [clojure.test :refer [is testing]]
+  (:require [clojure
+             [string :as str]
```
ereteog: I know that it was already here, but `str` is a clojure function, would you mind to rename that alias into `string`, it is used only once in that ns: https://github.com/threatgrid/ctia/blob/631468b35718f24ec2521c0ab8ad35cb506392f5/test/ctia/test_helpers/graphql.clj#L62

src/ctia/entity/tool.clj
gbuisson: unfortunately it is implicitly used, loading the ns loads the GraphQL type, otherwise you get the default one from CTIM which doesn't include the custom fields.

project.clj
```diff
@@ -164,6 +164,7 @@
   :profiles {:dev {:dependencies [[cheshire ~cheshire-version]
                                   [org.clojure/test.check "0.9.0"]
                                   [com.gfredericks/test.chuck "0.2.8"]
+                                  [clj-http-fake "1.0.3"]
```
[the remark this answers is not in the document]
gbuisson: Yes, if you don't include it it's not loaded when you jack-in and then you cannot launch the tests in cider.

hshahman:
Verified in rel-v1.39

PR #863, opened 2019-12-04
using CTIM version 1.0.14

jyoverma:
Using CTIM version 1.0.14

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: CTIM update with a new indicator type
```

### DIFF
--- a/src/ctia/entity/vulnerability/mapping.clj
+++ b/src/ctia/entity/vulnerability/mapping.clj
@@ -67,7 +67,7 @@
       :temporal_score em/float-type
       :environmental_score em/float-type
       :remediation_level em/token
-      :temporal_severity em/float-type
+      :temporal_severity em/token
       :integrity_requirement em/token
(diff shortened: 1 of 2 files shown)


src/ctia/entity/vulnerability/mapping.clj
```diff
@@ -67,7 +67,7 @@
       :temporal_score em/float-type
       :environmental_score em/float-type
       :remediation_level em/token
-      :temporal_severity em/float-type
+      :temporal_severity em/token
```
yogsototh: Wouldn't that force a complete migration?

msprunck: In theory yes, but our instance does not use this field. Its definition was wrong.

I checked with this ElasticSearch query:

```json
{
    "query": {
        "exists": {
            "field": "impact.cvss_v3.temporal_severity"
        }
    }
}
```

PR #866, opened 2019-12-11
Bundle: filter relationships by node types

ereteog:
> closes #threatgrid/iroh/issues/3051

This PR enables to filter relationships by nodes type (based on `source_ref` and `target_ref` fields) in bundle export.
That version is based on regex on source_type and target_type uris, since adding these instances types in the documents would require a migration of relationships index.

You can now filter bundles with source and target types.

<a name="qa">[§](#qa)</a> QA
============================

1. import that bundle [bundle1.json.zip](https://github.com/threatgrid/ctia/files/3959819/bundle1.json.zip)
2. Note the id of the sighting id created with original id `transient:3`, and the id of the indicator.
3. Call bundle export with the id of the sighting id retrieved in step 2, and target_type=incident, it should retrieve the corresponding sighting, and the incident, but not the indicator.
4. Call bundle export with the id of sighting id retrieved in step 2, and target_type=indicator, it should retrieve the corresponding sighting, the inserted indicator, but not the incident.
5. Call bundle export with the id of indicator id retrieved in step 2, and source_type=sighting, it should retrieve that indicator, 2 inserted sightings 
6. Call bundle export with the id of indicator id retrieved in step 2, and source_type=sighting, it should retrieve that indicator and 2 inserted sightings but not the incident.

```
intern: Bundle Export route now proposes to filter relationships by source and target types.
```

### DIFF
--- a/src/ctia/bundle/core.clj
+++ b/src/ctia/bundle/core.clj
@@ -12,13 +12,12 @@
     [store :refer [list-fn
                    read-fn
                    read-store]]]
-   [ctia.lib.collection :refer [fmap]]
+   [ctia.lib.collection :as coll :refer [fmap]]
    [ctia.bulk.core :as bulk]
    [ctia.bundle.schemas
     :refer
     [BundleImportData BundleImportResult EntityImportData]]
    [ctia.domain.entities :as ent :refer [with-long-id]]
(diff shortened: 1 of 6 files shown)


hshahman:
Verified in rel-v1.40

PR #869, opened 2020-01-21
Feed: remove feed_view_url_csv and update feed_view_url according to the output

gbuisson:
> Close https://github.com/threatgrid/iroh/issues/3116

This PR removed the feed_view_url_csv field from the Feed entity,
instead the feed_view_url field is populated according to the Feed output

<a name="qa">[§](#qa)</a> QA
============================

1.  Create a Feed setting its output to `judgements`
2. the Response should contain the non CSV route as `feed_view_url`

1.  Create a Feed setting its output to `observables`
2. the Response should contain the CSV route as `feed_view_url`

```
public: Feed routes feed_view_url_csv removed
```


hshahman:
Verified in rel-v1.41

PR #893, opened 2020-03-10
New investigation fields

noprompt:
This is a work in progress. The notes here will evolve as needed.

The goal of this effort/patch is to ultimately eliminate the use of the double encoded investigation `"actions"` field by 

1. defining investigation fields,
2. updating ES mappings as needed, and
3. migrating existing `"actions"` field data to the new investigation fields.

# Related

https://github.com/threatgrid/iroh-engine/issues/202 (Origin story)
https://github.com/threatgrid/iroh-engine/issues/255 (Prescription)
https://github.com/threatgrid/ctim/pull/282 (Update to CTIM model)

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
```

### DIFF
--- a/src/ctia/entity/investigation.clj
+++ b/src/ctia/entity/investigation.clj
@@ -21,18 +21,10 @@
    [ctia.schemas.sorting :as sorting]))
 
 (def snapshot-action-fields-mapping
-  {:object_ids {:type "text"
-                :analyzer "text_analyzer"
-                :search_quote_analyzer "text_analyzer"
-                :search_analyzer "search_analyzer"
-                :include_in_all false}
+  {:object_ids em/token
    :targets {:type "nested"
              :include_in_all false}
-   :investigated_observables {:type "text"
-                              :analyzer "text_analyzer"
-                              :search_quote_analyzer "text_analyzer"
-                              :search_analyzer "search_analyzer"
-                              :include_in_all false}})
+   :investigated_observables em/text})
 
 (def investigation-mapping
   {"investigation"
--- a/src/ctia/entity/investigation.clj
+++ b/src/ctia/entity/investigation.clj
@@ -27,21 +27,24 @@
             [schema.core :as s]
             [ctia.schemas.graphql.ownership :as go]))
 
+(s/defschema InvestigationActionData
+  {(s/required-key :object_ids) [s/Str]
+   ;; The value of this field should look like "type:value".
+   (s/required-key :investigated_observables) [s/Str]
+   (s/required-key :targets) [(f-schema/->schema IdentitySpecification)]})
+
 (s/defschema Investigation
   (st/merge (f-schema/->schema inv/Investigation)
             CTIAEntity
-            {(s/required-key :actions) (s/either s/Str {s/Keyword s/Any})
-             (s/required-key :object_ids) [s/Str]
-             ;; The value of this field should look like "type:value".
-             (s/required-key :investigated_observables) [s/Str]
-             (s/required-key :targets) [(f-schema/->schema IdentitySpecification)]
-             s/Keyword s/Any}))
+            InvestigationActionData
+            {s/Keyword s/Any}))
 
 (f-spec/->spec inv/Investigation "investigation")
 
 (s/defschema PartialInvestigation
   (st/merge (f-schema/->schema (fu/optionalize-all inv/Investigation))
             CTIAEntity
+            InvestigationActionData
             {s/Keyword s/Any}))
 
 (s/defschema PartialInvestigationList
@@ -50,6 +53,7 @@
 (s/defschema NewInvestigation
   (st/merge (f-schema/->schema inv/NewInvestigation)
             CTIAEntity
+            InvestigationActionData
             {s/Keyword s/Any}))
 
 (f-spec/->spec inv/NewInvestigation "new-investigation")
@@ -105,7 +109,10 @@
                          :type
                          :search-txt
                          :short_description
-                         :created_at])))
+                         :created_at
+                         :object_ids
+                         :investigated_observables
+                         :targets])))
 
 (s/defschema InvestigationFieldsParam
   {(s/optional-key :fields) [investigation-select-fields]})
--- a/src/ctia/entity/investigation.clj
+++ b/src/ctia/entity/investigation.clj
@@ -17,6 +17,7 @@
             [ctia.stores.es
              [mapping :as em]
              [store :refer [def-es-store]]]
+            [ctim.schemas.common :refer [IdentitySpecification]]
             [ctim.schemas.investigation :as inv]
             [flanders
              [schema :as f-schema]
@@ -29,7 +30,12 @@
 (s/defschema Investigation
   (st/merge (f-schema/->schema inv/Investigation)
             CTIAEntity
-            {s/Keyword s/Any}))
+            {(s/required-key :actions) (s/either s/Str {s/Keyword s/Any})
+             (s/required-key :object_ids) [s/Str]
+             ;; "type:value"
+             (s/required-key :investigated_observables [s/Str])
+             (s/required-key :targets [(f-schema/->schema IdentitySpecification)])
+             s/Keyword s/Any}))
 
 (f-spec/->spec inv/Investigation "investigation")
 
@@ -64,7 +70,10 @@
                 :include_in_all false}
    :targets {:type "nested"
              :include_in_all false}
-   :investigated_observables {:type "nested"
+   :investigated_observables {:type "text"
+                              :analyzer "text_analyzer"
+                              :search_quote_analyzer "text_analyzer"
+                              :search_analyzer "search_analyzer"
                               :include_in_all false}})
 
 (def investigation-mapping
--- a/src/ctia/entity/investigation.clj
+++ b/src/ctia/entity/investigation.clj
@@ -1,70 +1,24 @@
 (ns ctia.entity.investigation
-  (:require [ctia.domain.entities :refer [default-realize-fn]]
-            [ctia.store :refer :all]
-            [ctia.http.routes
-             [common :refer [BaseEntityFilterParams PagingParams SourcableEntityFilterParams]]
-             [crud :refer [entity-crud-routes]]]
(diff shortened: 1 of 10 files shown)


noprompt:
Once it is decided where the schema changes belong, I will back out, if needed, the unnecessary changes.

noprompt:
There is another commit here that isn't showing up which I think may be due to the temporary outage today.

noprompt:
@gbuisson [This](https://github.com/threatgrid/ctia/compare/new-investigation-fields) view has all the commits.

noprompt:
@gbuisson Is there anything else that needs to be done to wrap this up and get this merged?

src/ctia/entity/investigation.clj
```diff
@@ -1,60 +1,38 @@
...
+            NewInvestigation
+            StoredInvestigation
+            PartialInvestigationList
+            realize-investigation]]
+   [schema-tools.core :as st]
+   [schema.core :as s]
+   [ctia.schemas.sorting :as sorting]))
+
+(def snapshot-action-fields-mapping
+  {:object_ids {:type "text"
```
ereteog, line 24: I would use `token` here, unless you have special ids and need to search in sub tokens.

src/ctia/task/migration/migrations.clj
```diff
@@ -162,4 +242,6 @@
                  pluralize-target)
    :1.0.0 (comp (append-version "1.0.0")
                 (rename-observable-type "pki-serial" "pki_serial")
-                simplify-incident)})
+                simplify-incident)
+   :1.1.0 (comp (append-version "1.1.0")
+                migrate-action-data)})
```
ereteog, line 247: Could you test this migration, please?
You have only one chance with that. In case of failure, the best case is that it fails and does not migrate the documents and we do not restart on the new documents. The worst case is that we switch on the newly created index with bad documents, insert new investigation entities, discover the bug and have to write a new migration.

src/ctia/task/migration/migrations.clj
```diff
@@ -149,6 +150,85 @@
...
+                                              (get m2 :investigated_observables))
+              :targets (into (get m1 :targets)
+                             (get m2 :targets))})))
+
+(defmulti derive-action-data
+  (fn [action]
+    (get action :type))
+  :default ::default)
+
+(defmethod derive-action-data ::default [_]
```
ereteog, line 191: "Never use `defmethod`"
https://github.com/threatgrid/ctia/blob/master/CONTRIBUTING.org#guidelines
The underlying idea is that it's hard to follow and debug, it can be gracefully replaced by a `case` here that redirect to proper function.

src/ctia/entity/investigation.clj
```diff
@@ -1,60 +1,38 @@
...
+                :analyzer "text_analyzer"
+                :search_quote_analyzer "text_analyzer"
+                :search_analyzer "search_analyzer"
+                :include_in_all false}
+   :targets {:type "nested"
+             :include_in_all false}
```
ereteog, line 30: You should add a more specific mapping. If you do not do so, some fields mapping will be dynamically inferred and probably with a bad type (for instance, "2020-05-01" could be inferred as a string instead of a date).
https://github.com/threatgrid/ctia/blob/master/src/ctia/stores/es/mapping.clj#L227

src/ctia/entity/investigation/examples.clj
```diff
@@ -0,0 +1,43 @@
...
+   :external_references [{:source_name "source"
+                          :external_id "T1067"
+                          :url "https://ex.tld/wiki/T1067"
+                          :hashes ["#section1"]
+                          :description "Description text"}]
+   :timestamp #inst "2017-10-23T19:25:27.278-00:00"
+   :language "language"
+   :object_ids []
+   :investigated_observables []
+   :targets []
```
ereteog, line 22: Could you example of values here?

src/ctia/entity/investigation.clj
```diff
@@ -1,60 +1,38 @@
...
+   :investigated_observables {:type "text"
+                              :analyzer "text_analyzer"
+                              :search_quote_analyzer "text_analyzer"
+                              :search_analyzer "search_analyzer"
+                              :include_in_all false}})
```
ereteog, line 35: We have specific mapping for observable, you should use this one:
https://github.com/threatgrid/ctia/blob/master/src/ctia/stores/es/mapping.clj#L130

src/ctia/entity/investigation.clj
noprompt: Do you mean `token` instead of `text`?

ereteog: yes

src/ctia/task/migration/migrations.clj
noprompt: Absolutely! I also prefer `case` to `defmethod` exactly for that reason. 😄

src/ctia/entity/investigation.clj
noprompt: > You should add a more specific mapping.

I'll see what I can do but IIRC this is a bit of a junk drawer.

src/ctia/entity/investigation.clj
noprompt: This field is not for full observables; its for a sequence of strings of the form `"type:value"`. Updating the examples will make that clear. The reason why its a sequence of strings has been discussed [here](https://github.com/threatgrid/ctim/pull/282#pullrequestreview-369379556).

ereteog: Ok, I remember, thus you should keep it as a text.
Please, directly use [this variable ](https://github.com/threatgrid/ctia/blob/master/src/ctia/stores/es/mapping.clj#L33) in for all your text mapping (there is other to replace in this PR).
It is the same value, but, in addition of avoiding code duplication, if we need to change its behavior it will be propagated to all text values.

src/ctia/entity/investigation.clj
ereteog: Well, it seems that it is a [pretty well defined](https://github.com/threatgrid/ctia/pull/893/files/63b2d7a26805d48429f3bb446e4f62f073a6e744#diff-8613a4405deeeb909529d9ba89de2591R21) junk drawer :-)
https://github.com/threatgrid/ctim/blob/master/src/ctim/schemas/common.cljc#L573

noprompt: Ha! Sure. But I don't know what to do with this information... what do you want/think should to go here?

src/ctia/task/migration/migrations.clj
noprompt: I added tests for the migration and the primary unit of work `derive-action-data`.

src/ctia/entity/investigation.clj
noprompt: Ok. Done. I swapped in `em/text` and `em/token`.

src/ctia/entity/investigation.clj
ereteog: Apparently it's the same schema than [sighting targets](https://github.com/threatgrid/ctim/blob/master/src/ctim/schemas/sighting.cljc#L80).
Thus I would simply use the [sighting target mapping](https://github.com/threatgrid/ctia/blob/master/src/ctia/stores/es/mapping.clj#L227)
``` clojure
:targets em/sighting-target
```

noprompt:
@ereteog I've updated all the spots you requested; thanks for all your help!

src/ctia/task/migration/migrations.clj
```diff
@@ -149,6 +150,91 @@
...
+
+(defn object-id [doc]
+  (let [id (get doc :id)]
+    (when (not (transient-id? id))
+      id)))
+
+(defn read-actions [investigation]
+  (let [actions (get investigation :actions)]
+    (if (string? actions)
+      (json/parse-string actions true)
```
ereteog, line 167: I would wrap it in someway just in case it contains garbage values.

src/ctia/entity/investigation/examples.clj
```diff
@@ -0,0 +1,46 @@
...
+   :source "a source"
+   :source_uri "http://example.com/somewhere-else"
+   :tlp "green"})
+
+(def investigation-minimal
+  {:id "http://ex.tld/ctia/investigation/investigation-2805d697-66b3-4e14-9b32-179e7a72eab6"
+   :source "a source"
+   :object_ids []
+   :investigated_observables []
+   :targets []
```
ereteog, line 38: Since `object_ids`, `investigated_observables` and `targets` are optional, the minimal example should not contain these fields.

src/ctia/task/migration/migrations.clj
noprompt: Got it.

noprompt:
> you only add new optional fields, isn't it?

Yes.

src/ctia/task/migration/migrations.clj
```diff
@@ -162,4 +254,6 @@
                  pluralize-target)
    :1.0.0 (comp (append-version "1.0.0")
                 (rename-observable-type "pki-serial" "pki_serial")
-                simplify-incident)})
+                simplify-incident)
+   :1.1.0 (comp (append-version "1.1.0")
```
ereteog, line 258: I was redacting the tenzin issue, and while filling the migration parameter I remembered to modify this
```suggestion
   :investigation-actions (comp (append-version "1.1.0")
```

src/ctia/entity/investigation/examples.clj
noprompt: Got it!

noprompt:
@ereteog I think thats all of it. Updated with current `master`.

noprompt:
@ereteog Everything look good?

ereteog:
@noprompt to quickly summarize our conversation:
We successfully tried it locally, you can merge it whenever it suits you. We will then migrate the old data whenever you need it. 
We could launch the migration on INT for your development requirements, the migration task is meant for being restarted where it stopped.
Thx :-)

noprompt:
@ereteog Perhaps we should being the migration at the beginning of next week?

PR #896, opened 2020-04-01
E3258: metric routes: topn, cardinality, histogram

ereteog:
> related https://github.com/threatgrid/iroh/issues/3258

This PR implements 3 metric endpoints for incident, sighting and judgement: `topn`, `cardinality` and`histogram`.
Each of these routes contains an `aggregate-on` parameter which accepted values are specific to each entity type. Then each metric type has it's own specific paremeters.
This PR also enables us to filter both search and metric requests with a date range using new `from` and/or `to` parameters. 
This PR extends the `IQueryStringSearchableStore` with an `aggregate` method, which is implemented in the ES store using [terms](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/search-aggregations-bucket-terms-aggregation.html), [cardinality](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/search-aggregations-metrics-cardinality-aggregation.html) and [date histogram](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/search-aggregations-bucket-datehistogram-aggregation.html) aggregations in Elasticsearch. Each of these aggregations is performed against the results of a search query which is identical to `search` route behavior.
The `query-string-search` method is slightly refactored with to add a `date-range` option. In order to avoid inflating the number of search parameters, all search parameters are now embedded in a `search-query` map with following schema:
``` clojure
(s/defschema SearchQuery
  (st/optional-keys
   {:query-string s/Str
    :filter-map {s/Keyword s/Any}
    :date-range DateRange}))
```
There is many modifications that are only to enable `:query` to be optional.
The main modifications to focus on are the insertion of aggregation and from/to filters in following files: ` src/ctia/http/routes/crud.clj`, `src/ctia/http/routes/common.clj ` (search query build), `src/ctia/stores/es/crud.clj `. In test the whole logic to test metric routes is in this helper ns `test/ctia/test_helpers/aggregate.clj`, that generate examples of entities and call the 3 types of metric. The unit tests are performed at the lowest level. The route tests are mostly meant to check that parameters are properly passed to `handle-aggregate` in the ES store. `handle-aggregate` has its own tests that are responsible for checking the different intended behaviors.

In CTIA, 3 metric endpoints are available for incident, sighting and judgement: `topn`, `histogram`, `cardinality`.

<a name="qa">[§](#qa)</a> QA
============================

## Metrics
1. Create an incident and note its timestamp to use it in next requests
2. Search incidents since the previous timestamp by replacing `{timestamp`} in that request:
`GET /ctia/incident/search?sort_by=timestamp&query=*from={timestamp}
Check that the result contains created entity.
3. Search incidents before the previous timestamp by replacing `{timestamp`} in that request:
`GET /ctia/incident/search?sort_by=timestamp&sort_order=desc&query=*to={timestamp}
Check that the result does not contain created entity (it should be empty if you can ensure that no incident have been created since step 1).
4. Search incidents since April 1 2020 and previous timestamp:
`GET /ctia/incident/search?sort_by=timestamp&sort_order=desc&query=*from="2020-04-01"&to={timestamp}
Check that the result does not contain created entity

## Top N and Cardinality metrics

1. Log in with a user whose org has incident created in previous weeks.
2. create 20 incidents: 8 with status Open,  6 with status `Close`, 4 with status `New`, 2 with status `Stalled`.
4. note the most recent `timestamp` value that will be referenced as `{timestamp}` in next requests and has to be replaced with this value.
3. Ensure that the environment is isolated during that test.

###  Top N Metric

1. request top `status` values in incidents since `{timestamp}` 
`GET /ctia/incident/metric/topn/aggregate-on=status&query=*&from={timestamp}`
it should return:
``` javascript
[
  {"key": "open", "value": 8},
  {"key": "close", "value": 6},
  {"key": "new", "value": 4},
  {"key": "stalled", "value": 2}
]
```
2. request top 3 `status` values in incident since `{timestamp}` 
`GET /ctia/incident/metric/topn/aggregate-on=status&query=*&from={timestamp}&limit=3`
``` javascript
[
  {"key": "open", "value": 8},
  {"key": "close", "value": 6},
  {"key": "new", "value": 4}
]
```

### Cardinality

1. Retrieve the number of distinct `status` values in incidents since `{timestamp}`:
`GET /ctia/incident/metric/cardinality/aggregate-on=status&query=*&from={timestamp}`
Check that it returns `4`.

### Histogram

1. Retrieve the histogram of incidents on `timestamp` field since the first of April with a `day` granularity (`timestamp` is the aggregated field it, do not replace it with the date):
`GET /ctia/incident/metric/histogram/aggregate-on=timestamp&granularity=day&query=*&from="2020-04-1"`
Check that it returns values for different days in April, and at least 20 for current day.
2. Perform the same request with `month` granularity since the begining of the year:
`GET /ctia/incident/metric/histogram/aggregate-on=timestamp&granularity=month&query=*&from="2020-01-1"`
Check that it returns values for each first month day of any incident was created in that month (you can check value using the `from`/`to` field in search and check `total-hits` response header.
 
This PR introduces metrics based on Elasticsearch aggregations: [terms](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/search-aggregations-bucket-terms-aggregation.html), [cardinality](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/search-aggregations-metrics-cardinality-aggregation.html) and [date histogram](https://www.elastic.co/guide/en/elasticsearch/reference/5.6/search-aggregations-bucket-datehistogram-aggregation.html).
I limited the date range aggregation to a maximum 1 year interval and manually selected fields that can be aggregated to limit potential troubles. However, we still need to check that these modifications have no impact on the cluster performances.

```
intern:  A new metric endpoint is available for incident, sighting and judgement at `/CTIA/incident/aggregate`, `/CTIA/judgement/aggregate`, `/CTIA/sighting/aggregate`.
```

### DIFF
--- a/src/ctia/entity/event/crud.clj
+++ b/src/ctia/entity/event/crud.clj
@@ -6,12 +6,6 @@
    [ctia.stores.es.crud :as crud]
    [clj-momo.lib.es.schemas :refer [ESConnState]]))
 
-'(defn attach-bulk-fields [index event]
-  (assoc event
-         :_type "event"
-         :_id (:id event)
-         :_index index))
-
 (def ^:private handle-create-fn
   (crud/handle-create :event Event))
 
--- a/src/ctia/entity/event/crud.clj
+++ b/src/ctia/entity/event/crud.clj
@@ -4,12 +4,9 @@
    [ctia.entity.event.schemas
     :refer [Event PartialEvent]]
    [ctia.stores.es.crud :as crud]
-   [clj-momo.lib.es
-    [document :as document]
-    [schemas :refer [ESConnState SliceProperties]]
-    [slice :refer [get-slice-props]]]))
+   [clj-momo.lib.es.schemas :refer [ESConnState]]))
 
-(defn attach-bulk-fields [index event]
+'(defn attach-bulk-fields [index event]
   (assoc event
          :_type "event"
          :_id (:id event)
@@ -27,19 +24,15 @@
                     {}
                     {}))
 
-(def ^:private handle-list-fn
+(def handle-list
   (crud/handle-find :event Event))
 
-(s/defn handle-list
-  [state :- ESConnState
-   filter-map :- crud/FilterSchema
-   ident
-   params]
-  (handle-list-fn state
-                  filter-map
-                  ident
-                  params))
+(def handle-read
+  (crud/handle-read :event PartialEvent))
 
 (def handle-event-query-string-search
   (crud/handle-query-string-search
-   "event" PartialEvent))
+   :event PartialEvent))
+
+(def handle-aggregate
+  (crud/handle-aggregate :event))
--- a/src/ctia/entity/judgement/schemas.clj
+++ b/src/ctia/entity/judgement/schemas.clj
@@ -84,6 +84,20 @@
   (concat judgement-fields
           ["valid_time.start_time,timestamp"]))
 
+(def judgement-enumerable-fields
+  [:disposition
+   :priority
+   :confidence
+   :severity
+   :observable.type
+   :observable.value])
+
+(def judgement-histogram-fields
+  [:timestamp
+   :valid_time.start_time
+   :valid_time.end_time])
+
+
 (def judgements-by-observable-sort-fields
   (map name (conj judgement-fields
                   "disposition:asc,valid_time.start_time:desc"
--- a/src/ctia/entity/judgement/schemas.clj
+++ b/src/ctia/entity/judgement/schemas.clj
@@ -97,7 +97,6 @@
    :valid_time.start_time
    :valid_time.end_time])
 
-
 (def judgements-by-observable-sort-fields
   (map name (conj judgement-fields
                   "disposition:asc,valid_time.start_time:desc"
--- a/src/ctia/schemas/search_agg.clj
+++ b/src/ctia/schemas/search_agg.clj
@@ -3,18 +3,23 @@
             [schema-tools.core :as st]))
 
 (s/defschema DateRange
+  "Date range query, includes lowerfrom and excludes to"
   {s/Keyword
    (st/optional-keys
     {:gte s/Inst
      :lt s/Inst})})
 
 (s/defschema SearchQuery
+  "components of a search query:
+   - query-string: free text search, with lucene syntax enabled"
   (st/optional-keys
    {:query-string s/Str
     :filter-map {s/Keyword s/Any}
     :date-range DateRange}))
 
-(s/defschema AggType (s/enum :histogram :topn :cardinality))
+(s/defschema AggType
+  "supported aggregation types"
+  (s/enum :histogram :topn :cardinality))
 
 (s/defschema AggCommonParams
   (st/merge
--- a/src/ctia/schemas/search_agg.clj
+++ b/src/ctia/schemas/search_agg.clj
@@ -1,0 +1,62 @@
+(ns ctia.schemas.search-agg
+  (:require [schema.core :as s]
+            [schema-tools.core :as st]))
+
+(s/defschema DateRange
+  {s/Keyword
+   (st/optional-keys
+    {:gte s/Inst
+     :lt s/Inst})})
+
+(s/defschema SearchQuery
+  (st/optional-keys
+   {:query-string s/Str
+    :filter-map {s/Keyword s/Any}
+    :date-range DateRange}))
+
+(s/defschema AggType (s/enum :histogram :topn :cardinality))
+
+(s/defschema AggCommonParams
+  (st/merge
+   {:aggregate-on s/Str}))
+
+(s/defschema Timezone
+  (let [positives (map #(format "+%02d:00" %) (range 12))
+        negatives (map #(format "-%02d:00" %) (range 1 12))]
+    (->> (concat positives negatives)
+         (apply s/enum))))
+
+(s/defschema HistogramParams
+  (st/merge
+   AggCommonParams
+   {:granularity (s/enum :day :week :month)
+    (s/optional-key :timezone) Timezone}))
+
+(s/defschema HistogramQuery
+  (st/merge
+   {:agg-type (s/eq :histogram)}
+   HistogramParams))
+
+(s/defschema TopnParams
+  (st/merge
+   AggCommonParams
+   (st/optional-keys
+    {:limit s/Int
+     :sort_order (s/enum :asc :desc)})))
+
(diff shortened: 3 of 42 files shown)


src/ctia/entity/judgement/schemas.clj
```diff
@@ -84,6 +84,20 @@
...
+   :confidence
+   :severity
+   :observable.type
+   :observable.value])
+
+(def judgement-histogram-fields
+  [:timestamp
+   :valid_time.start_time
+   :valid_time.end_time])
+
```
msprunck, line 99:
```suggestion
```

src/ctia/entity/event/crud.clj
```diff
@@ -4,12 +4,9 @@
...
     :refer [Event PartialEvent]]
    [ctia.stores.es.crud :as crud]
-   [clj-momo.lib.es
-    [document :as document]
-    [schemas :refer [ESConnState SliceProperties]]
-    [slice :refer [get-slice-props]]]))
+   [clj-momo.lib.es.schemas :refer [ESConnState]]))
 
-(defn attach-bulk-fields [index event]
+'(defn attach-bulk-fields [index event]
```
vemv, line 9: ?

src/ctia/schemas/search_agg.clj
```diff
@@ -0,0 +1,62 @@
...
+    (->> (concat positives negatives)
+         (apply s/enum))))
+
+(s/defschema HistogramParams
+  (st/merge
+   AggCommonParams
+   {:granularity (s/enum :day :week :month)
+    (s/optional-key :timezone) Timezone}))
+
+(s/defschema HistogramQuery
```
vemv, line 35: Maybe some of these defschemas could have a docstring. The names (`HistogramQuery`) seem pretty self-explanatory tbh, although sometimes some extra context can be added

src/ctia/stores/es/crud.clj
```diff
@@ -306,32 +311,107 @@ It returns the documents with full hits meta data including the real index in wh
...
+        date-range-query (when date-range
+                           {:range date-range})
+        filter-terms (-> (ensure-document-id-in-map filter-map)
+                         q/prepare-terms)]
+    {:bool
+     {:filter
+      (cond-> [(find-restriction-query-part ident)]
+        (seq filter-map) (into filter-terms)
+        (seq date-range) (conj date-range-query)
+        (not-empty query-string) (conj es-query-string))}}))
```
vemv, line 331: why `not-empty`?

generally I use `seq` as a predicate, and `not-empty` only for obtaining an object I intend to further use

The following seems to apply:

```
user=> (doc empty?)
-------------------------
clojure.core/empty?
([coll])
  Returns true if coll has no items - same as (not (seq coll)).
  Please use the idiom (seq x) rather than (not (empty? x))
```

test/ctia/stores/es/crud_test.clj
```diff
@@ -192,3 +199,238 @@
...
+                       {:refresh "true"})
+          query-string "title1"
+          date-range {:created {:gte timestamp-1
+                                :lt timestamp-2}}
+          filter-map {:confidence "High"}
+          search (fn [q params]
+                   (search-fn es-conn-state q ident params))]
+
+      (testing "Properly handle different search query options"
+        (is (= (count (concat high-t1-title1
```
vemv, line 332: Say that for whatever reason (refactorings...), both hands of the `=` eval'ed to 0.

Then we'd be comparing `(= 0 0)` which would pass but lose its original intent.

Accordingly I'd place `assert`s here and there.

test/ctia/stores/es/crud_test.clj
```diff
@@ -192,3 +199,238 @@
...
+      (testing "Properly handle search params"
+        (let [search-page-0 (search {:query-string query-string}
+                                    {:limit 2})
+              search-page-1 (search {:query-string query-string}
+                                    (get-in search-page-0 [:paging :next]))]
+          (is (= (count (concat high-t1-title1
+                                medium-t1-title1))
+                 (get-in search-page-0 [:paging :total-hits])))
+          (is (= 2 (count (:data search-page-0))))
+          (is (not= (:data search-page-0)
```
vemv, line 360: same observation about asserts (but now concerned with `nil`s instead of `0`s)

test/ctia/test_helpers/search.clj
```diff
@@ -152,95 +159,140 @@
...
-  (let [search-uri (format "ctia/%s/search" (name entity))]
-    (testing (format "GET %s" search-uri)
-      (let [response (search search-uri query)]
-        (is (= 200 (:status response)))
-        (is (= query (first (map query-field (:parsed-body response))))
-            "query term works"))
+  (testing "search term filter"
+    (let [response (search-text entity query)]
+      (is (= 200 (:status response)))
+      (is (= query (first (map query-field (:parsed-body response))))
```
vemv, line 197: Stronger assertion: `= [query] (map ...)`

test/ctia/test_helpers/search.clj
```diff
@@ -152,95 +159,140 @@
...
+      (is (empty? (:parsed-body response))
+          "filters should be applied, and should discriminate"))
+
+    (let [{:keys [status parsed-body]} (search-raw entity {:query query
+                                                           :tlp "green"})
+          matched-fields {:tlp "green"
+                          (keyword query-field) query}]
+      (is (= 200 status))
+      (is (<= 1 (count parsed-body)))
+      (is (every? #(= (select-keys % [(keyword query-field) :tlp])
```
vemv, line 217: This `every?` looks like it'd report not very expressively, if the tests failed.

i.e. it'd report `(not (every? ...)`

`doseq` would be free of the gotcha.

src/ctia/entity/event/crud.clj
ereteog: not done in purpose :-)
But it clearly reveals that I can complete this `ns` clean up and remove it :-)

test/ctia/test_helpers/search.clj
ereteog: I totally agree, but I am not sure that there is only one document here, it requires a deeper refacto after iterative test additions on this ns and crud-test ns. 
I discovered here that it was not properly tested (not the purpose of that PR) and thus added this test.
I will test all the results instead with a `doseq` as suggested in your other comment.
thx!

hshahman:
Verified in rel-v1.47

PR #897, opened 2020-04-24
Metrics add context 3508

ereteog:
> Related #https://github.com/threatgrid/iroh/issues/3508

This PR adds context metadata to metric results. The rest of the task will be done in another PR after the CTIA Investigate tile.

<a name="qa">[§](#qa)</a> QA
============================

Check that metric results contain metadata: `from`, `to`, `type` and `filters` if some have been provided in the query. see query examples in [aggregation doc](https://github.com/threatgrid/ctia/blob/master/doc/search-metrics.org#aggregating).

### DIFF
--- a/doc/search-metrics.org
+++ b/doc/search-metrics.org
@@ -82,12 +82,12 @@ GET /ctia/incident/topn?&aggregate-on=status&limit=3&from=2020-01-01
 #+END_SRC
 it will return for instance
 #+BEGIN_SRC javascript
-{"status": {"Closed": 12,
-            "Open": 6,
-            "New": 2},
+{"data": {"status": {"Closed": 12,
+                     "Open": 6,
+                     "New": 2}},
  "type": "topn",
- "from": "2020-01-01",
- "to": "2020-05-01"}
+ "filters": {"from": "2020-01-01",
+             "to": "2020-05-01"}}
 #+END_SRC
 
 The following request returns the top 20 observed values in judgements with malicious disposition, created between 2020/01/01 and 2020/04/01:
@@ -97,19 +97,19 @@ GET /ctia/judgement/topn?aggregate-on=observable.value&limit=20&from=2020-01-01&
 #+END_SRC
 
 #+BEGIN_SRC javascript
-{"observable":
-  {"value":
-    [
-     {"key": "abuse.ch", "value": 62},
-     {"key": "8.8.8.8", "value": 62},
-     {"key": "4964ab7e8d5959bb42b8ef78582082686dbe6565b010824785bc44595d8ebeee", "value": 38},
-     {"key": "https://www.badbabbad.net/worse.html", "value": 25},
-     ...
-    ]},
+{"data": {"observable":
+           {"value":
+             [
+              {"key": "abuse.ch", "value": 62},
+              {"key": "8.8.8.8", "value": 62},
+              {"key": "4964ab7e8d5959bb42b8ef78582082686dbe6565b010824785bc44595d8ebeee", "value": 38},
+              {"key": "https://www.badbabbad.net/worse.html", "value": 25},
+              ...
+             ]}},
  "type": "topn",
- "from": "2020-01-01",
- "to": "2020-04-01",
- "filters": {"disposition": 2}}
+ "filters": {"from": "2020-01-01",
+             "to": "2020-04-01",
+             "disposition": 2}}
 #+END_SRC
 
 ** Cardinality
@@ -125,10 +125,10 @@ GET /ctia/judgement/cardinality?aggregate-on=obsevable.value&from=2020-01-01
 
 It will return an integer that represent the estimated number of unique values
 #+BEGIN_SRC javascript
-{"observable": {"value": 5384},
+{"data": {"observable": {"value": 5384}},
  "type": "cardinality",
- "from": "2020-01-01",
- "to": "2020-05-01"}
+ "filters": {"from": "2020-01-01",
+             "to": "2020-05-01"}}
 #+END_SRC
 ** Histogram
    This aggregation returns an histogram of entity count for a given datetime field and a granularity, e.g. the number of incidents per day.
@@ -144,17 +144,17 @@ GET /ctia/incident/histogram?aggregate-on=timestamp&granularity=day&status=Open&
 #+END_SRC
 it will return for instance
 #+BEGIN_SRC javascript
-{"timestamp": [{"key": "2020-01-01", "value": 10},
-               {"key": "2020-01-02", "value": 0},
-               {"key": "2020-01-03", "value": 6},
-               {"key": "2020-01-04", "value": 2},
-               {"key": "2020-01-05", "value": 0},
-               {"key": "2020-01-06", "value": 1},
-               {"key": "2020-01-07", "value": 6},
-               {"key": "2020-01-08", "value": 4},
-               {"key": "2020-01-09", "value": 3}],
+{"data": {"timestamp": [{"key": "2020-01-01", "value": 10},
+                        {"key": "2020-01-02", "value": 0},
+                        {"key": "2020-01-03", "value": 6},
+                        {"key": "2020-01-04", "value": 2},
+                        {"key": "2020-01-05", "value": 0},
+                        {"key": "2020-01-06", "value": 1},
+                        {"key": "2020-01-07", "value": 6},
+                        {"key": "2020-01-08", "value": 4},
+                        {"key": "2020-01-09", "value": 3}]},
  "type": "histogram",
- "from": "2020-04-01",
- "to": "2020-05-01",
- "filters": {"Status": "Open"}} 
+ "filters": {"from": "2020-04-01",
+             "to": "2020-05-01",
+             "Status": "Open"}} 
 #+END_SRC
--- a/doc/search-metrics.org
+++ b/doc/search-metrics.org
@@ -82,10 +82,12 @@ GET /ctia/incident/topn?&aggregate-on=status&limit=3&from=2020-01-01
 #+END_SRC
 it will return for instance
 #+BEGIN_SRC javascript
-{
- "Closed": 12,
- "Open": 6,
- "New": 2
+{"status": {"Closed": 12,
+            "Open": 6,
+            "New": 2},
+ "type": "topn",
+ "from": "2020-01-01",
+ "to": "2020-05-01"
 }
 #+END_SRC
 
@@ -96,13 +98,19 @@ GET /ctia/judgement/topn?aggregate-on=observable.value&limit=20&from=2020-01-01&
 #+END_SRC
 
 #+BEGIN_SRC javascript
-[
- {"key": "abuse.ch", "value": 62},
- {"key": "8.8.8.8", "value": 62},
- {"key": "4964ab7e8d5959bb42b8ef78582082686dbe6565b010824785bc44595d8ebeee", "value": 38},
- {"key": "https://www.badbabbad.net/worse.html", "value": 25},
- ...
-]
+{"observable":
+  {"value":
+    [
+     {"key": "abuse.ch", "value": 62},
+     {"key": "8.8.8.8", "value": 62},
+     {"key": "4964ab7e8d5959bb42b8ef78582082686dbe6565b010824785bc44595d8ebeee", "value": 38},
+     {"key": "https://www.badbabbad.net/worse.html", "value": 25},
+     ...
+    ]},
+ "type": "topn",
+ "from": "2020-01-01",
+ "to": "2020-04-01",
+ "filters": {"disposition": 2}}
 #+END_SRC
 
 ** Cardinality
@@ -118,7 +126,10 @@ GET /ctia/judgement/cardinality?aggregate-on=obsevable.value&from=2020-01-01
 
 It will return an integer that represent the estimated number of unique values
 #+BEGIN_SRC javascript
-5391
+{"observable": {"value": 5384},
+ "type": "cardinality",
+ "from": "2020-01-01",
+ "to": "2020-05-01"}
 #+END_SRC
 ** Histogram
    This aggregation returns an histogram of entity count for a given datetime field and a granularity, e.g. the number of incidents per day.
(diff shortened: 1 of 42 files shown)


test/ctia/test_helpers/aggregate.clj
```diff
@@ -50,17 +50,29 @@
                     (vector? (first values)) flatten-list-values)]
     (map string/lower-case flattened)))
 
+(defn- check-from-to
+  [from to]
```
vemv, line 54: Suppose that both `from` and `to` are empty strings. Then this test would happily pass:

```clj
(<= (time/in-years
     (time/interval (tc/from-string "")
                    (tc/from-string "")))
    1) ;; => true
```

I tend to be wary of possibly empty strings in date helpers, as multiple Clojure date libs are known to be a bit permissive there.

More generally, in test helpers (or around assertions) I try to add extra checks for preventing false positives, negatives etc.

test/ctia/test_helpers/aggregate.clj
```diff
@@ -50,17 +50,29 @@
...
   [examples entity field]
   (testing (format "cardinality %s %s" entity field)
     (let [expected (-> (normalized-values examples field)
                        set
-                       count)]
-      (is (= expected (cardinality entity
-                                   {:query "*"
-                                    :from "2020-01-01"}
-                                   {:aggregate-on (name field)}))))))
+                       count)
```
vemv, line 67: I'd assert that `count` is `pos?` (since probably that's always the case)

---

Similar suggestions for the rest of this ns

test/ctia/http/routes/common_test.clj
```diff
@@ -2,20 +2,24 @@
...
-      (is (= (tc/from-string "2019-12-31")
-             (sut/coerce-from (tc/from-string "2019-12-30") nil)))
-      (is (= (tc/from-string "2019-06-01")
-             (sut/coerce-from (tc/from-string "2019-06-1") to)))
-      (is (= from
-             (sut/coerce-from from nil)))
-      (is (= from
-             (sut/coerce-from from to))))))
+      (is (= {:gte (tc/from-string "2019-12-31")
+              :lt (sut/now)}
```
vemv, line 12: Can't comparing two `now`s (expected/actual) result in flaky tests?

```clj
(= (java.util.Date.)
   (do
     (Thread/sleep 5)
     (java.util.Date.))) ;; => false
```

ereteog: There is a `with-redefs` of `now` that wraps that execution to ensure that this function always returns the same date.

vemv: 👍didn't realise!

test/ctia/test_helpers/aggregate.clj
```diff
@@ -126,6 +132,7 @@
           res-days (map #(to-granularity-first-day granularity %)
                         date-values)
           expected (make-histogram-res res-days)
+          _ (every? #(-> % first :value) expected)
```
vemv, line 135: missing `(assert` I believe

ereteog: Indeed :-)! I guess that time is running out to wake up today ;-)

doc/search-metrics.org
```diff
@@ -82,11 +82,12 @@ GET /ctia/incident/topn?&aggregate-on=status&limit=3&from=2020-01-01
...
- "Closed": 12,
- "Open": 6,
- "New": 2
-}
+{"status": {"Closed": 12,
+            "Open": 6,
+            "New": 2},
+ "type": "topn",
+ "from": "2020-01-01",
+ "to": "2020-05-01"}
```
msprunck, line 90: You merge the result with the query params. Not easy to extract the result from the whole response. Wouldn't it be better to separate the query params in a submap with all metas of the request.

ereteog: There is only `type`, `from` and `to` that are at the route, like the aggregated field. Then all the optional search filters are put in a `filters` field:
https://github.com/threatgrid/ctia/pull/897/files#diff-e59b44685eb63cb4899cbef38c1d7d94R159

msprunck: Why `from` and `to` are not in the filters field?

ereteog: well, even if they act as a filter to they are also some kind of metadata for the computed metric. Unlike other filters they are somehow dynamic while filters are exactly the user input.
`to` is optional and set to the query date.
`from` is mandatory but coerced to `to` - 1 year if the interval exceed 1 year.
I prefer this version but I can put it in `filters` if you consider that it would be more consistent.

yogsototh: I agree that I would have naturally but all the metas info in a submap and also put the result of the aggregations inside a `:data` because it appear the field is not always `"status"` but could be anything. What if I want to aggregate a field named "from" or `type` (in observable I think) or `filters` or if we want to add more metas in the future? Isn't there a collision risk?

Wouldn't it be safer to envelop the result in something like this?

```
{:data {"status": ...}
 :metas {:from ... :to ... :type ... :filters ...}}
```

ereteog: Considering this last argument and after a sync with @yogsototh I propose the following schema:
```clojure
(s/defschema AggResult
  {:data {:keyword s/Any}
   :type (s/enum :topn :histogram :cardinality)
   :filters (st/open-schema
                {:from s/Inst
                 :to s/Inst}}
```

doc/search-metrics.org
```diff
@@ -96,13 +97,19 @@ GET /ctia/judgement/topn?aggregate-on=observable.value&limit=20&from=2020-01-01&
...
 #+BEGIN_SRC javascript
-[
- {"key": "abuse.ch", "value": 62},
- {"key": "8.8.8.8", "value": 62},
- {"key": "4964ab7e8d5959bb42b8ef78582082686dbe6565b010824785bc44595d8ebeee", "value": 38},
- {"key": "https://www.badbabbad.net/worse.html", "value": 25},
- ...
-]
+{"data": {"observable":
+           {"value":
```
yogsototh, line 101: I'm not sure why the level `value` is needed here.
aren't all `topn` results a list?
So I think you could expect that observale value would be directly the list.
to prevent a bit of redundancy.

doc/search-metrics.org
```diff
@@ -118,7 +125,10 @@ GET /ctia/judgement/cardinality?aggregate-on=obsevable.value&from=2020-01-01
 
 It will return an integer that represent the estimated number of unique values
 #+BEGIN_SRC javascript
-5391
+{"data": {"observable": {"value": 5384}},
```
yogsototh, line 128: same remark about value here. Unless perhaps you would like to add redundancy by
also providing a `type` along `value` here it would be `"type":"integer"` for example.
But I think you could get rid of `"value"` here.

doc/search-metrics.org
```diff
@@ -134,13 +144,17 @@ GET /ctia/incident/histogram?aggregate-on=timestamp&granularity=day&status=Open&
...
-[{"key": 2020-01-01", "value": 10},
- {"key": 2020-01-02", "value": 0},
- {"key": 2020-01-03", "value": 6},
- {"key": 2020-01-04", "value": 2},
- {"key": 2020-01-05", "value": 0},
- {"key": 2020-01-06", "value": 1},
- {"key": 2020-01-07", "value": 6},
- {"key": 2020-01-08", "value": 4},
- {"key": 2020-01-09", "value": 3}]
+{"data": {"timestamp": [{"key": "2020-01-01", "value": 10},
```
yogsototh, line 147: here there is not value and I think this is fine.

src/ctia/http/routes/crud.clj
```diff
@@ -229,45 +231,59 @@
                 :auth-identity identity
                 :identity-map identity-map
                 (GET "/histogram" []
-                     :return [{:key s/Str :value s/Int}]
+                     :return AggResult
```
yogsototh, line 234: As the name is presented to users in SwaggerUI, I think you could use a longer name here:
`AggregationResult`. You might perhaps provide even more schema insight by having a specific
`HistogramResult` or `Histogram` schema here.
My guess is that each kind of aggregation share a common pattern but have a few useful distinctions
that could be shared via the schema.

src/ctia/http/routes/crud.clj
```diff
@@ -229,45 +231,59 @@
...
                      :summary (format "Histogram for a %s field" capitalized)
                      :query [params histogram-q-params]
```
yogsototh, line 236: why is this a lowercase symbol here and not a Camelcase named symbol as we use for all schemas?

src/ctia/http/routes/crud.clj
```diff
@@ -229,45 +231,59 @@
...
-                          entity
-                          aggregate
-                          (search-query (keyword
-                                         (:aggregate-on params))
-                                        (st/select-schema params agg-search-schema)
-                                        coerce-from)
-                          (st/assoc (st/select-schema params HistogramParams)
-                                    :agg-type :histogram)
-                          identity-map)))
+                     (let [aggregate-on (keyword (:aggregate-on params))
```
yogsototh, line 237: I couldn't see the `histogram-q-params` but, my guess is that it might be quite open.
I would really take care about using `keyword` here.
Because `keyword` is quite fragile on strings containing special chars.
So if user can provide any string here, I would take extra precaution to prevent any form of attack.

for example `(keyword "abc def") => :abc def` or `(keyword "123abc") => :123acb` etc...

src/ctia/http/routes/crud.clj
```diff
@@ -229,45 +231,59 @@
...
+                            entity
+                            aggregate
+                            search-q
+                            agg-q
+                            identity-map)
+                           (format-agg-result :histogram aggregate-on search-q)
+                           ok)))
                 (GET "/topn" []
-                     :return [{:key s/Any :value s/Int}]
+                     :return AggResult
```
yogsototh, line 252: I think the same remark apply here.

src/ctia/schemas/search_agg.clj
```diff
@@ -65,3 +65,10 @@
   (st/open-schema
    {:agg-type AggType
     :aggregate-on s/Str}))
+
+(s/defschema AggResult
+  {:data {s/Keyword s/Any}
+   :type AggType
+   :filters (st/open-schema
+             {:from s/Inst
+              :to s/Inst})})
```
yogsototh, line 74: As I said in a previous comment, you might be able to take advantage of generating
a few more precise schemas for each kind of aggregation.

test/ctia/http/routes/common_test.clj
```diff
@@ -72,7 +76,85 @@
...
+      (is (= {:data {:observable {:type cardinality}}
+              :type :cardinality
+              :filters {:from from
+                        :to to
+                        :query-string "baddomain"
+                        :field1 "value1"
+                        :field2 "value2"}}
+             (sut/format-agg-result cardinality
+                                    :cardinality
+                                    "observable.type"
```
yogsototh, line 113: I think you should try to check what is occuring if using special char here.
in particular `*`, `/` and ` `.

test/ctia/http/routes/common_test.clj
```diff
@@ -72,7 +76,85 @@
...
+                                    {:date-range
+                                     {:timestamp {:gte from
+                                                  :lt to}}
+                                     :query-string "baddomain"
+                                     :filter-map {:field1 "value1"
+                                                  :field2 "value2"}})))
```
yogsototh, line 119: I think you should check what occurs if `query-string` contains special chars. 
and the same for `:field*`. I would try `:field1 "foo/bar"`, `"123abc"`, `"abc def"`, etc...

test/ctia/test_helpers/aggregate.clj
```diff
@@ -35,8 +35,8 @@
 
 (defn flatten-list-values
   [values]
-  (apply concat
-         (map set values)))
+  (->> (map set values)
+       (apply concat)))
```
yogsototh, line 39: Doesn't [`mapcat`](https://clojuredocs.org/clojure.core/mapcat) work here?

```clojure
(mapcat set values)
```

doc/search-metrics.org
yogsototh: as some of my remarks during my review, I think you should not use `:keyword` for the keys of `:data` because it prevent you to support search on field containing special chars. I would use `string` here.

doc/search-metrics.org
ereteog: :-)
`value` in this example is the nested field of the object [`observable`](https://github.com/threatgrid/ctim/blob/master/src/ctim/schemas/common.cljc#L321). In this example we queried a topn on `observable.value`. On `observable.type` the result will look like:
``` javascript 
{
 "data": {"observable": "type": [...]}
}
```

doc/search-metrics.org
ereteog: here we aggregate on `timestamp`

doc/search-metrics.org
ereteog: Sorry, my comment contains a typo:
```
(s/defschema AggResult
  {:data {s/Keyword s/Any}
   :type (s/enum :topn :histogram :cardinality)
   :filters (st/open-schema
                {:from s/Inst
                 :to s/Inst}}
```
However, I do not understand your comment, that's the schema of the response here.

test/ctia/http/routes/common_test.clj
ereteog: this value is the name of exsiting CTIM field, I could strengthen that function with such case, but I do not think that it would happen, would it?

src/ctia/http/routes/crud.clj
ereteog: `histogram-q-params` is [built-on](https://github.com/threatgrid/ctia/pull/896/files#diff-e934e438d94b4fd1a73d3a138adced8dR92) `search-q-params`, it keeps entity filters and adds histogram specific params.
`aggregate-on` possible values is defined per entity type, and should not contain weird values, since its fields or fields + subfields.
Here is the possible values for [incident](https://github.com/threatgrid/ctia/pull/896/files#diff-af4c6ba2f07524c30fea358b19b8beeeR177):
``` clojure
(def incident-histogram-fields
  [:timestamp
   :incident_time.opened
   :incident_time.discovered
   :incident_time.reported
   :incident_time.remediated
   :incident_time.closed
   :incident_time.rejected])
```

src/ctia/http/routes/crud.clj
ereteog: `histogram-q-params` is built at the beginning of the function like stated in my previous comments.
It's similar to `[search-q-params](https://github.com/threatgrid/ctia/blob/master/src/ctia/http/routes/crud.clj#L213)` that is passed to the function.

yogsototh: Doesn't it mess with swagger-UI?

yogsototh: I just meant, I think a `{SearchQueryParameters :search-q-params}` would do the trick for the macro.

ereteog: That's a query parameters here, it's not the same behavior than for the `body` parameter of a `POST`.
In swagger it produces:
Screenshot 2020-05-04 at 15 34 01
In the case of a body, it uses the Schema name but it's not dependant of the variable name:
`:body [new-entity new-schema {:description (format "a new %s" capitalized)}]`
Screenshot 2020-05-04 at 15 32 40

hshahman:
Verified in rel-v1.49

PR #898, opened 2020-05-05
Overhaul Update event field diffs

frenchy64:
Related https://github.com/threatgrid/iroh/issues/3551

`added` and `deleted` fields don't show up in update events. This is explicit in the code, but we believe this is an oversight. This PR allows these events and tests the behavior.

Also supports truncating of values in diffs based on depth and count.

## TODO
- [x] loosen Update schema to include added and deleted
- [x] add tests for changing/removing composite data structures
- [x] make diff printing limits configurable in ctia.properties (see properties.clj)

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: allow `added` and `deleted` fields in update change event
```

### DIFF
--- a/src/ctia/entity/event/obj_to_event.clj
+++ b/src/ctia/entity/event/obj_to_event.clj
@@ -116,28 +116,31 @@
          added-keys (set/difference new-keys old-keys)
          deleted-keys (set/difference old-keys new-keys)]
      (vec
-       (concat
-         (mapcat (fn [k]
-                   {:pre [(keyword? k)]}
-                   (let [[oldv newv] (map k [old new])]
-                     (when (not= oldv newv)
-                       [{:field k
-                         :action "modified"
-                         :change {:before (truncate oldv)
-                                  :after (truncate newv)}}])))
-                 same-keys)
-         (map (fn [k]
-                {:pre [(keyword? k)]}
-                {:field k
-                 :action "added"
-                 :change {:after (truncate (k new))}})
-              added-keys)
-         (map (fn [k]
-                {:pre [(keyword? k)]}
-                {:field k
-                 :action "deleted"
-                 :change {:before (truncate (k old))}})
-              deleted-keys))))))
+       ;; sort by :field then :action
+       (sort-by
+         (juxt :field :action)
+         (concat
+           (mapcat (fn [k]
+                     {:pre [(keyword? k)]}
+                     (let [[oldv newv] (map k [old new])]
+                       (when (not= oldv newv)
+                         [{:field k
+                           :action "modified"
+                           :change {:before (truncate oldv)
+                                    :after (truncate newv)}}])))
+                   same-keys)
+           (map (fn [k]
+                  {:pre [(keyword? k)]}
+                  {:field k
+                   :action "added"
+                   :change {:after (truncate (k new))}})
+                added-keys)
+           (map (fn [k]
+                  {:pre [(keyword? k)]}
+                  {:field k
+                   :action "deleted"
+                   :change {:before (truncate (k old))}})
+                deleted-keys)))))))
 
 (def default-max-count
   "Default maximum length of collections to preserve in Update diffs"
--- a/src/ctia/entity/event/obj_to_event.clj
+++ b/src/ctia/entity/event/obj_to_event.clj
@@ -57,6 +57,20 @@
                     (concat (map tru pre)
                             (when (seq suf)
                               [dots]))))
+       (set? v) (let [pre (take max-count v)
+                      ;; in case keys conflict after truncation,
+                      ;; remove all conflicting keys.
+                      ;; eg., if kvs is [:a :a :b]
+                      ;;      then truncate to [:b]
+                      ks (mapv tru pre)
+                      s (->> ks
+                             frequencies
+                             (filter (comp #{1} second))
+                             (map first)
+                             set)]
+                  (cond-> s
+                    (not= (count s) (count v)) (conj dots)))
+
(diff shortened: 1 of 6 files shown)


src/ctia/entity/event/obj_to_event.clj
```diff
@@ -22,10 +22,7 @@
   "Given the output of a `diff` between maps return a list
   of edit distance operation under the form of an Update map"
   [[diff-after diff-before _]]
-  (filter
-   (fn [{:keys [change]}]
-     (seq change))
-   (concat
+  (concat
```
yogsototh, line 25: maybe `mapcat`?

ereteog: this `concat` is applied to the results of 2 `map`.

frenchy64:
Re-requesting reviews because this PR turned into adding truncation support for diffs.

src/ctia/entity/event/obj_to_event.clj
```diff
@@ -41,27 +41,146 @@
...
+                      ;; eg., if kvs is [:a :a :b]
+                      ;;      then truncate to [:b]
+                      ks (mapv tru pre)
+                      s (->> ks
+                             frequencies
+                             (filter (comp #{1} second))
+                             (map first)
+                             set)]
+                  (cond-> s
+                    (not= (count s) (count v)) (conj dots)))
```
ereteog, line 72: I would truncate the values first, remove conflicting ones and then remove elements.
``` clojure
(let [s (->> (map tru pre)
             set
             (take max-count))]
  (cond-> s
      (not= (count s) (count v)) (conj (vec dots))))
```

ereteog: BTW, here we change the type of the value, it's not a big deal when we just look at it, but it could impact an agent that would reason on these events.

src/ctia/entity/event/obj_to_event.clj
```diff
@@ -41,27 +41,146 @@
...
+
+       (map? v) (let [pre (take max-count (keys v))
+                      ;; in case keys conflict after truncation,
+                      ;; remove all conflicting keys.
+                      ;; eg., if kvs is [[:a 1] [:a 2] [:b 3]]
+                      ;;      then truncate to [[:b 3]]
```
ereteog, line 78: why truncating keys?

src/ctia/entity/event/obj_to_event.clj
```diff
@@ -41,27 +41,146 @@
...
+                      ;; in case keys conflict after truncation,
+                      ;; remove all conflicting keys.
+                      ;; eg., if kvs is [:a :a :b]
+                      ;;      then truncate to [:b]
+                      ks (mapv tru pre)
+                      s (->> ks
+                             frequencies
+                             (filter (comp #{1} second))
+                             (map first)
+                             set)]
```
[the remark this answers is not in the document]
frenchy64: Preserving collection type has been tested pretty well. Do you have an example where the wrong type is returned from `truncate`?

src/ctia/entity/event/obj_to_event.clj
frenchy64: Hmm I think you're confusing this case with the `vector?` case. That's at the bottom.

src/ctia/entity/event/obj_to_event.clj
frenchy64: I'm treating each time we recur into a collection as 1 depth, so if the depth limit is 1 we truncate `{:a 1 :b 2 :c 2}` to `{... ...}` instead of `{:a ... :b ... :c ...}`.

frenchy64:
@ereteog thanks for looking! I'll move the conversation about motivations to https://github.com/threatgrid/iroh/issues/3551.

src/ctia/entity/event/obj_to_event.clj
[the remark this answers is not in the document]
ereteog: Your right the type preservation is ok, sorrry (and it's not ok in my code proposal below).
Considering this particular comment, it was to suggest to directly use `set` on the list of truncated element to ensure the unicity of elements.

src/ctia/entity/event/obj_to_event.clj
ereteog: Sorry, I think that this is equivalent to your code
``` clojure
(let [s (->> (take max-count)
                   (map tru pre)
                   set)]
  (cond-> s
      (not= (count s) (count v)) (conj dots)))
```

In any case, the result could be weird for a set since it's unordered:
``` clojure
user=> (conj #{1 2 3} 4)
#{1 4 3 2}
user=> (conj #{1 2 3} "...")
#{"..." 1 3 2}
user=> (conj #{"el1" "el2" "el3"} "...")
#{"..." "el3" "el1" "el2"}
```

src/ctia/entity/event/obj_to_event.clj
ereteog: Ok I have not understood it properly.

frenchy64:
Closing. This PR includes solutions for both https://github.com/threatgrid/iroh/issues/3551 and https://github.com/threatgrid/iroh/issues/3680, but they should be done separately. AFAICT, https://github.com/threatgrid/iroh/issues/3551 is blocked by https://github.com/threatgrid/iroh/issues/3680 ([comment](https://github.com/threatgrid/iroh/issues/3551#issuecomment-634083978)), so we should tackle https://github.com/threatgrid/iroh/issues/3680 first.

PR #899, opened 2020-05-06
E3508: count endpoint

ereteog:
> **Epic** #https://github.com/threatgrid/iroh/issues/3508

This PR adds a `/ctia/{entity-type}/search/count` endpoint to enable clients to easily count the number of entities that match a given search query. It is more straightforward than using the `total-hits` header of `search` route, and since it is based on the ES `_count` endpoint it is less resource consuming on ES side (it does not load documents from disk and only aggregate count results of answering shards).

<a name="qa">[§](#qa)</a> QA
============================
This PR adds a `/ctia/{entity-type}/search/count` endpoint that count the number of matched entities with the same behavior than search. You could thus reuse your search test and check that the corresponding `count` returns same value than the `total-hits` header of the corresponding `search` response.

<!-- UNCOMMENT THIS SECTION IF NEEDED

### DIFF
--- a/doc/search-metrics.org
+++ b/doc/search-metrics.org
@@ -93,9 +93,9 @@ GET /ctia/incident/topn?&aggregate-on=status&limit=3&from=2020-01-01
 #+END_SRC
 it will return for instance
 #+BEGIN_SRC javascript
-{"data": {"status": {"Closed": 12,
-                     "Open": 6,
-                     "New": 2}},
+{"data": {"status": [{"key": "Closed", "value": 12},
+                     {"key": "Open", "value": 6},
+                     {"key": "New", "value": 2}]},
  "type": "topn",
  "filters": {"from": "2020-01-01",
              "to": "2020-05-01"}}
--- a/doc/search-metrics.org
+++ b/doc/search-metrics.org
@@ -59,6 +59,17 @@ For instance you can search malicious judgments with high confidence matching "a
 GET /ctia/incident/search?query=abuse.ch&confidence=high&limit=10&sort_by=disposition&sort_order=desc
 #+END_SRC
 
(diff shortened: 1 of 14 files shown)


doc/search-metrics.org
```diff
@@ -59,6 +59,17 @@ For instance you can search malicious judgments with high confidence matching "a
 GET /ctia/incident/search?query=abuse.ch&confidence=high&limit=10&sort_by=disposition&sort_order=desc
 #+END_SRC
 
+Finally you can only query entity count with the same filtering parameters through the endpoint ~GET /ctia/{entity-type}/search/count~.
```
yogsototh, line 62: same remark as I do most of the time.
I think using GET parameters while better to respect HTTP protocol is
harder and less composable than using a full JSON body as a request.
This is why I think you should propose you should duplicate the handler with another
POST route that uses a JSON body instead of query parameters.

doc/search-metrics.org
```diff
@@ -59,6 +59,17 @@ For instance you can search malicious judgments with high confidence matching "a
...
+So reusing a previous search query example, one can count the indicators with "android" in the title, having a high confidence, created since 2020/01/01 with both text and term filtering:
+#+BEGIN_SRC
+GET /ctia/indicator/search/count?query=title:android&confidence=high&from=2020-01-01
```
yogsototh, line 65: is `:` an ES notation? Because the most usual notation is `.`.

doc/search-metrics.org
ereteog: I think it worth doing that but I just documented existing route.

doc/search-metrics.org
ereteog: it's a lucene notation inside the `query` text field, it will match entities with `title` field that matches `android`.

hshahman:
Verified in rel-v1.49

PR #900, opened 2020-05-15
CTIA metrics on remaining entities

ereteog:
> **Epic** #https://github.com/threatgrid/iroh/issues/3508
> Close #https://github.com/threatgrid/iroh/issues/3508

This PR enable metrics on other CTIA entities: Metrics were introduced for already searchable CTIM entities: Actor, AttackPattern, Campaign, Casebook, COA, IndentityAssertion, Indicator, Investigation, Malware, Relationship, Tool, Vulnerability and Weakness.
This PR also strenghtens generative testing on metric to avoid some random failures due too bad ES simulation.

<a name="qa">[§](#qa)</a> QA
============================

Metrics were introduced for already searchable CTIM entities: Actor, AttackPattern, Campaign, Casebook, COA, IndentityAssertion, Indicator, Investigation, Malware, Relationship, Tool, Vulnerability and Weakness.
This changes are only about activiting metric routes on specific fields. These routes work on the same code that were used in previous PR. Thus you should only check that metric routes return properly formatted results for each new entity and proposed fields for `aggregate-on` parameter.

```
intern: metric routes are now available for all entities.
```

### DIFF
--- a/test/ctia/entity/casebook_test.clj
+++ b/test/ctia/entity/casebook_test.clj
@@ -6,17 +6,18 @@
    [clojure
     [set :refer [subset?]]
     [test :refer [deftest is join-fixtures testing use-fixtures]]]
-   [ctia.entity.casebook :refer [casebook-fields]]
+   [ctia.entity.casebook :as sut]
    [ctia.test-helpers
     [access-control :refer [access-control-test]]
     [auth :refer [all-capabilities]]
     [core :as helpers :refer [post put patch post-entity-bulk]]
     [crud :refer [entity-crud-test]]
+    [aggregate :refer [test-metric-routes]]
     [fake-whoami-service :as whoami-helpers]
     [field-selection :refer [field-selection-tests]]
     [http :refer [doc-id->rel-url]]
     [pagination :refer [pagination-test]]
-    [store :refer [test-for-each-store]]]
+    [store :refer [test-for-each-store store-fixtures]]]
    [ctim.examples.casebooks
     :refer
     [new-casebook-maximal new-casebook-minimal]]
@@ -338,13 +339,13 @@
        (pagination-test
         "ctia/casebook/search?query=*"
         {"Authorization" "45c1f5e3f05d0"}
-        casebook-fields)
+        sut/casebook-fields)
 
        (field-selection-tests
         ["ctia/casebook/search?query=*"
          (doc-id->rel-url (first ids))]
         {"Authorization" "45c1f5e3f05d0"}
-        casebook-fields)))))
+        sut/casebook-fields)))))
 
 (deftest test-casebook-routes-access-control
   (test-for-each-store
@@ -353,3 +354,15 @@
                           new-casebook-minimal
                           true
                           true))))
+
+(deftest test-casebook-metric-routes
+  ((:es-store store-fixtures)
+   (fn []
+     (helpers/set-capabilities! "foouser" ["foogroup"] "user" all-capabilities)
+     (whoami-helpers/set-whoami-response "45c1f5e3f05d0" "foouser" "Administrators" "user")
+     (test-metric-routes {:entity :casebook
+                          :plural :casebooks
+                          :entity-minimal new-casebook-minimal
+                          :enumerable-fields sut/casebook-enumerable-fields
+                          :date-fields sut/casebook-histogram-fields
+                          :schema sut/NewCasebook}))))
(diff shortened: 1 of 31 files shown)


test/ctia/test_helpers/aggregate.clj
```diff
@@ -37,17 +37,45 @@
   [values]
   (mapcat set values))
 
+(defn l-get-in
```
vemv, line 40: Why is it called l-get-in?

test/ctia/test_helpers/aggregate.clj
```diff
@@ -37,17 +37,45 @@
...
+  "like get-in, but match keys in maps embedded in collections.
+  ~~~clojure
+  > (l-get-in {:a [{:b 2} {:b 3}]} [:a :b])
```
vemv, line 43: It'd sound desirable to replace most of the docstring with tests.

Tests-as-documentation can be a great thing to be emphasized, while large docstrings can be awkward in IDE widgets.

test/ctia/test_helpers/aggregate.clj
```diff
@@ -37,17 +37,45 @@
...
+  > (l-get-in {:a {:d 2 :c 3}} [:a :b])
+    nil
+  > (l-get-in {:a [{:d 2} {:b 3}]} [:a :b])
+    (3)
+  ~~~"
+  [m ks]
+  (reduce (fn [acc k]
+            (cond
+              (map? acc) (get acc k)
+              (coll? acc) (seq (keep #(get % k) acc))
```
vemv, line 62: what's the behavior is there's a `set` somewhere?

test/ctia/test_helpers/aggregate.clj
ereteog: that's already a helper for testing, it would be meta tests :-)?

test/ctia/test_helpers/aggregate.clj
ereteog: It will go to the `(coll? acc)` case
``` clojure
> (map? #{1 2 3})
false
> (coll? #{1 2 3})
true
```

test/ctia/test_helpers/aggregate.clj
ereteog: The `l` stands for list but I still do not have any relevant name. I have to confess that I cannot find a proper name to name that weird `get-in`.
I had to modify the behavior of `get-in` to simulate how ES matches elements inside vectors. That's in particular for that case:
``` clojure
 > (l-get-in {:a [{:b 2} {:b 3}]} [:a :b])
    (2 3)
;; while
 > (get-in {:a [{:b 2} {:b 3}]} [:a :b])
    nil
```
And if you ask in ES for the cardinality of `a.b` with only one document `{:a [{:b 2} {:b 3}]}`, it will return 2. So I introduced this custom `get-in` to extract these values in generated docs.
I added a detailed doc-string for this test helper to try to clarify it, but I apparentely failed :-).

test/ctia/test_helpers/aggregate.clj
ereteog: BTW I could add some assert?

vemv: I'm all for meta ;p

The following works (`:test` has an official meaning in Clojure):

```
(defn foo
  {:test (fn []
           (is ...))}
  []
  )
```

CIDER, Lein, etc all will run them as any other deftest.

These inlined tests seem a nice middle ground between having real coverage, and some accessible doc.

test/ctia/test_helpers/aggregate.clj
vemv: maybe `es-get-in`?

ereteog: Should I add this explanation as a comment in that test helper `ns` ?
To summarize that's pretty complicate to properly cover the different possible cases for metrics. Thus I opted to check some filters and have generative tests with different types of field values and metric on different type of nested fields.
I do not go to far on search filters, I mainly check that some are properly propagated to ES. Search filters (and thus mapping) are common to search routes and tested for search routes (in particular the different types of text mapping).

ereteog: > maybe es-get-in?

sold!

test/ctia/entity/casebook_test.clj
```diff
@@ -353,3 +354,15 @@
                           new-casebook-minimal
                           true
                           true))))
+
+(deftest test-casebook-metric-routes
+  ((:es-store store-fixtures)
+   (fn []
+     (helpers/set-capabilities! "foouser" ["foogroup"] "user" all-capabilities)
+     (whoami-helpers/set-whoami-response "45c1f5e3f05d0" "foouser" "Administrators" "user")
+     (test-metric-routes {:entity :casebook
```
gbuisson, line 363: the plural namings are already defined on the entity namespace, it would be better to reuse it to avoid duplication

https://github.com/threatgrid/ctia/blob/master/src/ctia/entity/actor.clj#L129

hshahman:
Verified in rel-v1.50

PR #906, opened 2020-05-26
Allow `added` and `deleted` fields in update change event (Take 2)

frenchy64:
Close https://github.com/threatgrid/iroh/issues/3551

Update limits are not handled and are deferred to a future fix of https://github.com/threatgrid/iroh/issues/3680.

- [x] test this actually solves https://github.com/threatgrid/iroh/issues/3551
- [x] move test to own ns
- [x] fix query in test

<a name="qa">[§](#qa)</a> QA
============================

This is a general fix for a specific problem, in particular for missing `:assignees` in update events. See attached issue https://github.com/threatgrid/iroh/issues/3551

There are three interesting cases to verify this fixes the attached issue.

1. Add an `assignees` field to an `Incident` which doesn't already have one. Verify the resulting event contains an `added` entry similar to this in its `:fields`:

```clojure
      {
        "field": "assignees",
        "action": "added",
        "change": {
          "after": <your-assignees>
        }
      }
```

2. Modify an existing `assignees` field in an `Incident. Verify the resulting event contains a `modified` entry similar to this in its `:fields`:

```clojure
      {
        "field": "assignees",
        "action": "modified",
        "change": {
          "before": <your-assignees-before-change>
          "after": <your-assignees-after-change>
        }
      }
```

3. Delete an existing `assignees` field in an `Incident. Verify the resulting event contains a `deleted` entry similar to this in its `:fields`:

```clojure
      {
        "field": "assignees",
        "action": "deleted",
        "change": {
          "before": <your-assignees-before-deletion>
        }
      }
```

```
intern: Allow `added` and `deleted` fields in update change event
```


src/ctia/entity/event/obj_to_event.clj
```diff
@@ -43,12 +46,49 @@
...
+        (juxt :field :action)
+        (concat
+          (mapcat (fn [k]
+                    {:pre [(keyword? k)]}
+                    (let [[oldv newv] (map k [old new])]
+                      (when (not= oldv newv)
+                        [{:field k
+                          :action "modified"
+                          :change {:before oldv
+                                   :after newv}}])))
```
ereteog, line 72: shouldn't we handle nested changes?

src/ctia/entity/event/schemas.clj
```diff
@@ -15,10 +15,24 @@
    DeleteEventType])
 
 (s/defschema Update
-  {:field s/Keyword
-   :action s/Str
-   :change {:before s/Any
-            :after s/Any}})
+  (s/conditional
```
ereteog, line 18: I think that it breaks swagger UI.
If it does break it, you should opt for a more open schema
```clojure
{:field s/Keyword
 :action (s/enum "modified" "added" "deleted")
 :change (st/optional-keys
                  {:before s/Any
                    :after s/Any})}
```

src/ctia/entity/event/obj_to_event.clj
frenchy64: This PR adds `added` and `removed` events. `modified` events are left unchanged.

src/ctia/entity/event/schemas.clj
frenchy64: Thanks, I forgot about Swagger. Reverted to a simpler schema.

ereteog: it was an easy catch, I broke Swagger similarly few weeks ago :-)

src/ctia/entity/event/obj_to_event.clj
frenchy64: https://github.com/threatgrid/iroh/issues/3551#issuecomment-635444332

frenchy64:
Thanks for the review @ereteog! I checked and we don't need to further enhance the diffs. I requested another review.

test/ctia/entity/event_test.clj
```diff
@@ -367,3 +367,92 @@
...
+                                   :parsed-body
+                                   (dissoc :assignees))
+                         :headers {"Authorization" "user1"})
+                    _ (is (= 200 (:status deleted-assignees-incident))
+                          deleted-assignees-incident)]
+
+                (testing ":fields are correctly set"
+                  (let [initial-id (get-in initial-incident [:parsed-body :id])
+                        q (uri/uri-encode
+                            (format "entity.id:\"%s\" OR entity.source_ref:\"%s\" OR entity.target_ref:\"%s\""
```
ereteog, line 440: You only need to search in `entity.id`.
`source_ref` and `target_ref` corresponds to the source and target node of a relationships, and even for a relationship we are only interested in the difference of the edited relationship.

test/ctia/events_test.clj
```diff
@@ -75,3 +75,53 @@
     (is (= 1 (-> (<!! output) :entity :data)))
     (is (= 2 (-> (<!! output) :entity :data)))
     (is (nil? (poll! output)))))
+
+(deftest to-update-event-test
```
ereteog, line 79: `ctia.events` implements an event channel, and `ctia.events-test` is dedicated to test it.
IMHO this test belongs to an ns dedicated to test `ctia.entity.event.obj-to-event` functions which is lacking.
Could you move this test please?

test/ctia/entity/event_test.clj
frenchy64: Thanks! Fixed.

test/ctia/events_test.clj
frenchy64: Great catch, moved to `ctia.entity.event.obj-to-event-test`.

NeelSaini:
Update on three scenarios to verify for missing :assignees in update events 

1) The 1st scenario works fine we now see the below entry when we add a new assignee to any incident 

```
  {
        "field": "assignees",
        "action": "added",
        "change": {
          "after": <your-assignees>
        }
      }
```

image


2) The 2nd scenario works fine too 

image

3) The 3rd scenario doesn't seem to be working as expected. We expected to see the below entry according to the description in ticket 

```
   {
        "field": "assignees",
        "action": "deleted",
        "change": {
          "before": <your-assignees-before-deletion>
        }
      }
```

But On deleting the assignee there is no entry for action deleted and also the UI status update under timeline says "Assignee changed"  instead of "Assignee deleted"

image

frenchy64:
@NeelSaini thanks for the report! This might be a miscommunication on my part. Step 3 was supposed to instruct to delete the `assignees` field altogether, not delete a single assignee. Perhaps this isn't possible from the frontend.

ie., step 3 was intended to verify that changing `{:assignees [uuid-1] ...}` to `{...}` triggered a delete event. I believe what you're seeing is a `{:assignees [uuid-1] ...}` => `{:assignees [] ...}` change, which is correctly a "changed" event.

Are there any fields available to you that you can delete altogether? If so, please verify it generates a "deleted" event. This is not special to `assignees`.

NeelSaini:
I have verified the 3rd scenario which is on deleting the assignees from backend generates a "deleted" event

image

I have one question though :

I created an incident and added assignees , Then I deleted the assignees from background via API and could see the "delete" event as row 6 in the below screen shot but I am wondering what is row 7 after deleting the assignees? I did not add any new assignee but I see in row 7 the field with action "modified" is added just wondering if that is something to take care of 

image

For your reference

Row 6: I deleted the assignees from the incident via API but left the assignees field there with [ ] value so that generates the delete event and with timeline status updated to "Assignee Changed" with [ ]

image

Row 7: After confirming the delete event on UI I went back to the API again and removed the assignee field along with [ ] completely that generates the modified event as shown below with timeline status updated to "Assignee Changed" with no [ ]

image

NOTE: Just wondering if that's something expected

Issue #905, opened 2020-05-26
Flaky metrics routes test


frenchy64:
[Observed](https://travis-ci.org/github/threatgrid/ctia/jobs/691444305#L796-L801) in JDK 11 only while working on unrelated test file:

```clojure
lein test :only ctia.entity.tool-test/test-tool-metric-routes
FAIL in (test-tool-metric-routes) (aggregate.clj:107)
cardinality :tool :labels
expected: (= expected (get-in (:data res) (parse-field field)))
  actual: (not (= 540 539))
```

Introduced with [this](https://github.com/threatgrid/ctia/commit/4c27df5259414f9c67f3f86a9a1ea4a2cf271dba) commit.

frenchy64:
Seems fixed.

PR #907, opened 2020-05-27
strenghten string generation in metric tests

ereteog:
> Related #https://github.com/threatgrid/ctia/issues/905

Sometime the metric tests fail on labels. I did not manage to reproduce this issue.
This PR strengthen the label generator and improve test failing messages to help understanding if it happens again. I do not close the issue since I cannot confirm that it solves the problem, but it will help ease to improve this label generation if necessary.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

### DIFF
--- a/test/ctia/test_helpers/aggregate.clj
+++ b/test/ctia/test_helpers/aggregate.clj
@@ -109,8 +109,7 @@
                                  {:aggregate-on (name field)})]
       (is (= expected
              (get-in (:data res) (parse-field field)))
-          (error-helper-msg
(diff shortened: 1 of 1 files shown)


frenchy64:
It might help to add more debugging information so we have a better idea of why it fails. I think right now we just get information about a mismatched `count`.

ereteog:
> It might help to add more debugging information so we have a better idea of why it fails. I think right now we just get information about a mismatched `count`.

I added a message with formatted values for cardinality and topn to help debugging it.

PR #908, opened 2020-06-11
Migration: prevent misconfiguration.

ereteog:
> close #threatgrid/iroh/issues/3766

This PR enables a migration to abort in case of misconfiguration:
- check if any of entity stores is configured with the same source and target indices.
- check the schema of the parameters.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

The migration now aborts in case of misconfiguration when the source and target store are identical. We should plan to run it on INT before the release.

### DIFF
--- a/src/ctia/task/migration/migrate_es_stores.clj
+++ b/src/ctia/task/migration/migrate_es_stores.clj
@@ -11,12 +11,14 @@
             [ctia.store :refer [stores]]
             [ctia.entity.entities :refer [entities]]
             [ctia.entity.sighting.schemas :refer [StoredSighting]]
+            [ctia.properties :refer [properties]]
             [ctia.stores.es
              [crud :refer [coerce-to-fn]]
              [store :refer [StoreMap]]]
             [ctia.task.migration.migrations :refer [available-migrations]]
             [ctia.task.migration.store :as mst])
-  (:import (java.util UUID)))
+  (:import [java.util UUID]
+           [java.lang AssertionError]))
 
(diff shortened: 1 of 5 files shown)


src/ctia/task/migration/migrate_es_stores.clj
```diff
@@ -269,25 +278,30 @@
...
 
-(defn run-migration
-  [migration-id prefix migrations store-keys batch-size buffer-size confirm? restart?]
-  (assert migration-id "Please provide an unique ID for this migration process")
+(s/defn ^:always-validate check-migration-params
+  [{:keys [prefix
+           store-keys]} :- MigrationParams]
+  (doseq [store-key store-keys]
+    (let [index (get-in @properties [:ctia :store :es store-key :indexname])]
+      (assert (not= (mst/prefixed-index index prefix)
```
vemv, line 286: throwing AssertionError by hand may be better here since assert depends on `*assert*` which ideally would be set to false in production (I think currently it's not).

i.e. given you `^:always-validate`, it seems well-aligned to not rely on `*assert*` being bound to its default val

src/ctia/task/migration/migrate_es_stores.clj
```diff
@@ -269,25 +278,30 @@
...
+                    index)
+              (format "the source and target indices are identical. The migration was misconfigured."))))
```
vemv, line 288: `are identical: %s [...]" index)`

PR #910, opened 2020-06-17
Fix investigation schema

ereteog:
This PR Fixes Investigation Schemas. CTIM flanders schemas were internalized in CTIA, but the former schema were not removed and thus actually used instead of the new schema. 
Since the former schema was open, it did not cause any trouble on added fields, but this was the cause of missing fields in grapqhl.

<a name="qa">[§](#qa)</a> QA
============================

The graphql endpoint should now enable to retrieve `object_ids`, `investigated_observables` and `targets`.

```
intern: fixed a bug in investigation graphql schema.
```

### DIFF
--- a/test/ctia/http/routes/graphql/investigation_test.clj
+++ b/test/ctia/http/routes/graphql/investigation_test.clj
@@ -1,0 +1,84 @@
+(ns ctia.http.routes.graphql.investigation-test
+  (:require [clj-momo.test-helpers.core :as mth]
+            [clojure.test :refer [deftest is join-fixtures testing use-fixtures]]
+            [ctia.entity.investigation :as inv]
+            [ctia.entity.investigation.examples :refer [investigation-maximal]]
+            [ctia.test-helpers.auth :refer [all-capabilities]]
+            [ctia.test-helpers.core :as helpers]
+            [ctia.test-helpers.fake-whoami-service :as whoami-helpers]
+            [ctia.test-helpers.fixtures :as fixt]
+            [ctia.test-helpers.graphql :as gh]
+            [ctia.test-helpers.store :refer [test-for-each-store]]))
+
+(use-fixtures :once (join-fixtures [mth/fixture-schema-validation
+                                    helpers/fixture-properties:clean
+                                    whoami-helpers/fixture-server]))
+
+(use-fixtures :each whoami-helpers/fixture-reset-state)
+
+(def ownership-data-fixture
+  {:owner "foouser"
+   :groups ["foogroup"]})
+
+(def investigation-1 (fixt/randomize investigation-maximal))
+(def investigation-2 (assoc
+                      (fixt/randomize investigation-maximal)
+                      :source "ngfw"))
+
+(defn select-graphql-fields
+  [investigation]
+  (select-keys investigation
+               [:id
+                :type
+                :schema_version
+                :title
+                :timestamp
+                :tlp
+                :source
+                :owner
+                :groups
+                :object_ids
+                :investigated_observables
+                :targets]))
+
+(deftest investigation-graphql-test
+  (test-for-each-store
+   (fn []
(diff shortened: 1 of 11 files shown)


ereteog:
OK, I introduce a test on investigation (there is no tests at all).
In any case, only the fields of the maximal that are in the schema will be exposed. That's related to the point that I mentioned yesterday, we cannot control anything with this open schema.

msprunck:
> OK, I introduce a test on investigation (there is no test at all).
> In any case, only the fields of the maximal entity example that are in the schema will be exposed. That's related to the point that I mentioned yesterday, we cannot control anything with this open schema.

There is a test here https://github.com/threatgrid/ctia/blob/master/test/ctia/http/routes/graphql_test.clj#L513

ereteog:
thank you!
I missed this one and only saw: https://github.com/threatgrid/ctia/tree/master/test/ctia/http/routes/graphql

test/ctia/http/routes/graphql/investigation_test.clj
```diff
@@ -0,0 +1,84 @@
...
+     (whoami-helpers/set-whoami-response "45c1f5e3f05d0"
+                                         "foouser"
+                                         "foogroup"
+                                         "user")
+     (let [inv1 (select-graphql-fields
+                 (gh/create-object "investigation" investigation-1))
+           inv2 (select-graphql-fields
+                 (gh/create-object "investigation" investigation-2))
+           graphql-queries (slurp "test/data/queries.graphql")]
+       (testing "investigations query"
```
msprunck, line 57: Could you also test the `investigation` Graphql query? This part has not been moved in this ns.

test/ctia/http/routes/graphql/investigation_test.clj
```diff
@@ -0,0 +1,84 @@
...
+                :schema_version
+                :title
+                :timestamp
+                :tlp
+                :source
+                :owner
+                :groups
+                :object_ids
+                :investigated_observables
+                :targets]))
```
msprunck, line 42: If you select the keys, the graphql test will not fail if new fields are added to the schema or the example. I would dissoc fields that will never be returned by the graphql query.

test/data/queries.graphql
```diff
@@ -275,6 +286,9 @@ fragment investigationFields on Investigation {
   source
   owner
   groups
+  object_ids
+  investigated_observables
+  targets { ...identitySpecificationFields }
```
msprunck, line 291: It's very hard to identify typos in this big graphql query, I mean the query.graphql file. Like we did for the new entities, I would move investigation queries to a new file.

ereteog: Thanks. I added this file, but I replicated all the fragments because I could not reuse `fragments.graphql` which contains `feedbackFields` fragment that is not used in investigation queries.

test/ctia/http/routes/graphql/investigation_test.clj
```diff
@@ -0,0 +1,87 @@
...
+
+(def investigation-1 (fixt/randomize investigation-maximal))
+(def investigation-2 (assoc
+                      (fixt/randomize investigation-maximal)
+                      :source "ngfw"))
+
+(defn prepare-result
+  [investigation]
+  (dissoc investigation
+          :actions
```
msprunck, line 31: The `actions` is exposed in the flanders schema and consequently in the graphql schema. In addition I would add the `search-txt` in the flanders schema. This field is deprecated like the `actions` field but can still be retrieved in old investigations.

ereteog: I add `actions`, but `search-text` contains a dash it cannot be used in graphql queries.

ereteog:
I am blocked to add `search-txt` with a dedicated resolver.
Since it is unrelated with the initial scope of this request (even if related), I prefer to merge this PR and fill a dedicated issue: https://github.com/threatgrid/iroh/issues/3865.

hshahman:
Verified as part of C708850

PR #911, opened 2020-06-29
bundle import, take first available existing entity if duplicates

ereteog:
> Close #https://github.com/threatgrid/iroh/issues/3880

In `bundle/import` route, when duplicate entities are actually found, use the last provided entity to avoid further duplicate and notify it in the response about these duplicates.

<a name="qa">[§](#qa)</a> QA
============================

- use `bundle/import` route without the the `external-key-prefixes` value: `test` to import the following bundle:
[SWE Bundle.json.zip](https://github.com/threatgrid/ctia/files/4847617/SWE.Bundle.json.zip)
For all imported entities, the field `result` should have the value `created`.
- make the same request once again:
For all imported entities, the field `result` should have the value `created`.
- import the same bundle another time but with the `external-key-prefixes` value: `swe-alarm`:
For all imported entities, the field `result` should have the value `existing` with no error.

```
intern: enables CTIA `bundle/import` to deal with duplicated external ids.
```

### DIFF
--- a/src/ctia/bundle/core.clj
+++ b/src/ctia/bundle/core.clj
@@ -166,23 +166,18 @@
                              first
                              :entity
                              with-long-id
-                             ent/un-store)
-          num-old-entities (count old-entities)]
+                             ent/un-store)]
+      (when (< (count old-entities) 1)
+        (log/warn
+         (format
+          (str "More than one entity is "
+               "linked to the external id %s (%s)")
+          external_id
+          (pr-str (map :id old-entities)))))
       (cond-> entity-data
         ;; only one entity linked to the external ID
-        (and old-entity
-             (= num-old-entities 1)) (assoc :result "exists"
-                                            ;;:old-entity old-entity
-                                            :id (:id old-entity))
-        ;; more than one entity linked to the external ID
-        (> num-old-entities 1)
-        (assoc :result "error"
-               :error
-               (format
-                (str "More than one entity is "
-                     "linked to the external id %s (%s)")
-                external_id
-                (pr-str (map :id old-entities))))))
+        old-entity (assoc :result "exists"
+                          :id (:id old-entity))))
     entity-data))
 
 (s/defn with-existing-entities :- [EntityImportData]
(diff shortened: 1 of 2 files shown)


src/ctia/bundle/core.clj
```diff
@@ -166,23 +166,18 @@
                              first
                              :entity
                              with-long-id
-                             ent/un-store)
-          num-old-entities (count old-entities)]
+                             ent/un-store)]
+      (when (< (count old-entities) 1)
```
vemv, line 170: `(< (count old-entities) 1)` means `(zero? (count old-entities))`, which I'm not sure is the right thing?

test/ctia/bundle/core_test.clj
```diff
@@ -44,3 +48,33 @@
             :query "source_ref:*malware*"}
            (sut/relationships-filters "id" {:source_type :malware
                                             :related_to [:source_ref]})))))
+
+(deftest with-existing-entity-test
+  (testing "with-existing-entity"
+    (let [indicator-id-1 (make-id "indicator")
+          indicator-id-2 (make-id "indicator")
+          indicator-id-3 (make-id "indicator")]
+      (testing "1 existing external id"
```
vemv, line 57: You could try an `are` here

As I see it, it's not only about DRYing, but also about making the differences evident (rn I'd have to squint and look for differences very carefully)

src/ctia/bundle/core.clj
ereteog: That's a bug, thx!
I am punished for not testing that the logging is actually done :-).

test/ctia/bundle/core_test.clj
ereteog: Tried to address this with a test fn, it's a little bit "harder" to make it readable with logging test. Tell me what you think.

ereteog:
@vemv FYI the force push is just a clean of the test ns and rebase (removed unused fn import)

ereteog:
> Close #https://github.com/threatgrid/iroh/issues/3880

In `bundle/import` route, when duplicate entities are actually found, use the last provided entity to avoid further duplicate and notify it in the response about these duplicates.

hshahman:
Verified in rel-v1.52

PR #914, opened 2020-07-07
Riemann request reporter starter

frenchy64:
Related https://github.com/threatgrid/iroh/issues/2297

I've ported most of riemann-reporter's functionality as-is, as suggested by Yann. See below for an updated sample (July 14th).

```clojure
{:jwt-https---schemas-cisco-com-iroh-identity-claims-user-email
 "gbuisson+qa_sdc_iroh@cisco.com",
 :service "CTIA CTIA",
 :jwt-exp "1487772850",
 :jwt-jti "a268ae7a3-09c9-4149-b495-b98c8c5de666",
 :identity "56bb5f8c-cc4e-4ed3-a91a-c6604287fe32",
 :client-id "iroh-ui",
 :_params "{}\n",
 :_headers
 "{\"Content-Type\" \"application/edn; charset=utf-8\", \"Content-Length\" \"570\"}\n",
 :jwt-https---schemas-cisco-com-iroh-identity-claims-user-idp-id "amp",
 :jwt-https---schemas-cisco-com-iroh-identity-claims-version "1",
 :jwt-https---schemas-cisco-com-iroh-identity-claims-oauth-client-id
 "iroh-ui",
 :jwt-email "gbuisson+qa_sdc_iroh@cisco.com",
 :remote-addr "127.0.0.1",
 :user-email "gbuisson+qa_sdc_iroh@cisco.com",
 :jwt-https---schemas-cisco-com-iroh-identity-claims-oauth-kind
 "session-token",
 :jwt-sub "56bb5f8c-cc4e-4ed3-a91a-c6604287fe32",
 :jwt-https---schemas-cisco-com-iroh-identity-claims-user-id
 "56bb5f8c-cc4e-4ed3-a91a-c6604287fe32",
 :client-name "iroh-ui",
 :org-id "63489cf9-561c-4958-a13d-6d84b7ef09d4",
 :user-id "56bb5f8c-cc4e-4ed3-a91a-c6604287fe32",
 :jwt-https---schemas-cisco-com-iroh-identity-claims-org-id
 "63489cf9-561c-4958-a13d-6d84b7ef09d4",
 :jwt-iss "IROH Auth",
 :jwt-https---schemas-cisco-com-iroh-identity-claims-oauth-client-name
 "iroh-ui",
 :jwt-https---schemas-cisco-com-iroh-identity-claims-org-name
 "IROH Testing",
 :idp-id "amp",
 :status "403",
 :org-name "IROH Testing",
 :jwt-iat "1487168050",
 :uri "/ctia/judgement",
 :jwt-https---schemas-cisco-com-iroh-identity-claims-user-nick
 "gbuisson+qa_sdc_iroh@cisco.com",
 :jwt-https---schemas-cisco-com-iroh-identity-claims-scopes
 "[\"casebook\" \"global-intel\" \"private-intel\" \"collect\" \"enrich\" \"inspect\" \"integration\" \"iroh-auth\" \"response\" \"ui-settings\"]",
 :user-nick "gbuisson+qa_sdc_iroh@cisco.com",
 :metric 8.885594,
 :jwt-nbf "1487167750",
 :request-method ":post"}
```

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: Riemann request reporter starter
```

### DIFF
--- a/src/ctia/http/server.clj
+++ b/src/ctia/http/server.clj
@@ -186,7 +186,7 @@
             (when-let [lifetime (:lifetime-in-sec jwt)]
               {:jwt-max-lifetime-in-sec lifetime}))))
 
-         (get-in @prop/properties [:ctia :log :riemann :enabled])
+         (get-in @properties [:ctia :log :riemann :enabled])
          ((rie/wrap-request-logs "http req"))
 
          access-control-allow-origin
--- a/src/ctia/http/server.clj
+++ b/src/ctia/http/server.clj
@@ -161,6 +161,11 @@
          (:enabled jwt)
          auth-jwt/wrap-jwt-to-ctia-auth
 
+         ;; just after :jwt and :identity is attached to request
+         ;; by rjwt/wrap-jwt-auth-fn below.
+         (get-in @properties [:ctia :log :riemann :enabled])
+         (rie/wrap-request-logs "CTIA")
+
          (:enabled jwt)
          ((rjwt/wrap-jwt-auth-fn
            (merge
@@ -186,8 +191,6 @@
             (when-let [lifetime (:lifetime-in-sec jwt)]
               {:jwt-max-lifetime-in-sec lifetime}))))
 
-         (get-in @properties [:ctia :log :riemann :enabled])
-         (rie/wrap-request-logs "http req")
 
          access-control-allow-origin
          (wrap-cors :access-control-allow-origin
--- a/src/ctia/http/server.clj
+++ b/src/ctia/http/server.clj
@@ -187,7 +187,7 @@
               {:jwt-max-lifetime-in-sec lifetime}))))
 
          true
-         ((rie/wrap-request-logs "placeholder"))
+         ((rie/wrap-request-logs "http req"))
 
          access-control-allow-origin
          (wrap-cors :access-control-allow-origin
--- a/src/ctia/http/server.clj
+++ b/src/ctia/http/server.clj
@@ -186,7 +186,7 @@
             (when-let [lifetime (:lifetime-in-sec jwt)]
               {:jwt-max-lifetime-in-sec lifetime}))))
 
-         (:enabled jwt)
+         true
          ((rie/wrap-request-logs "placeholder"))
 
          access-control-allow-origin
--- a/src/ctia/http/server.clj
+++ b/src/ctia/http/server.clj
@@ -187,7 +187,7 @@
               {:jwt-max-lifetime-in-sec lifetime}))))
 
          (get-in @properties [:ctia :log :riemann :enabled])
-         ((rie/wrap-request-logs "http req"))
+         (rie/wrap-request-logs "http req")
 
          access-control-allow-origin
          (wrap-cors :access-control-allow-origin
--- a/src/ctia/http/server.clj
+++ b/src/ctia/http/server.clj
@@ -186,7 +186,7 @@
             (when-let [lifetime (:lifetime-in-sec jwt)]
               {:jwt-max-lifetime-in-sec lifetime}))))
 
-         true
+         (get-in @prop/properties [:ctia :log :riemann :enabled])
          ((rie/wrap-request-logs "http req"))
 
          access-control-allow-origin
--- a/src/ctia/http/server.clj
+++ b/src/ctia/http/server.clj
@@ -10,6 +10,7 @@
             [ctia.http.middleware
              [auth :as auth]
              [ratelimit :refer [wrap-rate-limit]]]
+            [ctia.lib.riemann :as rie]
             [ring-jwt-middleware.core :as rjwt]
             [ring.adapter.jetty :as jetty]
             [ring.middleware
@@ -184,6 +185,9 @@
                                           {}))}))
             (when-let [lifetime (:lifetime-in-sec jwt)]
               {:jwt-max-lifetime-in-sec lifetime}))))
+
+         (:enabled jwt)
+         ((rie/wrap-request-logs "placeholder"))
 
          access-control-allow-origin
          (wrap-cors :access-control-allow-origin
--- a/src/ctia/http/server.clj
+++ b/src/ctia/http/server.clj
@@ -191,7 +191,6 @@
             (when-let [lifetime (:lifetime-in-sec jwt)]
               {:jwt-max-lifetime-in-sec lifetime}))))
 
-
          access-control-allow-origin
          (wrap-cors :access-control-allow-origin
                     (allow-origin-regexps access-control-allow-origin)
--- a/src/ctia/lib/riemann.clj
+++ b/src/ctia/lib/riemann.clj
@@ -91,12 +91,6 @@
            :trace-id (search-event [:trace :id] event)})))
 
 ;; copied from riemann-reporter.core
-(defn find-and-add-metas
-  [e]
-  (let [infos (extract-infos-from-event e)]
-    (into infos e)))
-
-;; copied from riemann-reporter.core
 (defn str-protect [s]
   (str/replace s #"[^a-zA-Z_-]" "-"))
 
@@ -136,7 +130,7 @@
        utils/deep-remove-nils
        (into {})
        utils/deep-filter-out-creds
-       find-and-add-metas
+       extract-infos-from-event ;; difference from iroh (calls find-and-add-metas)
        stringify-values))
 
 ;; based on riemann-reporter.core/send-event
--- a/src/ctia/lib/riemann.clj
+++ b/src/ctia/lib/riemann.clj
@@ -13,6 +13,10 @@
          :remote-addr (str (if-let [xff (get-in request [:headers "x-forwarded-for"])]
                              (peek (str/split xff #"\s*,\s*"))
                              (:remote-addr request)))
+         :request-headers (prn-str (:headers request))
+         :request-body (let [;; HttpInputOverHTTP => string
+                             bstr (pr-str (:body request))]
+                         (subs bstr 0 (min (count bstr) 100)))
          :request-method (str (:request-method request))
          :mask (str (= "Mask" (get-in request [:headers "x-client-app"])))
          :identity (:identity request)
@@ -43,12 +47,12 @@
           (let [ms (ms-elapsed start)]
             (send-request-metrics send-event-fn request
                                   {:metric ms
-                                   :description (str "Request took "
+                                   :tags ["ctia" "http"]
+                                   :description (str "Response took "
                                                      (.format (java.text.DecimalFormat. "#.##")
                                                               (/ ms 1000))
                                                      " seconds")
                                    :service service-name
-                                   :_headers (prn-str (:headers response))
                                    :status (str (:status response))}))
           response)
         (catch ExceptionInfo e
--- a/src/ctia/lib/riemann.clj
+++ b/src/ctia/lib/riemann.clj
@@ -9,15 +9,15 @@
 ;; based on riemann-reporter.core/request-to-event
 (defn request->event
   [request extra-fields]
-  (merge {:uri (str (:uri request))
-          :_params (utils/safe-pprint-str (:params request))
-          :remote-addr (str (if-let [xff (get-in request [:headers "x-forwarded-for"])]
-                              (peek (str/split xff #"\s*,\s*"))
-                              (:remote-addr request)))
-          :request-method (str (:request-method request))
-          :identity (:identity request)
-          :jwt (:jwt request)}
-         extra-fields))
+  (into {:uri (str (:uri request))
+         :_params (utils/safe-pprint-str (:params request))
+         :remote-addr (str (if-let [xff (get-in request [:headers "x-forwarded-for"])]
+                             (peek (str/split xff #"\s*,\s*"))
+                             (:remote-addr request)))
+         :request-method (str (:request-method request))
+         :identity (:identity request)
+         :jwt (:jwt request)}
+        extra-fields))
 
 (defn ms-elapsed
   "provide how much ms were elapsed since `nano-start`."
@@ -165,6 +165,7 @@
         (let [client (-> (select-keys config
                                       [:host :port :interval-in-ms])
                          riemann/tcp-client
+                         #_
                          (riemann/batch-client
                            (or (:batch-size config) 10)))]
           (fn [event]
--- a/src/ctia/lib/riemann.clj
+++ b/src/ctia/lib/riemann.clj
@@ -127,7 +127,6 @@
         (let [client (-> (select-keys config
                                       [:host :port :interval-in-ms])
                          riemann/tcp-client
-                         #_
                          (riemann/batch-client
                            (or (:batch-size config) 10)))]
           (fn [event]
--- a/src/ctia/lib/riemann.clj
+++ b/src/ctia/lib/riemann.clj
@@ -53,6 +53,11 @@
            :idp-id (search-event [:idp :id] event)
            :trace-id (search-event [:trace :id] event)})))
 
+(defn find-and-add-metas
+  [e]
+  (let [infos (extract-infos-from-event e)]
+    (into infos e)))
+
 ;; copied from riemann-reporter.core
 (defn str-protect [s]
   (str/replace s #"[^a-zA-Z_-]" "-"))
@@ -93,7 +98,7 @@
        utils/deep-remove-nils
        (into {})
        utils/deep-filter-out-creds
-       extract-infos-from-event ;; difference from iroh (calls find-and-add-metas)
+       find-and-add-metas
        stringify-values))
 
 ;; based on riemann-reporter.core/send-event
--- a/src/ctia/lib/riemann.clj
+++ b/src/ctia/lib/riemann.clj
@@ -6,39 +6,37 @@
             [ctia.lib.utils :as utils])
   (:import [clojure.lang ExceptionInfo]))
 
+;; based on riemann-reporter.core/request-to-event
 (defn request->event
   [request extra-fields]
-  (into {:uri (str (:uri request))
-         :_params (utils/safe-pprint (:params request))
-         :remote-addr (str (if-let [xff (get-in request [:headers "x-forwarded-for"])]
-                             (peek (str/split xff #"\s*,\s*"))
-                             (:remote-addr request)))
-         :request-headers (pr-str (:headers request))
-         :request-body (let [;; HttpInputOverHTTP => string
-                             bstr (pr-str (:body request))]
-                         (subs bstr 0 (min (count bstr) 100)))
-         :request-method (str (:request-method request))
-         :mask (str (= "Mask" (get-in request [:headers "x-client-app"])))
-         :identity (:identity request)
-         :jwt (:jwt request)}
-        extra-fields))
+  (merge {:uri (str (:uri request))
+          :_params (utils/safe-pprint-str (:params request))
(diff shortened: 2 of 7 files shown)


src/ctia/http/server.clj
```diff
@@ -185,6 +186,9 @@
             (when-let [lifetime (:lifetime-in-sec jwt)]
               {:jwt-max-lifetime-in-sec lifetime}))))
 
+         (:enabled jwt)
```
yogsototh, line 189: I think this should always be `true` here.
Or perhaps should be configurable to be on or off via a configuration. But this should not be linked to the `jwt` property.

src/ctia/lib/riemann.clj
```diff
@@ -0,0 +1,115 @@
+(ns ctia.lib.riemann
+  (:require [clojure.string :as str]
+            [clojure.tools.logging :as log]
+            [riemann.client :as riemann]
+            [ctia.properties :as prop]
+            [ctia.lib.utils :as utils])
+  (:import [clojure.lang ExceptionInfo]))
+
+(defn request-to-event
```
vemv, line 9: `request->event` is a quite usual naming convention

frenchy64: Indeed, 90% of this code is copy-pasted from RiemannReporterService as I get my bearings. Will circle back.

src/ctia/http/server.clj
frenchy64: Fixed and added a `ctia.log.riemann.enabled` property.

src/ctia/lib/riemann.clj
```diff
@@ -0,0 +1,124 @@
...
+(defn wrap-request-logs
+  "Middleware to log all incoming connections to Riemann"
+  [service-name]
+  (let [{enabled? :enabled :as config}
+        (get-in @prop/properties [:ctia :log :riemann])]
+    (if-not enabled?
+      identity
+      (let [_ (log/info "riemann metrics reporting")
+            send-event-fn 
+            (let [client (riemann/tcp-client
```
msprunck, line 109: With the large number of requests addressed to CTIA I would wrap the client in a `RiemannBatchClient` with [`batch-client`](https://github.com/riemann/riemann-clojure-client/blob/master/src/riemann/client.clj#L114). See its usage in [IROH](https://github.com/threatgrid/iroh/blob/master/services/riemann-reporter/src/riemann_reporter/core.clj#L92)

src/ctia/lib/riemann.clj
```diff
@@ -0,0 +1,124 @@
...
+                   (utils/safe-pprint prepared-event))))))
+
+(defn wrap-request-logs
+  "Middleware to log all incoming connections to Riemann"
+  [service-name]
+  (let [{enabled? :enabled :as config}
+        (get-in @prop/properties [:ctia :log :riemann])]
+    (if-not enabled?
+      identity
+      (let [_ (log/info "riemann metrics reporting")
```
msprunck, line 107:
```suggestion
      (let [_ (log/info "Riemann request logging initialization")
```

src/ctia/lib/riemann.clj
```diff
@@ -0,0 +1,124 @@
...
+            [clojure.tools.logging :as log]
+            [riemann.client :as riemann]
+            [ctia.properties :as prop]
+            [ctia.lib.utils :as utils])
+  (:import [clojure.lang ExceptionInfo]))
+
+(defn request->event
+  [request extra-fields]
+  (into {:uri (str (:uri request))
+         :_params (utils/safe-pprint (:params request))
```
msprunck, line 12: Why this param is prefixed with `_`? I don't remember :(

src/ctia/lib/riemann.clj
```diff
@@ -0,0 +1,124 @@
...
+         :remote-addr (str (if-let [xff (get-in request [:headers "x-forwarded-for"])]
+                             (peek (str/split xff #"\s*,\s*"))
+                             (:remote-addr request)))
+         :request-headers (prn-str (:headers request))
+         :request-body (let [;; HttpInputOverHTTP => string
+                             bstr (pr-str (:body request))]
+                         (subs bstr 0 (min (count bstr) 100)))
+         :request-method (str (:request-method request))
+         :mask (str (= "Mask" (get-in request [:headers "x-client-app"])))
```
msprunck, line 21: `Mask` is the Threat Grid frontend. It has no meaning in our context

src/ctia/http/server.clj
```diff
@@ -185,6 +186,9 @@
             (when-let [lifetime (:lifetime-in-sec jwt)]
               {:jwt-max-lifetime-in-sec lifetime}))))
 
+         true
```
yogsototh, line 189: You should probably check the configuration value here.
Otherwise this will be confusing to see the middleware is not used depending on the configuration elsewhere.

src/ctia/lib/riemann.clj
```diff
@@ -0,0 +1,124 @@
...
+        (riemann/send-event conn prepared-event))
+      (when-not (= "log" (:event-type prepared-event))
+        (log/warnf "Riemann doesn't seem configured. Event: %s"
+                   (utils/safe-pprint prepared-event))))))
+
+(defn wrap-request-logs
+  "Middleware to log all incoming connections to Riemann"
+  [service-name]
+  (let [{enabled? :enabled :as config}
+        (get-in @prop/properties [:ctia :log :riemann])]
```
yogsototh, line 104: I think you should not worry about the configuration in the middleware here.
You should check the `enabled` value in the thread that append middleware. (see my previous comment)

src/ctia/lib/riemann.clj
```diff
@@ -0,0 +1,124 @@
...
+    (if-not enabled?
+      identity
+      (let [_ (log/info "riemann metrics reporting")
+            send-event-fn 
+            (let [client (riemann/tcp-client
+                           (select-keys config
+                                        [:host :port :interval-in-ms]))]
```
yogsototh, line 111: You should probably don't want to query the config for each call.
You should probably get the client in a let before returning the function.

src/ctia/lib/riemann.clj
```diff
@@ -0,0 +1,124 @@
...
+
+(defn wrap-request-metrics [handler service-name send-event-fn]
+  (fn [request]
+    (let [start (System/nanoTime)]
+      (try
+        (when-let [response (handler request)]
+          (let [ms (ms-elapsed start)]
+            (send-request-metrics send-event-fn request
+                                  {:metric ms
+                                   :tags ["ctia" "http"]
```
yogsototh, line 50: There might be a good reason we do not use `tags` in IROH. 
You should check that this does not make something funny once moved to ES/Kibana.

src/ctia/lib/riemann.clj
```diff
@@ -0,0 +1,124 @@
...
+                                   :description (str "Response took "
+                                                     (.format (java.text.DecimalFormat. "#.##")
+                                                              (/ ms 1000))
+                                                     " seconds")
```
yogsototh, line 54: I wouldn't lose resources for repeating that message as a string. I think having a `metric` is clear enough. We try to enforce the use of `ms`.

src/ctia/lib/riemann.clj
```diff
@@ -0,0 +1,124 @@
...
+                                {:metric (ms-elapsed start)
+                                 :service (str service-name " error")
+                                 :description (.getMessage e)
+                                 :stacktrace (.getStackTrace e)
+                                 :status "500"
+                                 :error "true"})
+          (throw e))))))
+
+(defn prepare-event [event]
+  {:user-id (get-in event [:user :id])})
```
yogsototh, line 83: The `prepare-event` in IROH https://github.com/threatgrid/iroh/blob/master/services/riemann-reporter/src/riemann_reporter/core.clj#L172 does not only take care of adding useful metas it takes care of sending flattened hash-map, remove potential sensible informations that should never be visible in the logs (like API keys, raw JWT, etc...).

You should probably copy it. Except that `find-and-add-metas` function that you could get rid of by taking care of declaring manually `user-id`, `user-name`, `user-email`, `org-id`, `org-name`, `idp-id`, `client-id`, `client-name` and `trace-id` directly by using the `identity` value of the request.

It will be very important to have the same structure between IROH and CTIA logs otherwise it will be a nightmare for the people looking at the logs.

src/ctia/lib/riemann.clj
```diff
@@ -0,0 +1,124 @@
...
+          (wrap-request-metrics handler
+                                service-name
+                                send-event-fn))))))
+
+(defn log
+  "Produce a log and send an event to riemann.
+  The event should contains the fields :level to specify the log level.
+  The r-service parameter should be the service name that appear in riemann."
+  [_ r-service msg event]
+  '...)
```
yogsototh, line 124: I don't think you need to create log function for riemann in the PR.
More than that. We intend to use riemann directly via `timbre` and use the `timbre/with-context` to really improve our logging API inside IROH. And This is a bit refactorisation on its own for CTIA too.

src/ctia/lib/riemann.clj
yogsototh: This is because kibana will not index field starting with `_`.

src/ctia/lib/riemann.clj
```diff
@@ -0,0 +1,124 @@
...
+  (:import [clojure.lang ExceptionInfo]))
+
+(defn request->event
+  [request extra-fields]
+  (into {:uri (str (:uri request))
+         :_params (utils/safe-pprint (:params request))
+         :remote-addr (str (if-let [xff (get-in request [:headers "x-forwarded-for"])]
+                             (peek (str/split xff #"\s*,\s*"))
+                             (:remote-addr request)))
+         :request-headers (prn-str (:headers request))
```
yogsototh, line 16: This look dangerous. You should take care to get rid of the `authorization` HEADER before printing it.
There is a good reason not to show the headers in IROH.

In fact I would rather prefer having flattened headers and taking care of removing sensitive headers.
Also there should be only selected headers that should be flattened for all other headers this should be pretty printed.

So typically:

- `:request-header-content-type` ".."
- `:request-header-authorization-type "Bearer"` (hide the JWT/API key but keep the bearer name if it exists, so that could be "Bearer" "api-token" "api-key" or nothing if the header value does not contain any space. So the Authorization header contain a value like "Bearer <JWT>" or "api-token <api-token>" or "<key>", I think it is useful to keep track of the authorization type that could be "bearer", "api-token", etc... or "apikey")
- `:request-header-origin`
- `:request-header-referrer`
- ...

And for all other headers: `:request-headers-unkown (pr-str (dissoc (:headers req) known-headers))`.
If you don't do that a client could send CURL request with random string for header value and completely fill the index with random keys. And we should really take care about the keys of the events we send cannot grow too much.

src/ctia/lib/riemann.clj
frenchy64: I think I already return a closure that refers to the client, right?

src/ctia/http/server.clj
frenchy64: Done.

src/ctia/lib/riemann.clj
yogsototh: Yes you're right. Sorry my bad.

src/ctia/lib/riemann.clj
frenchy64: Thanks, changed to check at middleware creation.

src/ctia/lib/riemann.clj
frenchy64: Deleted.

src/ctia/lib/riemann.clj
frenchy64: Done.

src/ctia/lib/riemann.clj
frenchy64: Deleted.

src/ctia/lib/riemann.clj
frenchy64: Ok, I copied everything pretty much verbatim. I don't follow about the `identity` value, could you help me change `extract-infos-from-event` to follow your suggestion?

src/ctia/lib/riemann.clj
frenchy64: I deleted the extra stuff, now just has a subset of the functionality of riemann-reporter. Could you take another look to let me know if your concern is still relevant, it's too much information for me to process.

frenchy64:
Thanks for the reviews. I addressed the smaller concerns, and simply ported riemann-reporter to try and address the larger ones that I don't understand yet. I took Yann's advice for this PR to just be a starting point for IROH<->CTIA logging parity, from which we can evolve.

src/ctia/lib/riemann.clj
yogsototh: So `extract-infos-from-event` is a complex function that try to retrieve `user-id` that could be hidden in deeper keys of a nested hash-map. For example, we had both:

```clojure
(send-event {:jwt {"https://schemas.cisco.com/iroh/identity/claims/user/id" "foo"
             "https://schemas.cisco.com/iroh/identity/claims/user/name" "John Doe"
             "https://schemas.cisco.com/iroh/identity/claims/org/id" "..."
             :some-value {:inner "foo" ,,,}})

;; and

(send-event {:identity {:user {:id "foo" :name "John Doe",,,}
                               :org {:id "..." ,,,}})

;;

(send-event {:user-id "foo" :org-id "..." ,,,})
```

So if you would like to search all event for some `user-id` the kibana query became complex. So to simplify that the `extract-infos-from-event` try to search for all value that could be `user-id`. This is not perfect, some events might have multiple potentially different value for `user-id`. But it works most the time. And the situation was bad because we haven't imposed any event format convention.

As you are starting new in CTIA, I think you can enforce such a convention.
So, right now, we have try to generate all the fields I mentioned (`user-id`, `user-name`, etc...).
You can get those by looking at the `:identity` of the ring request.

So the `ring-jwt-middleware/wrap-jwt-auth-fn` middleware take a ring request (a map) and add an `:identity` key containing an `Identity`:

https://github.com/threatgrid/iroh/blob/803120d750f3be604470a0965205128ea27f7f1a/lib/iroh-web/src/iroh_web/request_context.clj#L18--L31

From this you should be able to simply write something like

```clojure
(defn identity->event-fields [req-identity]
    {:user-id (get-in req-identity [:user :id]
     ,,,})
```

And use this function when generating the `event` from the ring request.

src/ctia/lib/utils.clj
```diff
@@ -1,11 +1,59 @@
...
+  (:require [clojure.string :as string]
+            [clojure.pprint :as pp]
+            [clojure.tools.logging :as log]
+            [clojure.walk :as walk]))
 
-(defn filter-out-creds [m]
+;; copied from log-helper.safe
+(def to-obfuscate-pattern
+  (let [keywords-to-obfuscate ["authorization"
+                               "creds"
```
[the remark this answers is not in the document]
yogsototh: note `jwt` must not be obfuscated in IROH. Some event use the `jwt` prefix but only contains the decoded claims, not the `jwt` itself.

src/ctia/lib/riemann.clj
yogsototh: Another note, if you keep using the `extract-infos-from-event` it will work but might just not be optimal.
And if you start to send big events it might potentially affect the response time.
In IROH there is a test trying the check that it this function is fast enough. 

So if you don't feel comfortable to change it right away you could keep it and improve the code later.

src/ctia/lib/riemann.clj
```diff
@@ -0,0 +1,167 @@
...
+        (log/warnf "A Problem occured while sending request metrics event:\n\n%s\n\nException:\n%s"
+                   (utils/safe-pprint-str event)
+                   (utils/safe-pprint-str e))
+        (when (nil? send-event-fn)
+          (log/warn "The send-event-fn looks empty. This is certainly due to a configuration problem or perhaps simply a code bug."))))))
+
+;; should align with riemann-reporter.core/extract-infos-from-event
+(defn extract-infos-from-event
+  [event]
+  (let [search-event get-in]
```
yogsototh, line 42: So the `get-in` will not work here. Because I thin the `event` is already flattened when this function is called.

I think, if you are lucky, the function that might work is to rename all the keys starting with `identity-` to remove this `identity-` prefix. So replacing:

```clojure
{:identity-user-id ,,, 
 :identity-user-name ,,,
 :foo ,,,}

;; by

{:user-id ,,,
 :user-name ,,,
 :foo ,,,}
```

As I said, as a first step you could copy the `search-event` function verbatim from IROH, then improve build a better event from request before calling the `send-event` (see my other comments).

frenchy64:
I've rolled my own `extract-infos-from-event` that seems reliable enough for now. It at least coincides with a lot of fields in Kibana.

Here's what an event now looks like before it's sent to Riemann. Is this close enough? Also have a PR enabling logging for INT. https://github.com/threatgrid/tenzin-config/pull/264

```clojure
{:jwt-https---schemas-cisco-com-iroh-identity-claims-user-email
 "gbuisson+qa_sdc_iroh@cisco.com",
 :service "CTIA CTIA",
 :jwt-exp "1487772850",
 :jwt-jti "a268ae7a3-09c9-4149-b495-b98c8c5de666",
 :identity "56bb5f8c-cc4e-4ed3-a91a-c6604287fe32",
 :client-id "iroh-ui",
 :_params "{}\n",
 :_headers
 "{\"Content-Type\" \"application/edn; charset=utf-8\", \"Content-Length\" \"570\"}\n",
 :jwt-https---schemas-cisco-com-iroh-identity-claims-user-idp-id "amp",
 :jwt-https---schemas-cisco-com-iroh-identity-claims-version "1",
 :jwt-https---schemas-cisco-com-iroh-identity-claims-oauth-client-id
 "iroh-ui",
 :jwt-email "gbuisson+qa_sdc_iroh@cisco.com",
 :remote-addr "127.0.0.1",
 :user-email "gbuisson+qa_sdc_iroh@cisco.com",
 :jwt-https---schemas-cisco-com-iroh-identity-claims-oauth-kind
 "session-token",
 :jwt-sub "56bb5f8c-cc4e-4ed3-a91a-c6604287fe32",
 :jwt-https---schemas-cisco-com-iroh-identity-claims-user-id
 "56bb5f8c-cc4e-4ed3-a91a-c6604287fe32",
 :client-name "iroh-ui",
 :org-id "63489cf9-561c-4958-a13d-6d84b7ef09d4",
 :user-id "56bb5f8c-cc4e-4ed3-a91a-c6604287fe32",
 :jwt-https---schemas-cisco-com-iroh-identity-claims-org-id
 "63489cf9-561c-4958-a13d-6d84b7ef09d4",
 :jwt-iss "IROH Auth",
 :jwt-https---schemas-cisco-com-iroh-identity-claims-oauth-client-name
 "iroh-ui",
 :jwt-https---schemas-cisco-com-iroh-identity-claims-org-name
 "IROH Testing",
 :idp-id "amp",
 :status "403",
 :org-name "IROH Testing",
 :jwt-iat "1487168050",
 :uri "/ctia/judgement",
 :jwt-https---schemas-cisco-com-iroh-identity-claims-user-nick
 "gbuisson+qa_sdc_iroh@cisco.com",
 :jwt-https---schemas-cisco-com-iroh-identity-claims-scopes
 "[\"casebook\" \"global-intel\" \"private-intel\" \"collect\" \"enrich\" \"inspect\" \"integration\" \"iroh-auth\" \"response\" \"ui-settings\"]",
 :user-nick "gbuisson+qa_sdc_iroh@cisco.com",
 :metric 8.885594,
 :jwt-nbf "1487167750",
 :request-method ":post"}
```

PR #918, opened 2020-07-10
Cleanup project

vemv:
A few fixes moslty oriented towards making `(refresh)` pass (if you happen to use it. Right now it's not set up - I inject `refresh` via Lein).

* `ctia.task.migration.migrate-es-stores-test`: prevent spawning a server at compile time
  * loading this ns from `(refresh)` would create a blocking http server.
* Correctly place return value type hints
  * https://gist.github.com/pjstadig/6fa4ebf465f6d438b4dcc2a5bfbb3db6
  * https://clojure.org/reference/java_interop#typehints (last paragraph)
* Move Uberjar aspects to the :uberjar profile
  * `:aot` and `:main` emit .class files even when developing/testing, which is a usual cause for opaque errors.
* `:dev` profile: add `:source-paths ["dev"]`
  * Necessary for user.clj to be possibly loaded via `(refresh)`
* Add a missing argument

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: Cleanup project
```

### DIFF
--- a/test/ctia/task/migration/migrate_es_stores_test.clj
+++ b/test/ctia/task/migration/migrate_es_stores_test.clj
@@ -44,15 +44,25 @@
                   helpers/fixture-properties:clean
                   es-helpers/fixture-properties:es-store]))
 
-(setup!) ;; init migration conn and properties
-(def es-props (get-in @props/properties [:ctia :store :es]))
-(def es-conn (connect (:default es-props)))
-(def migration-index (get-in es-props [:migration :indexname]))
+;; This a is a `defn` to prevent side-effects at compile time
+(defn es-props []
+  (get-in @props/properties [:ctia :store :es]))
+
+;; This a is a `defn` to prevent side-effects at compile time
+(defn es-conn []
+  (connect (:default (es-props))))
+
+;; This a is a `defn` to prevent side-effects at compile time
+(defn migration-index []
+  (get-in (es-props) [:migration :indexname]))
 
 (defn fixture-clean-migration [t]
+  (defonce setup ;; ugly, but must be done in order to prevent an indefinitely blocking call (which can affect code reloading, or re-running this ns's tests)
+    (setup!) ;; init migration conn and properties
+    )
   (t)
-  (es-index/delete! es-conn "v0.0.0*")
-  (es-index/delete! es-conn (str migration-index "*")))
+  (es-index/delete! (es-conn) "v0.0.0*")
+  (es-index/delete! (es-conn) (str (migration-index) "*")))
 
 (use-fixtures :each
   (join-fixtures [helpers/fixture-ctia
@@ -150,8 +160,8 @@
       (rollover-post-bulk)
       ;; insert malformed documents
       (doseq [store-type store-types]
-        (es-index/get es-conn
-                      (str (get-in es-props [store-type :indexname]) "*")))
+        (es-index/get (es-conn)
+                      (str (get-in (es-props) [store-type :indexname]) "*")))
       (sut/migrate-store-indexes {:migration-id "test-3"
                                   :prefix       "0.0.0"
                                   :migrations   [:__test]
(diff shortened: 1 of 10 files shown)


test/ctia/task/migration/migrate_es_stores_test.clj
```diff
@@ -44,12 +44,22 @@
...
+(defn es-conn []
+  (connect (:default es-props)))
+
+;; This a is a `defn` to prevent side-effects at compile time
+(defn migration-index []
+  (get-in es-props [:migration :indexname]))
 
 (defn fixture-clean-migration [t]
+  (defonce setup ;; ugly, but must be done in order to prevent an indefinitely blocking call (which can affect code reloading, or re-running this ns's tests)
+    (setup!) ;; init migration conn and properties
```
vemv, line 61: This solution is not ideal (but not worse than the current one)

Ideally we'd teardown the setup (i.e. stop the underlying http server). I wasn't able to figure out how.

frenchy64: Perhaps `shutdown-ctia!`? But I really don't know if it's designed for iteration in tests.

How about something like:
```clojure
(defonce setup nil)
(defn fixture-clean-fixture [t]
  (alter-var-root! #'setup #(or % (setup!)))
  ...)
```
Or maybe just add a `declare` for the var to keep the ns graph stable.

vemv:
@frenchy64 : done!

> Perhaps shutdown-ctia!? But I really don't know if it's designed for iteration in tests.

Yeah, from what I could see, it looks like making CTIA testable from the repl (especially _multiple times_) would take some work, as there are some global atoms, shutdown hooks etc.

Could be done the day someone strongly feels the need :)

frenchy64:
As we realized in Webex, we might need to update the README to explain `lein run` is now `lein run -m ctia.main`.

An alternative approach might be to keep the configuration as-is and use `requiring-resolve` in `ctia.main` to limit the number of classes generated. I think `trapperkeeper.main` does something like this.

test/ctia/task/migration/migrate_es_stores_test.clj
```diff
@@ -37,22 +37,41 @@
            (java.lang AssertionError)
            (clojure.lang ExceptionInfo)))
 
+(declare setup)
+
+(defn fixture-setup [t]
+  (when-not setup
```
frenchy64, line 43: I don't think this branch is ever taken.

```clojure
user=> (declare foo)
#'user/foo
user=> foo
#object[clojure.lang.Var$Unbound 0x7c8326a4 "Unbound: #'user/foo"]
```

frenchy64: Perhaps `(bound? #'setup)`?

vemv:
> As we realized in Webex, we might need to update the README to explain lein run is now lein run -m ctia.main.

Done!

> An alternative approach might be to keep the configuration as-is and use requiring-resolve in ctia.main to limit the number of classes generated. I think trapperkeeper.main does something like this.

Saw you do that in https://github.com/threatgrid/ctia/pull/921 , it's a great technique. It also avoids other categories of issues (e.g. you want to set `*assert* false` in production... otherwise a vanilla `require` from our `main` can load code before the binding takes effect)

> I don't think this branch is ever taken.

Damn! Thanks.

frenchy64:
Just a minor comment that you probably want to `defonce` the `setup` var. I feel bad for sending you in a circle, your original solution was almost ok with me--not worth all this extra time, sorry!

vemv:
But `defonce` can be gone as well on a `(refresh)`...

Since CTIA doesn't support repl-driven tests, it doesn't seem to matter much which of either broken choice we use ;p

Of course, at some point we should fix the root cause, namely that there's no teardown step, and (AIUI) that we're running a http server for obtaining something just something smaller: migrations.

frenchy64:
Whoops I forgot.

PR #919, opened 2020-07-14
Asset, AssetMapping entities and corresponding tests

agzam:
To be merged only after https://github.com/threatgrid/ctim/pull/289 is merged and new artifact deployed.

We should consider it hiding in Swagger, until we build the rest of the schemas, tests, etc.

> **Epic** #
> Close #
> Related #

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
```


src/ctia/bulk/routes.clj
```diff
@@ -45,25 +46,27 @@
   (GET "/" []
        :return (s/maybe Bulk)
        :summary "GET many entities at once"
-       :query-params [{actors          :- [Reference] []}
+       :query-params [{actors :- [Reference] []}
```
gbuisson, line 49: I see parts on your changes where you align and others like here where you remove the alignement, how do you define what should be aligned or not?

src/ctia/stores/es/mapping.clj
```diff
@@ -122,6 +122,8 @@
 (def related-actors
   {:properties (assoc related
                       :actor_id token)})
+(def related-assets
```
gbuisson, line 125: not sure this is needed anymore, this pattern exists before we moved on to STIX 2.0 influenced schemas and the Relationship model.

test/ctia/entity/asset_test.clj
```diff
@@ -0,0 +1,92 @@
...
+  (store/test-for-each-store
+   (fn []
+     (helpers/set-capabilities! "foouser" ["foogroup"] "user" auth/all-capabilities)
+     (whoami-helpers/set-whoami-response http/api-key "foouser" "foogroup" "user")
+     (entity-crud-test
+      {:entity           "asset"
+       :example          new-asset-maximal
+       :invalid-tests?   false
+       :update-tests?    true
+       :search-tests?    false
```
gbuisson, line 51: is there a specific reason why you disabled some tests?

src/ctia/bulk/routes.clj
agzam: I generally prefer aligning things with `M-x clojure-align <C-SPC>` where that makes sense. Sometimes it gets wonky e.g. in let forms. Then I don't do that. I see that most of our code is not aligned, if my habit becomes a problem, let me know I will try not to commit any code that is aligned.
This particular piece I think got skewed because @ereteog  and I made commits on top of each other's.

src/ctia/stores/es/mapping.clj
agzam: Good to know. Honestly, I still don't exactly know what I'm doing. I just copy-pasta a bunch of things and changed it slightly.

test/ctia/entity/asset_test.clj
agzam: No... I think that's particular piece is the result of copy-pasta of some other test :) I'm still learning how things work. I'll try to enable them and see how it goes. Thank you for pointing that out! 👍

src/ctia/bulk/routes.clj
agzam: Oh, I was wrong. This piece is not aligned because clojure-align doesn't align them, they are not "related".

PR #920, opened 2020-07-15
adds direct lifecycle methods to start and stop server

agzam:
This is not to be merged "as is", at this point is more like a question.

 I would like to have direct "start" and "stop" methods so they can be hooked to CIDER. CIDER has `cider-ns-refresh-before-fn` and `cider-ns-refresh-after-fn` vars, so whenever you run `(cider-ns-refresh)`, it executes those functions. The problem with that - CIDER requires them to be in "direct" form so you set them like this `(setq cider-ns-refresh-before-fn "stop")`, that means, you need to expose functions that take no arguments.

So my question is: what would be the best place/way to expose those? Maybe someone can suggest a better approach?

### DIFF
--- a/src/ctia/main.clj
+++ b/src/ctia/main.clj
@@ -1,6 +1,7 @@
 (ns ctia.main
   (:gen-class)
   (:require
+   [ctia.http.server :as http-server]
    [ctia.init :refer [log-properties start-ctia!]]))
 
 (defn -main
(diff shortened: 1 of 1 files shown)


frenchy64:
I think a zero-argument call to `start!` will do what you want, since all the options are false by default. For shutting down, I found this testing fixture you might want to emulate, which has a zero-argument public shutdown function.

https://github.com/threatgrid/ctia/blob/758a98f2debbec1825363c75d761bcf016bb19e2/test/ctia/test_helpers/core.clj#L157-L161

agzam:
Oh cool, it seems I just needed to do this:

```
(setq cider-ns-refresh-before-fn "ctia.shutdown/shutdown-ctia!"
      cider-ns-refresh-after-fn "ctia.init/start-ctia!")
```
Thank you @frenchy64

PR #926, opened 2020-07-21
Add route to link Incidents to Investigations

frenchy64:
Close https://github.com/threatgrid/iroh/issues/2431

Extends `/incident/:id/link` to also support linking to Investigations.

<a name="qa">[§](#qa)</a> QA
============================

This PR broadens the scope of the `/incident/:id/link` route. For a body, it now accepts *one of* an `investigation_id` or `casebook_id` field, as well as some other optional keys (at the moment, just `tlp`).

Please follow these steps to create a relationship between an incident and an investigation, and verify it was created.

1. Visit `https://private.intel.int.iroh.site/index.html#/Incident/post_ctia_incident__id__link` (or equivalent)
2. Add a known incident id to the `id` param.
3. Add a body of the form `{ "investigation_id" : "<known-investigation-id>", ...}` with any relevant optional keys, then execute.
4. The response should be `200` and contain an `id` field. This is the relationship that was created.
5. Verify the `id` returned in step 4 corresponds to a relationship from the incident id to the investigation id.

The previous step should also still work for linking casebooks, via the `casebook_id` field. This behavior should be unchanged.

Please also verify providing zero or more than one of `investigation_id` or `casebook_id` fields in the body of the `/incident/:id/link` route triggers a 400 error response. Error messages should also be descriptive enough to diagnose the problem immediately.

```
intern: Add route to link Incidents to Investigations
```

### DIFF
--- a/src/ctia/entity/relationship.clj
+++ b/src/ctia/entity/relationship.clj
@@ -1,5 +1,6 @@
 (ns ctia.entity.relationship
-  (:require [compojure.api.sweet :refer [POST]]
+  (:require [clojure.string :as str]
+            [compojure.api.sweet :refer [POST]]
             [ctia
              [properties :refer [get-http-show]]
              [store :refer :all]]
@@ -72,13 +73,22 @@
   (st/merge PagingParams
             RelationshipFieldsParam))
 
-(s/defschema IncidentLinkRequest
-  {;; Exactly one of:
-   (s/optional-key :casebook_id) Reference
-   (s/optional-key :investigation_id) Reference
+(s/defschema IncidentLinkRequestOptional
+  {(s/optional-key :tlp) TLP})
 
-   ;; Extra keys
-   (s/optional-key :tlp) TLP})
+(def incident-link-source-types
+  [:casebook_id
+   :investigation_id])
+
+(s/defschema IncidentLinkRequest
+  (st/merge
+    ;; Exactly one of these fields is required for a
+    ;; valid request.
+    (into {}
+          (map (fn [k]
+                 [(s/optional-key k) Reference]))
+          incident-link-source-types)
+    IncidentLinkRequestOptional))
 
 (def incident-link-route
   (POST "/:id/link" []
@@ -93,12 +103,24 @@
                         :create-relationship}
         :auth-identity identity
         :identity-map identity-map
-        (let [source-type-kw (let [types (select-keys link-req
-                                                      [:casebook_id
-                                                       :investigation_id])]
-                               (when-not (= 1 (count types))
-                                 (bad-request! {:error "Missing source type"}))
-                               (-> types first key))
+        (let [source-type-kw (let [one-of-kws [:casebook_id
+                                               :investigation_id]
+                                   provided-types (select-keys link-req one-of-kws)]
+                               (when-not (= 1 (count provided-types))
+                                 (bad-request!
+                                   {:error
+                                    (str "Please provide exactly one of the following fields: "
+                                         (str/join ", " (->> one-of-kws (map name) sort))
+                                         "\n"
+                                         (if-let [given-str (when (seq provided-types)
+                                                              (->> provided-types
+                                                                   keys
+                                                                   (map name)
+                                                                   sort
+                                                                   (str/join ", ")))]
+                                           (str "Provided: " given-str)
+                                           "None provided."))}))
+                               (-> provided-types first key))
               additional-required-capabilities
               (case source-type-kw
                 :casebook_id #{:read-casebook}
--- a/src/ctia/entity/relationship.clj
+++ b/src/ctia/entity/relationship.clj
@@ -137,6 +137,69 @@
                       un-store)]
               (created stored-relationship))))))
 
+(s/defschema IncidentInvestigationLinkRequest
+  {:investigation_id Reference})
+
+(def incident-investigation-link-route
+  (POST "/:id/link" []
+        :return rs/Relationship
+        :body [link-req IncidentInvestigationLinkRequest
+               {:description "an Incident Link request"}]
+        :summary "Link an Incident to an Investigation"
+        :path-params [id :- s/Str]
+        :capabilities #{:read-incident
+                        :read-investigation
+                        :read-relationship
+                        :create-relationship}
+        :auth-identity identity
+        :identity-map identity-map
+        (let [incident (read-store :incident
+                                   read-record
+                                   id
+                                   identity-map
+                                   {})
+              investigation-id (-> link-req
+                                   :investigation_id
+                                   long-id->id
+                                   :short-id)
+              investigation (when investigation-id
+                              (read-store :investigation
+                                          read-record
+                                          investigation-id
+                                          identity-map
+                                          {}))
+              target-ref (short-id->long-id id
+                                            get-http-show)]
+          (cond
+            (or (not incident)
+                (not target-ref))
+            (not-found {:error "Invalid Incident id"})
+            (not investigation)
+            (bad-request {:error "Invalid Investigation id"})
+            :else
+            (let [{:keys [investigation_id]} link-req
+                  new-relationship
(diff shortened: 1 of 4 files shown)


test/ctia/entity/relationship_test.clj
```diff
@@ -172,11 +178,84 @@
...
+
+             {relationship-status :status
+              relationship-response :parsed-body}
+             (get (str "ctia/relationship/" (-> (:id link-response)
+                                                long-id->id
+                                                :short-id))
+                  :headers {"Authorization" "45c1f5e3f05d0"})
+             _ (is (= 200 relationship-status))]
+
+         (is (= (:id investigation-body)
```
vemv, line 252: There's the risk of a nil vs nil comparison here

test/ctia/entity/relationship_test.clj
```diff
@@ -172,11 +178,84 @@
...
+
+(deftest incident-investigation-link-routes-test
+  (test-for-each-store
+   (fn []
+     (helpers/set-capabilities! "foouser" ["foogroup"] "user" all-capabilities)
+     (whoami-helpers/set-whoami-response "45c1f5e3f05d0"
+                                         "foouser"
+                                         "foogroup"
+                                         "user")
+     (testing "Incident <-> Investigation link"
```
vemv, line 198: The `testing` block is quite large, but in this case it seems justified since it's a stateful test that "tells a story" out of multiple requests.

Maybe expand `"Incident <-> Investigation link"` to summarize that story, else one finds himself reading a large `let` without knowing exactly what to expect.

msprunck:
I think there is a misunderstanding with the current route:

```
POST /incident/:incident_id/link
{:casebook_id "casebook_id"}
```

The path param defines the resource ID on which the `link` action is performed, in this case the incident ID.

I would just add a `investigation_id` in the body of the current route. IMO There is no need to add a new route.

frenchy64:
Yes that makes sense @msprunck. I'll do that!

frenchy64:
I believe I've addressed all feedback. Please take another look.

src/ctia/entity/relationship.clj
```diff
@@ -71,51 +72,75 @@
...
                         :read-relationship
                         :create-relationship}
         :auth-identity identity
         :identity-map identity-map
-        (let [incident (read-store :incident
+        (let [source-type-kw (let [types (select-keys link-req
+                                                      [:casebook_id
+                                                       :investigation_id])]
+                               (when-not (= 1 (count types))
+                                 (bad-request! {:error "Missing source type"}))
```
msprunck, line 100:
```suggestion
                                 (bad-request! {:error "Please provide a casebook_id OR an investigation_id"}))
```

The current error message is not very explicit.

msprunck: Could you test the cases where:
1. `casebook_id` and `investigation_id` are provided
2. none of them are provided

frenchy64:
Added some tests, and refactored a bit to make them easier to generate in the future.

```clojure
body = {}
; 400 response with message:
;; Please provide exactly one of the following fields: casebook_id, investigation_id
;; None provided.

body = {:casebook_id ":", :investigation_id ":"}
; 400 response with message:
;; Please provide exactly one of the following fields: casebook_id, investigation_id
;; Provided: casebook_id, investigation_id
```

frenchy64:
Also added QA instructions. Unfortunately I pushed this branch directly to the threatgrid repo, so I triggered a double build. The build is also very flaky today. Will merge in the next few hours.

test/ctia/entity/relationship_test.clj
```diff
@@ -195,70 +194,88 @@
...
-         (is (= relationship-response
-                link-response)
-             "Link Response is the created relationship"))))))
+         (is (= (or (:id incident-body)
+                    ::no-incident-body)
+                (or (:target_ref link-response)
+                    ::no-target-ref))
+             "The new Relationship targets the incident")
+         (is (= (or relationship-response
+                    ::no-relationship-response)
```
vemv, line 261: Nice trick!

hshahman:
> Please also verify providing zero or more than one of investigation_id or casebook_id fields in the body of the /incident/:id/link route triggers a 400 error response. Error messages should also be descriptive enough to diagnose the problem immediately.

There are 2 issues wrt above statement.
1. Sending a request with 2 investigation_id to `/link` endpoint is going through and response is 201, which is not expected. Here is the body:
```
{
  "investigation_id": "https://private.intel.test.iroh.site:443/ctia/investigation/investigation-d8c17fe0-d78b-4ac1-b08e-8bfbd549861f",
  "investigation_id": "https://private.intel.test.iroh.site:443/ctia/investigation/investigation-17873611-8caa-4053-ac65-f1d480138d2a",
  "tlp": "green"
}
```


Request:
```
curl -X POST "https://private.intel.test.iroh.site/ctia/incident/incident-1df8534a-1e3d-4802-ab13-b09a3395556e/link" -H "accept: application/json" -H "authorization: Bearer <JWT_TOKEN>" -H "Content-Type: application/json" -d "{ \"investigation_id\": \"https://private.intel.test.iroh.site:443/ctia/investigation/investigation-d8c17fe0-d78b-4ac1-b08e-8bfbd549861f\", \"investigation_id\": \"https://private.intel.test.iroh.site:443/ctia/investigation/investigation-17873611-8caa-4053-ac65-f1d480138d2a\", \"tlp\": \"green\"}"
```

Response contains only one relationship between the incident_id and the **_second_** investigation_id:
```
{
  "schema_version": "1.0.16",
  "target_ref": "https://private.intel.test.iroh.site:443/ctia/incident/incident-1df8534a-1e3d-4802-ab13-b09a3395556e",
  "type": "relationship",
  "source_ref": "https://private.intel.test.iroh.site:443/ctia/investigation/investigation-17873611-8caa-4053-ac65-f1d480138d2a",
  "id": "https://private.intel.test.iroh.site:443/ctia/relationship/relationship-64fc1e84-984b-4d3f-95d5-e29ab5247ee0",
  "tlp": "green",
  "groups": [
    "cc6a35bc-1739-4fcd-a285-aa95adbd5e41"
  ],
  "timestamp": "2020-07-31T18:25:18.239Z",
  "owner": "idb-amp:f6b8d1c4-189b-43eb-b1ba-191d53357a8e",
  "relationship_type": "related-to"
}
```

2. Sending a query to the `/link` endpoint with no investigation_id in the body is returning a 401 as expected, but there is no error message. I think a descriptive error message is required here as mentioned in the QA instructions by Ambrose.

frenchy64:
Wow, I did not think of that case! I had no idea JSON allowed duplicate keys in maps.

frenchy64:
Thanks @hshahman for such a thorough review. After some discussion, we decided that duplicate keys in JSON maps are out of scope for our API (or at least, future work), so we'll just fix 2 at the moment. I'll refine the QA instructions on the next PR.

hshahman:
hmmm, ok...but may I ask  why it is out of scope of our APIs? It doesn't make sense to me why duplicate keys are allowed.

frenchy64:
My interpretation is that you'll find this problem with all SecureX/CTR APIs. This is because none of our code deals with raw JSON, instead we use the Clojure representation, which clobbers duplicate keys.

To tackle the issue, we'd probably need our own JSON parser that checks for duplicate keys. You could imagine that response times would be slower doing these extra checks and the maintenance burden is higher.

I think GB's opinion is that all that work is not worth the benefits of fixing the problem. But that's just my interpretation, feel free to ask him if you want more info.

hshahman:
@frenchy64 thank you for the details above. I'm going to file a new ticket for this issue:
> 2. Sending a query to the /link endpoint with no investigation_id in the body is returning a 401 as expected, but there is no error message. I think a descriptive error message is required here as mentioned in the QA instructions by Ambrose.

So we can pass this ticket since it's almost working as expected.

hshahman:
Verified in rel-v1.54

PR #924, opened 2020-07-17
Port CTIA to Trapperkeeper

frenchy64:
Related https://github.com/threatgrid/iroh/issues/3986

- [ ] add :services entry to ESConnState in clj-momo to share service function with ES stores (for porting `get-access-control` to `get-in-config`)
- [ ] make a `get-in-config` service that's compatible with fixtures before splitting PRs
- [ ] how does pinning the cheshire version interact with the recent jackson-databind pinned version?
- [ ] who is in charge of shutdown on JVM exit signal? should ctia.shutdown be its own service? do we need to replace the thread `join` in ctia.main?
- [ ] ensure we always call `shutdown-agents` when needed
- [ ] double check IEncryption is added to `:services` map when needed
- consider if `ctia-server-service` can be refactored to just provide routes to the TK built-in jetty service

Tentative order of PR's:
- fix `short-id->entity-type` properties bug
- "transform" properties test helpers
- realize-fn wrappers
- routes changes
- graphql changes
- tk port

<a name="actualprs">[§](#actualprs)</a> Actual PRs
============================
1. [x] TK1.1 `Seed routes with global services map argument` (under review) https://github.com/threatgrid/ctia/pull/945
2. [x] TK2.1 `use services map in bulk` (depends on TK1) https://github.com/threatgrid/ctia/pull/950
- next diff https://github.com/frenchy64/ctia/compare/iroh-3986-tk-init-get-in-config-2-bulk...iroh-3986-tk-init
3. [x] TK2.5 `add service args to ctia.bundle.core` (depends on TK2) https://github.com/threatgrid/ctia/pull/954
- next diff https://github.com/frenchy64/ctia/compare/iroh-3986-tk-init-3-bundle...iroh-3986-tk-init
4. [x] TK2.3: `add service args to entity.event`  (depends on TK1.1) https://github.com/threatgrid/ctia/pull/952
- next diff https://github.com/frenchy64/ctia/compare/iroh-3986-tk-init-4-entity-event...iroh-3986-tk-init
5. [x] TK2.7 `add services args to ctia.entity.feed` (depends on TK1.1) https://github.com/threatgrid/ctia/pull/956
- next diff https://github.com/frenchy64/ctia/compare/iroh-3986-tk-init-5-entity-feed...iroh-3986-tk-init
6. [x] TK1.2 `add service arguments to auth.threatgrid` (no dependencies, under review) https://github.com/threatgrid/ctia/pull/948
[...]

PR #933, opened 2020-07-30
Encapsulate global properties atom

frenchy64:
Related https://github.com/threatgrid/iroh/issues/3986

This PR encapsulates the global properties atom to help enable future refactorings, such as to Trapper Keeper's `get-in-config`.

Also fixes miscellaneous small issues.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: Encapsulate global properties atom
```

### DIFF
--- a/test/ctia/test_helpers/core.clj
+++ b/test/ctia/test_helpers/core.clj
@@ -197,10 +197,10 @@
      ~@(for [[name-key fixture-fn] fixture-map]
          `(test/deftest ~(with-meta (symbol (str test-name "-" (name name-key)))
                                     {(keyword name-key) true})
-            (~fixture-fn (fn [] ~@body))))))
+            (~fixture-fn (fn [] (do ~@body)))))))
 
 (defn get-http-port []
-  (get-in (p/read-global-properties) [:ctia :http :port]))
+  (p/get-in-global-properties [:ctia :http :port]))
 
 (def get
   (mthh/with-port-fn get-http-port mthh/get))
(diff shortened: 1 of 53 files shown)


test/ctia/test_helpers/core.clj
```diff
@@ -197,12 +195,12 @@
 (defmacro deftest-for-each-fixture [test-name fixture-map & body]
   `(do
      ~@(for [[name-key fixture-fn] fixture-map]
-         `(clojure.test/deftest ~(with-meta (symbol (str test-name "-" (name name-key)))
-                                   {(keyword name-key) true})
+         `(test/deftest ~(with-meta (symbol (str test-name "-" (name name-key)))
```
vemv, line 198: Not sure if aliases in macros is an improvement - if I removed the alias from this file's `ns` form, the macro would break without a compile-time error

Whereas fqn is less likely to be broken

test/ctia/test_helpers/core.clj
```diff
@@ -197,12 +195,12 @@
...
+                                    {(keyword name-key) true})
             (~fixture-fn (fn [] ~@body))))))
```
vemv, line 200: Maybe `(fn []  (do ~@body))`?

frenchy64:
I implemented @ereteog's suggested `get-in-*` function. I would prefer `global-` in the name so it's easy to refactor these calls to `get-in-config` in the future. Plus, it really is using global/top-level state, so I think it's appropriately named, especially as moving to TK eliminates almost all global state.

test/ctia/test_helpers/core.clj
frenchy64: Thanks for the thoughts. I'm aware of the tradeoffs, and IMO it's more robust overall to use aliases.

PR #934, opened 2020-07-31
Ambigous index names 4012

ereteog:
> Related #https://github.com/threatgrid/iroh/issues/4012 -

This PR prevents CTIA to start with ambiguous index names like `ctia_feed` and `ctia_feedback` (we use `ctia_feeds` in production). It also turns on the Sytem/exit on invalid mappings that was waiting the migration of the vulnerability index that is now done in all environment.

<a name="qa">[§](#qa)</a> QA
============================
No QA is needed.

This PR prevents CTIA from starting when ambiguous index names are defined like `ctia_feed` and `ctia_feedback`. BTW it also [turn on the System/exit](https://github.com/threatgrid/ctia/pull/895) when mapping are badly redifined (happened with vulnerability, but we had to start anyway when it was not migrated yet).
If it happens, in addition of CTIA that fails to start, you should notice the following error log: 
```
CTIA tried to start with an invalid configuration: 
- invalid mapping
- ambiguous index names
```


src/ctia/stores/es/init.clj
```diff
@@ -82,25 +84,42 @@
...
+  ;; retrieve existing indices using wildcard to identify ambiguous index names
+  (let [existing (-> (es-index/get conn (str index "*"))
+                     keys
+                     set)
+        index-pattern (re-pattern (str index "(-\\d{4}.\\d{2}.\\d{2}.*)?"))
+        matching (filter #(re-matches index-pattern (name %))
+                         existing)
+        ambiguous (difference existing (set matching))]
+    (if (seq ambiguous)
+      (do (log/warn (format "ambiguous index names, %s and %s."
```
vemv, line 98: * `%s and %s` -> `index: %s,  ambiguous: %s`
* Always pass `pr-str` before serializing items to be logged/println-ed
  * else, e.g. `(format "%s" nil)` returns `"null"` instead of `"nil"`

test/ctia/stores/es/init_test.clj
```diff
@@ -87,6 +87,52 @@
...
+      (let [test-fn (fn [msg
+                         input-indexname
+                         expected-exited?
+                         expected-output]
+                      (testing msg
+                        (let [ouput (sut/get-existing-indices es-conn input-indexname)]
+                        (when-not expected-exited?
+                          (is (= expected-output ouput)))
+                        (is (= expected-exited? @exited?))
+                        (reset! exited? false))))
```
vemv, line 104: You can move the `(reset! exited? false)` outside and above the `testing`, so that it's clearer that there are two phases ('setup' + 'exercise')

test/ctia/stores/es/init_test.clj
```diff
@@ -87,6 +87,52 @@
       (is (= "2s" refresh_interval)))
     (index/delete! conn (str indexname "*"))))
 
+(deftest get-existing-indices-test
+  (index/delete! es-conn (str indexname "*"))
+  (let [exited? (atom false)
```
vemv, line 92: Maybe I'd rename to `exited?` to `successful?`

That way the 'good ' tests say `true` instead of `false`, which seems more natural. Else it reads like some sort of double negative

test/ctia/stores/es/init_test.clj
```diff
@@ -87,6 +87,52 @@
...
+      (let [test-fn (fn [msg
+                         input-indexname
+                         expected-successful?
+                         expected-output]
+                      (reset! successful? true)
+                      (testing msg
+                        (let [ouput (sut/get-existing-indices es-conn input-indexname)]
+                        (when expected-successful?
+                          (is (= expected-output ouput)))
+                        (is (= expected-successful? @successful?)))))
```
agzam, line 104: This part is confusing, maybe the second `is` should be slurped in, no?

ereteog:
In order to clarify `test-fn` I relocated the `with-redef` around the actually impacted code, added an `assert`, and a comment to explain why testing the output only on success.

test/ctia/stores/es/init_test.clj
ereteog: see my comment below.

PR #937, opened 2020-08-03
Asset properties

agzam:
Related schema changes: https://github.com/threatgrid/ctim/pull/293


src/ctia/entity/asset_properties.clj
```diff
@@ -0,0 +1,134 @@
...
+  {"asset-properties"
+   {:dynamic false
+    :properties
+    (merge
+     em/base-entity-mapping
+     em/sourcable-entity-mapping
+     em/stored-entity-mapping
+     {:identity    em/tg-identity
+      :valid_time  em/valid-time
+      :asset_ref   em/token})}})
```
ereteog, line 48: you need to add the `properties` field mapping unless it will not be available for searching.

src/ctia/entity/asset_properties.clj
```diff
@@ -0,0 +1,134 @@
...
+
+(def asset-properties-mapping
+  {"asset-properties"
+   {:dynamic false
+    :properties
+    (merge
+     em/base-entity-mapping
+     em/sourcable-entity-mapping
+     em/stored-entity-mapping
+     {:identity    em/tg-identity
```
ereteog, line 46: This property is not defined in the schema: https://github.com/threatgrid/ctim/pull/293/files
In case it should be added to the schema (I doubt it), [here](https://github.com/threatgrid/ctia/blob/master/src/ctia/stores/es/mapping.clj#L151) is the mapping you are referring to.

src/ctia/entity/asset_properties.clj
```diff
@@ -0,0 +1,134 @@
...
+    :can-aggregate?           true
+    :histogram-fields         asset-properties-histogram-fields
+    :enumerable-fields        asset-properties-enumerable-fields
+    }))
+
+(def capabilities
+  #{:create-asset-properties
+    :read-asset-properties
+    :delete-asset-properties
+    :search-asset-properties})
```
ereteog, line 115: We have `asset`, `asset-mapping` and `asset-properties`. We define for each of this entities a set of capabilities. Their life cycle seems very correlated, I am wondering if we could not imagine having only `:read-asset`, `create-asset`, `delete-asset`, `search-asset` for all.
@gbuisson WDYT?

gbuisson: No I'd rather not make an exception and have all the capabilities explicit, we will group them under the same root scope.

src/ctia/entity/asset_properties.clj
agzam: Totally forgot about it. Will add the field. Thank you!

src/ctia/entity/asset_properties.clj
agzam: Removed. Not needed here.

PR #938, opened 2020-08-06
fixed asset mapping tests

ereteog:
Adapted an existing test helper for non describable entities.
Fix pagination tests.


agzam:
@ereteog Thank you for unblocking me. You are awesome! I have cherry-picked these changes, because they have to go into two different branches.

PR #939, opened 2020-08-10
Target Record entity and tests

agzam:
>  A `TargetRecord` is a `Sighting` that has no threat or observables associated with it, it's a way of saying they saw a set of observables together as a `Target`. This is useful when attempting to determine what assets were targeted by a threat when we only have weak transient observables, like IPs or dynamically assigned hostnames.

Related schema changes:

https://github.com/threatgrid/ctim/pull/295


agzam:
The only thing that's left to do here is metrics test. This entity is tricky, I can't seem to be able to make it right. `ctia.test-helpers.aggregate/generate-n-entity` generates collection of maps like this:

    {:type "target-record",
    :schema_version "1.0.19",
    :source "cisco:ise:dhcp-server",
    :target
    {:observed_time
      {:start_time "2020-02-07T19:00:00.000Z",
      :end_time "2020-07-06T19:00:00.000Z"}},
    :timestamp "2020-01-02T00:00:00.000Z"}
    
ignoring all enumerable-fields. 

I will fix this later.

src/ctia/entity/target_record.clj
```diff
@@ -0,0 +1,150 @@
...
+(s/defschema TargetRecordFieldsParam
+  {(s/optional-key :fields) [target-record-sort-fields]})
+
+(s/defschema TargetRecordByExternalIdQueryParams
+  (st/merge
+   routes.common/PagingParams
+   TargetRecordFieldsParam))
+
+(def target-record-enumerable-fields
+  [:targets.type])
```
ereteog, line 85: I would add
`:targets.os`, `:targets.internal` and `target.observables.type`

test/ctia/entity/target_record_test.clj
```diff
@@ -0,0 +1,96 @@
...
+        {"Authorization" "45c1f5e3f05d0"}
+        target-record/target-record-fields)
+
+       (pagination/pagination-test
+        "ctia/target-record/search?query=*"
+        {"Authorization" "45c1f5e3f05d0"}
+        target-record/target-record-fields)))))
+
+;; TODO: Fix
+#_(deftest target-record-metric-routes-test
```
ereteog, line 86: I co this branch and I manually tested the TargetRecord metric routes, it works.
Here there is a problem with the schema generation in the test helper.
2 solutions:
- figure out why the schema generation does not work and make it works, it is probably related to the schema generation that makes a field required for the generated entities (in orger to avoid aggregating empty values ;-)) https://github.com/threatgrid/ctia/blob/master/test/ctia/test_helpers/aggregate.clj#L180
```
(defn schema-enumerable-fields
  [schema fields]
  (->> (map (comp keyword first parse-field) fields)
       (st/select-keys schema)
       st/required-keys))
```
If I were you, I would first ensure that it produces the proper schema in your case (if it's actually the problem).
- Do not use the generative testing with a comment that explains why, and implement a non automatic testing (you can take some inspiration on the test helper anyway).

test/ctia/entity/target_record_test.clj
```diff
@@ -0,0 +1,96 @@
...
+                                                         :headers {"Authorization" "45c1f5e3f05d0"})]
+                                           (is (= 200 (:status response)))
+                                           (is (= expected (check-fn response)) desc))
+
+        "*"
+        (fn [r] (-> r :parsed-body first :targets first :type))
+        (-> target-record-sample :targets first :type)
+        "Searching for target-record:targets:type"
+
+        ;; TODO: Add more cases
```
ereteog, line 40: I would add search on `targets.observable.value` and `targets.observable.type`

ereteog (commented):
quick feedback for the metric test

test/ctia/entity/target_record_test.clj
agzam: I think I have fixed the test helpers, so they work with nested structures, see: https://github.com/threatgrid/ctia/pull/939/commits/8d5344b1748142a990a0c9d0e82d8dddfc2715c1
hope that doesn't break other things, I'm running tests now

frenchy64:
Hi Ag, if this PR is going to live much longer, I suggest pushing the branch to your fork and sending a PR from there (and doing the same for future CTIA PRs for now). The problem is we're not well set up for PR's from threatgrid/ctia because Travis triggers a double build (6 jobs). Our open source plan only has quota for 5 jobs so your builds will take 50+ minutes instead of 25m, and if there's a flaky test, you'll be pulling your hair out restarting jobs :)

agzam:
@frenchy64 Got it. will move it to a fork!

agzam:
closed in favor of #965 (forked base)

PR #941, opened 2020-08-11
do not obfuscate external-key-prefixes in events

ereteog:
`bundle/import` route has a parameter `external-key-prefixes` which is obfuscated in events
``` javascript
{"external-key-prefixes" "********"}
```
It is obfuscated because we automatically obfuscate any field that contains some words reflecting potential secret content, including `key`.
I added a list of protected keys, any cleaner option is welcome.

<a name="qa">[§](#qa)</a> QA
============================

import a bundle with `external-key-prefixes` param. Checks that the corresponding event in kibana does not obfuscate the value of this parameter.

### DIFF
--- a/src/ctia/lib/utils.clj
+++ b/src/ctia/lib/utils.clj
@@ -13,7 +13,6 @@
                                "pass"
                                "secret"
                                "tempfile"
-                               "customerKey"
                                "response-url"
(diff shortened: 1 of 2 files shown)


hisans:
QA Verified in 1.55 Release

PR #942, opened 2020-08-17
use `id` field for sorting

ereteog:
> closes #https://github.com/threatgrid/iroh/issues/2870

the `id` field is of type `keyword` and stored as `doc_values` since the mapping migration. This PR makes the migration sort on `id` instead of `_uid` to definitively get rid of fielddata (sorting on `_uid` is performed with fielddata on elaticsearch 5.6, which thus consume a lot of memory due to the cardinality of this field.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

<!-- UNCOMMENT THIS SECTION IF NEEDED

We should test the memory consumption in ES when migration a big index on INT (it should be significantly lower and it should not build fielddata which takes several minutes on big indices).

### DIFF
--- a/resources/ctia-default.properties
+++ b/resources/ctia-default.properties
@@ -142,7 +142,7 @@ ctia.metrics.console.enabled=false
 ctia.metrics.console.interval=60 # every minute
 ctia.metrics.jmx.enabled=false
 
-ctia.log.riemann.enabled=true
+ctia.log.riemann.enabled=false
 ctia.log.riemann.host=127.0.0.1
 ctia.log.riemann.port=5555
 ctia.log.riemann.interval-in-ms=1000 # every second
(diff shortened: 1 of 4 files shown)


yogsototh:
Looking at the failing CI tests: https://travis-ci.org/github/threatgrid/ctia/jobs/718621959#L737

I think you should use `~expected` to make the error message easier to interpret.

resources/ctia-default.properties
```diff
@@ -142,7 +142,7 @@ ctia.metrics.console.enabled=false
 ctia.metrics.console.interval=60 # every minute
 ctia.metrics.jmx.enabled=false
 
-ctia.log.riemann.enabled=true
+ctia.log.riemann.enabled=false
```
yogsototh, line 145: not sure you want to change that.

PR #943, opened 2020-08-19
move migration params in properties

ereteog:
> Related #threatgrid/iroh#3072

This PR moves the migration parameters in properties. 
These parameters were passed through the command line argument. For the next migration we need to add more arguments, in particular arguments for the target cluster and it will excessively increase the command line. Thus we decided to move these parameters in configuration. 
This is in particular the first step toward decorrelating the source cluster properties and target cluster properties for the ES7 migration.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

### DIFF
--- a/src/ctia/task/migration/migrate_es_stores.clj
+++ b/src/ctia/task/migration/migrate_es_stores.clj
@@ -281,7 +281,10 @@
 
 (s/defn ^:always-validate check-migration-params
   [{:keys [prefix
+           restart?
            store-keys]} :- MigrationParams]
+  (when-not restart?
+    (assert prefix "Please provide an indexname prefix for target store creation"))
   (doseq [store-key store-keys]
     (let [index (p/get-in-global-properties [:ctia :store :es store-key :indexname])]
       (when (= (mst/prefixed-index index prefix)
@@ -292,16 +295,14 @@
   true)
 
 (s/defn ^:always-validate run-migration
-  [{:keys [prefix
-           restart?]
-    :as params} :- MigrationParams]
-  (when-not restart?
-    (assert prefix "Please provide an indexname prefix for target store creation"))
+  []
   (log/info "migrating all ES Stores")
   (try
     (mst/setup!)
-    (check-migration-params params)
-    (migrate-store-indexes params)
+    (doto (p/get-in-global-properties [:ctia :store :migration])
+      clojure.pprint/pprint
+      check-migration-params
+      migrate-store-indexes)
     (log/info "migration complete")
     (catch AssertionError e
        (log/error (.getMessage e)))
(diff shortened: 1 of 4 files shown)


test/ctia/task/migration/migrate_es_stores_test.clj
```diff
@@ -158,6 +158,27 @@
...
+                                       :batch-size 100,
+                                       :migration-id "migration-test",
+                                       :restart? false,
+                                       :store-keys "malware,  tool,sighting  ",
+                                       :migrations "identity",
+                                       :confirm? false,
+                                       :prefix "v1.2.0"}]
+                  (fn [] (atom {:ctia {:migration migration-props}})))]
+    (is (= (sut/get-migration-params)
+           {:buffer-size 3
```
vemv, line 173: One has to squint a bit at the code to figure out the diff between `migration-props` and the SUT output.

You could make the diff more explicity with dissoc (for comparing the parts that are `=`) and select-keys (for comparing the parts that are `not=`)

src/ctia/task/migration/migrate_es_stores.clj
```diff
@@ -291,17 +294,23 @@
                         index))))))
   true)
 
-(s/defn ^:always-validate run-migration
-  [{:keys [prefix
-           restart?]
-    :as params} :- MigrationParams]
-  (when-not restart?
-    (assert prefix "Please provide an indexname prefix for target store creation"))
+(s/defn get-migration-params :- MigrationParams
```
vemv, line 297: The naming seems a bit off, this is not a "getter" so to call it.

I'd move the `(p/get-in-global-properties [:ctia :migration])` away, to `run-migration`, making this defn a pure function.

This pure function would be easier to name, and also would not need a `with-redefs` in the tests.

src/ctia/task/migration/migrate_es_stores.clj
```diff
@@ -310,52 +319,5 @@
       (exit true)))
   (exit false))
 
-(def cli-options
```
vemv, line 313: Can you comment on why the tools.cli shell was dropped?

...Not against it, simply it'd make the pr more complete

ereteog: I updated the PR description to make more explicit the fact that this PR moves migration arguments from the command line to the properties.
Is it better?

vemv: > I updated the PR description to make more explicit the fact that this PR moves migration arguments from the command line to the properties.
Is it better?

Sure thing thanks!

vemv (approved):
lgtm!

PR #945, opened 2020-08-20
TK1.1: Seed routes with global services map argument

frenchy64:
Related threatgrid/iroh#3986

This PR creates a "services map" which is passed to routes using existing global services. Once we introduce trapperkeeper, this map will be populated only by `defservice` dependencies.

`Entity` is extended with a `:routes-from-services` key that replaces (eventually) `:routes`. This PR sets up the scaffolding to support this, but doesn't use it yet (that's for future PRs).

Please see inline comments.

I'm using TK##.## to keep track of these PR's and their order. You can ignore them, or see the planned list [here](  https://github.com/threatgrid/ctia/pull/924#actualprs). If a PR is ready for review, assume you can review in any order.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: TK1.1: Seed routes with global services map argument
```

### DIFF
--- a/src/ctia/auth/jwt.clj
+++ b/src/ctia/auth/jwt.clj
@@ -64,16 +64,16 @@
     (log/info error-msg (pr-str (into infos err)))
     (resp/unauthorized (json/generate-string err))))
 
-(defn entity-root-scope []
-  (p/get-in-global-properties [:ctia :auth :entities :scope]
+(defn entity-root-scope [get-in-config]
+  (get-in-config [:ctia :auth :entities :scope]
                               "private-intel"))
 
-(defn casebook-root-scope []
-  (p/get-in-global-properties [:ctia :auth :casebook :scope]
+(defn casebook-root-scope [get-in-config]
+  (get-in-config [:ctia :auth :casebook :scope]
                               "casebook"))
 
-(def claim-prefix
-  (p/get-in-global-properties [:ctia :http :jwt :claim-prefix]
+(defn claim-prefix [get-in-config]
+  (get-in-config [:ctia :http :jwt :claim-prefix]
                               "https://schemas.cisco.com/iroh/identity/claims"))
 
 (defn unionize
@@ -119,18 +119,18 @@
 
 (defn scope-to-capabilities
   "given a scope generate capabilities"
-  [scope]
+  [scope get-in-config]
   (let [scope-repr (scopula/to-scope-repr scope)]
     (condp = (first (:path scope-repr))
-      (entity-root-scope)   (gen-entity-capabilities scope-repr)
-      (casebook-root-scope) (gen-casebook-capabilities scope-repr)
+      (entity-root-scope get-in-config)   (gen-entity-capabilities scope-repr)
+      (casebook-root-scope get-in-config) (gen-casebook-capabilities scope-repr)
       #{})))
 
 (defn scopes-to-capabilities
   "given a seq of scopes generate a set of capabilities"
-  [scopes]
+  [scopes get-in-config]
   (->> scopes
-       (map scope-to-capabilities)
+       (map #(scope-to-capabilities % get-in-config))
        unionize))
 
 (defn iroh-claim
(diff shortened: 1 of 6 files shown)


src/ctia/http/server.clj
```diff
@@ -141,8 +143,7 @@
            refresh-url (str " " refresh-url)))
        ";"))
 
-(defn- new-jetty-instance
-  ^Server
+(s/defn ^:private ^Server new-jetty-instance
```
frenchy64, line 146: I don't think `s/defn` support type hints on the argument vector.

test/ctia/auth/jwt_test.clj
```diff
@@ -1,20 +1,24 @@
 (ns ctia.auth.jwt-test
   (:require [ctia.auth.jwt :as sut]
             [ctia.auth.capabilities :as caps]
+            [ctia.properties :as p]
             [clojure.test :as t :refer [deftest is]]
             [clojure.set :as set])
   (:import [ctia.auth.jwt JWTIdentity]))
 
+(def get-in-config
+  p/get-in-global-properties)
```
frenchy64, line 10: Above is a placeholder for creating a dummy `get-in-config` function. Since there's no fixtures in this namespace, there'll be no need to boot up a TK app, so we have to construct our own `get-in-config`.

src/ctia/auth/jwt.clj
```diff
@@ -186,19 +186,19 @@
       limit-fn)))
 
 (defn unlimited?
-  [unlimited-properties jwt]
-  (let [client-id (get jwt (iroh-claim "oauth/client/id"))
+  [unlimited-properties get-in-config jwt]
```
frenchy64, line 189: Added as second argument because of `partial` usage below.

src/ctia/http/server.clj
```diff
@@ -153,18 +154,20 @@
            send-server-version]
     :or {access-control-allow-methods "get,post,put,patch,delete"
          send-server-version false}
-    :as http-config}]
+    :as http-config}
+   {{:keys [get-in-config]} :ConfigService
+     :as services} :- APIHandlerServices]
```
frenchy64, line 159: `services` local is unused right now, but will be passed to almost all other routes in future PRs.

src/ctia/http/handler.clj
```diff
@@ -70,17 +72,29 @@
...
   <a href='/doc/README.md'>CTIA Documentation</a>")
 
+(s/defn entity->routes [entity services-map :- APIHandlerServices]
+  {:pre [entity
+         (map? services-map)]
+   :post [%]}
+  (let [{:keys [routes routes-from-services]} entity
+        _ (assert (or routes routes-from-services))
+        _ (when (and routes routes-from-services)
+            (throw (AssertionError. (str "Bad routes definition: duplicate :routes/:routes-from-services "
```
vemv, line 82: Tends to be good to use `pr-str` as one builds strings

src/ctia/schemas/core.clj
```diff
@@ -11,6 +11,26 @@
             [schema-tools.core :as st]
             [schema.core :as s :refer [Bool Str]]))
 
+(s/defschema APIHandlerServices
+  "Maps of services available to routes"
+  {:ConfigService {:get-config (s/=> s/Any s/Any)
```
vemv, line 16: I reckon that you have considered this, but given s/=> is merely informative, we might as well discard this mechanism as it leads to false confidence, and can drift vs. reality

vemv (approved):
lgtm!

src/ctia/schemas/core.clj
```diff
@@ -11,6 +11,26 @@
...
+                   :get-in-config (s/=> s/Any
+                                        [s/Any]
+                                        [s/Any s/Any])}
+   :HooksService {:apply-hooks (s/pred ifn?) ;;keyword varargs
+                  :apply-event-hooks (s/=> s/Any s/Any)}
+   :StoreService {:read-store (s/pred ifn?) ;;varags
+                  :write-store (s/pred ifn?)} ;;varags
```
ereteog, line 23: you probably need more keys here: https://github.com/threatgrid/ctia/blob/master/src/ctia/store.clj

ereteog (approved):
lgtm

src/ctia/schemas/core.clj
frenchy64: Turns out the routes only use these two.

yogsototh (approved):
LGTM!

PR #950, opened 2020-08-24
TK2.1: use services map in bulk

frenchy64:
Epic: threatgrid/iroh#3986

Note: since bundle uses bulk, there's some overlap here with https://github.com/threatgrid/ctia/pull/954/files and will require a merge.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: TK2.1: use services map in bulk
```


yogsototh (approved):
That's very welcome!

Let me also state a remark.
Unlike in IROH the init phase only provide other services access and do not generate a static configuration map. The advantage I see, is that with this we should be able to provide a dynamic `get-in-config` that could take care of `config.edn` change at runtime without the need to restart the node.

OTOH that could be risky if this not thought during other services init and it makes the code slightly more verbose. So right now, I think this PR provide this great property to have a dynamic internal state from properties that could change at runtime. It would be nice to keep that in mind during ES connection init to be able to update the ES conf/conn at runtime too.

Ops asked me about this feature. So if made possible this would be a great bonus. 

Another remark: clearly we should not provide full runtime configuration update for all the configuration values. I think it would be perfectly fine to prevent the web port to not update at runtime etc...

So anyway, that's great to see that happening!

src/ctia/bulk/core.clj
```diff
@@ -35,28 +38,39 @@
   [k]
   (get inverted-bulk-entity-mapping k))
 
-(defn create-fn
+(s/defn create-fn
   "return the create function provided an entity type key"
-  [k auth-identity params]
+  [k auth-identity params {{:keys [write-store]} :StoreService :as _services_} :- APIHandlerServices]
```
ereteog, line 43: I would remove `:as _services_`, IMO the schema annotation is a sufficient documentation
```suggestion
  [k
   auth-identity
   params
   {{:keys [write-store]} :StoreService} :- APIHandlerServices]
```

src/ctia/bulk/core.clj
```diff
@@ -35,28 +38,39 @@
...
     k store/create-record
     % (auth/ident->map auth-identity) params))
 
-(defn read-fn
+(s/defschema ReadFnServices
+  {:StoreService (-> APIHandlerServices
+                     (st/get-in [:StoreService])
+                     (st/select-keys [:read-store])
+                     (st/assoc s/Keyword s/Any))
+   s/Keyword s/Any})
```
ereteog, line 53: perhaps [st/open-schema](https://cljdoc.org/d/metosin/schema-tools/0.12.2/api/schema-tools.core#open-schema)?

src/ctia/bulk/core.clj
```diff
@@ -35,28 +38,39 @@
...
+
+(s/defn read-fn
   "return the create function provided an entity type key"
-  [k auth-identity params]
+  [k auth-identity params
+   {{:keys [read-store]} :StoreService
+    :as _services_} :- ReadFnServices]
```
ereteog, line 59: I also think that here the schema annotation is enough
```suggestion
   {{:keys [read-store]} :StoreService :- ReadFnServices]
```

src/ctia/bulk/core.clj
```diff
@@ -35,28 +38,39 @@
...
     k store/read-record
     % (auth/ident->map auth-identity) params))
 
-(defn create-entities
+(s/defn create-entities
   "Create many entities provided their type and returns a list of ids"
-  [new-entities entity-type tempids auth-identity params]
+  [new-entities entity-type tempids auth-identity params
+   {{:keys [get-in-config]} :ConfigService
+    :as services} :- APIHandlerServices]
```
ereteog, line 68:
```suggestion
   {{:keys [get-in-config]} :ConfigService} :- APIHandlerServices]
```

src/ctia/bulk/core.clj
```diff
@@ -66,10 +80,18 @@
             :data (partial map (fn [{:keys [error id] :as result}]
                                  (if error result id))))))
 
-(defn read-entities
+(s/defschema ReadEntitiesServices
+  {:StoreService (-> APIHandlerServices
+                     (st/get-in [:StoreService])
+                     (st/select-keys [:read-store])
+                     (st/assoc s/Keyword s/Any))
+   s/Keyword s/Any})
```
ereteog, line 88: [st/open-schema](https://cljdoc.org/d/metosin/schema-tools/0.12.2/api/schema-tools.core#open-schema)?

src/ctia/bundle/core.clj
```diff
@@ -97,7 +97,7 @@
 (s/defn find-by-external-ids
   [import-data entity-type auth-identity
    {{:keys [read-store]} :StoreService
-    :as services} :- APIHandlerServices]
+    :as _services_} :- APIHandlerServices]
```
ereteog, line 100: same remark here if you agree with my previous argument :-)

ereteog (approved):
LGTM!

src/ctia/bulk/core.clj
frenchy64: Thanks, but I don't like its implementation (it recursively opens all maps).

src/ctia/bulk/core.clj
frenchy64: `services` is used here actually.

PR #951, opened 2020-08-24
TK2.2: add service args to auth.jwt

frenchy64:
Epic: threatgrid/iroh#3986

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: TK2.2: add service args to auth.jwt
```

### DIFF
--- a/src/ctia/auth/jwt.clj
+++ b/src/ctia/auth/jwt.clj
@@ -64,16 +64,16 @@
     (log/info error-msg (pr-str (into infos err)))
     (resp/unauthorized (json/generate-string err))))
 
-(defn entity-root-scope []
-  (p/get-in-global-properties [:ctia :auth :entities :scope]
+(defn entity-root-scope [get-in-config]
+  (get-in-config [:ctia :auth :entities :scope]
                               "private-intel"))
 
-(defn casebook-root-scope []
-  (p/get-in-global-properties [:ctia :auth :casebook :scope]
+(defn casebook-root-scope [get-in-config]
+  (get-in-config [:ctia :auth :casebook :scope]
                               "casebook"))
 
-(def claim-prefix
-  (p/get-in-global-properties [:ctia :http :jwt :claim-prefix]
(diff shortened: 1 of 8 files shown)


src/ctia/auth/jwt.clj
```diff
@@ -186,19 +186,19 @@
       limit-fn)))
 
 (defn unlimited?
-  [unlimited-properties jwt]
-  (let [client-id (get jwt (iroh-claim "oauth/client/id"))
+  [unlimited-properties get-in-config jwt]
```
frenchy64, line 189: Added as 2nd argument because `unlimited?` is used with `partial` below.

test/ctia/auth/jwt_test.clj
```diff
@@ -1,20 +1,24 @@
 (ns ctia.auth.jwt-test
   (:require [ctia.auth.jwt :as sut]
             [ctia.auth.capabilities :as caps]
+            [ctia.properties :as p]
             [clojure.test :as t :refer [deftest is]]
             [clojure.set :as set])
   (:import [ctia.auth.jwt JWTIdentity]))
 
+(def get-in-config
+  p/get-in-global-properties)
```
frenchy64, line 10: Stub.

src/ctia/auth/jwt.clj
yogsototh: yep, that's also one of the reason I preferred to put the service-context as first argument and not the last.

And also to make it easier to support functions with variable number of args.
Fortunately I don't think we have much of those in the code.

yogsototh (approved):
LGTM!

src/ctia/auth/jwt.clj
frenchy64: I see. This isn't a context argument though, it's a service dependency. Is this a reasonable place for it?

frenchy64: Merging to keep things moving, but still interested in your opinion here.

PR #959, opened 2020-08-24
TK2.10: add services args to version routes

frenchy64:
Epic: threatgrid/iroh#3986

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: TK2.10: add services args to version routes
```

### DIFF
--- a/src/ctia/version.clj
+++ b/src/ctia/version.clj
@@ -3,8 +3,7 @@
              [io :as io]
              [shell :as shell]]
             [clojure.string :as st]
-            [ctia.domain.entities :refer [schema-version]]
-            [ctia.properties :as p]))
+            [ctia.domain.entities :refer [schema-version]]))
 
 (def version-file "ctia-version.txt")
 
@@ -14,19 +13,19 @@
               (str (:out (shell/sh "git" "log" "-n" "1" "--pretty=format:%H "))
                    (:out (shell/sh "git" "symbolic-ref" "--short" "HEAD"))))))
 
-(defn current-config-version []
-  (p/get-in-global-properties [:ctia :versions :config] ""))
+(defn current-config-version [get-in-config]
+  (get-in-config [:ctia :versions :config] ""))
 
-(defn version-data []
+(defn version-data [get-in-config]
   {:base "/ctia"
    :ctim-version schema-version
    :beta true
    :ctia-build (st/replace (current-version) #"\s\n" "")
-   :ctia-config (current-config-version)
+   :ctia-config (current-config-version get-in-config)
    :ctia-supported_features []})
 
 
-(defn version-headers []
+(defn version-headers [get-in-config]
   {"X-Ctia-Version" (st/replace (current-version) #"\n" "")
-   "X-Ctia-Config" (current-config-version)
+   "X-Ctia-Config" (current-config-version get-in-config)
    "X-Ctim-Version" schema-version})
--- a/src/ctia/http/handler.clj
+++ b/src/ctia/http/handler.clj
@@ -200,7 +200,7 @@
          (middleware [wrap-rate-limit
                       wrap-not-modified
                       wrap-cache-control
-                      wrap-version
+                      #(wrap-version % get-in-config)
                       ;; always last
                       (metrics/wrap-metrics "ctia" api-routes/get-routes)]
            documentation-routes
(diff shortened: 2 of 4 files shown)


yogsototh (approved):
LGTM!

src/ctia/version.clj
```diff
@@ -14,19 +13,19 @@
               (str (:out (shell/sh "git" "log" "-n" "1" "--pretty=format:%H "))
                    (:out (shell/sh "git" "symbolic-ref" "--short" "HEAD"))))))
 
-(defn current-config-version []
-  (p/get-in-global-properties [:ctia :versions :config] ""))
+(defn current-config-version [get-in-config]
+  (get-in-config [:ctia :versions :config] ""))
 
-(defn version-data []
+(defn version-data [get-in-config]
```
vemv, line 19: Possibly I'm stating the obvious, but it wouldn't be a particularly scalable pattern if a given function ended up receiving `[get-in-config svc-fn-a svc-fn-b]`, as it would be PLOP-y and hard to :pre.

(This was mentioned recently in an private GH thread).

I think config is special (becase it's very often used) so I'm perfectly OK with this shape. But if it's part of a pattern to be grown, I'd look into passing an `app` argument instead, which would also allow using vanilla protocols instead of injected defns, for the usual reasons.

(In the other repo there was a strong use case for using injected defns given they use a middleware pattern in the CRUD case, so refactoring those would be very arduous. Here probably there's no such pattern, and TK is being introduced from scratch, so we're on time to use TK more conservatively upfront)

vemv (approved):
Thanks for all these PRs, appreciate the splitting!

src/ctia/version.clj
frenchy64: There's lots of factors to balance while this transition is in progress. I think you're making a good point, but I'd prefer we revisit it after the dust has settled, for now I'm optimizing for code size (the TK port is `+3,282 −2,062` lines of changes, almost all breaking changes).

I'm really laser-focused on not breaking anything during this transition.

> as it would be PLOP-y

I think I know which talk you're referring to, and I think this falls under coupling rather than PLOP. https://github.com/matthiasn/talk-transcripts/blob/ccc4a017294f6b6abdf62f47aa41298f93250af6/Hickey_Rich/EffectivePrograms-mostly-text.md

> an `app` argument instead

Personally I've grown to prefer the flexibility of passing closures rather than an entire TK app or specific instances of protocols (which themselves imply an entire app was started, otherwise you'd need to stub `get-service` etc). Maybe that introduces its own kind of coupling (to use a concept I was just reminded of :) ).

I agree with your jump-to documentation concerns. I have hope that a combination of documentation strings that contain qualified names of the protocol methods represented (ie., jumpable links), and/or more precise schemas could be a good solution.

But I'll have to look into that later.

frenchy64: I thought a bit harder about this, I think I'll take your advice. One pain point I had was using `APIHandlerServices` everywhere means that unnecessarily large maps of services need to be created in the unit tests.

I'll try and use `st/select-keys` and `st/get-in` to chop up this schema into smaller pieces so it's easier to test.

frenchy64: I managed to apply this successfully to another PR, but functions that just pass around `get-in-config` are much more common, and I agree with you that it might be worth a special exception.

vemv: > There's lots of factors to balance while this transition is in progress. 

Yes definitely, my comment was aimed at the long run, which is why I hit ✅

> I think I know which talk you're referring to

I think it was in `Maybe not` where Haskell records are denounced as PLOPy because they have "slots"?

The same could be said of Clojure's `(SomeRecord. foo bar baz)`, which is why I prefer `(map->SomeRecord {...})`.

So, one can make an analogy between those slots, and the slots of a function argv in the shape `[f g h i]`... one can only add so many arguments before it degenerates into plop :)

> Personally I've grown to prefer the flexibility of passing closures rather than an entire TK app or specific instances of protocols 

Probably if used intentfully, protocols wouldn't be much different from closures - just a reified version of them :) i.e. the design is coded somewhere instead of having `foo-fn` used everywhere but defined nowhere.

But yes, there's some risk of coupling - it wouldn't be good having to create intrincate stub setups. I haven't had that experience with `with-app-with-config`

frenchy64: > I think it was in Maybe not where Haskell records are denounced as PLOPy because they have "slots"?

Yeah that sounds about right. A "place" has evolved from a mutable box in 2017 to "something RH doesn't like" in 2019. :)

PR #963, opened 2020-08-24
TK2.14: add services args to auth middleware

frenchy64:
Epic: threatgrid/iroh#3986

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: TK2.14: add services args to auth middleware
```

### DIFF
--- a/src/ctia/http/middleware/auth.clj
+++ b/src/ctia/http/middleware/auth.clj
@@ -1,5 +1,6 @@
 (ns ctia.http.middleware.auth
-  (:require [ctia.auth :as auth :refer [auth-service]]
+  (:require [ctia.auth :as auth]
+            [ctia.schemas.core :refer [APIHandlerServices]]
             [compojure.api.meta :as meta]
             [ring.util.http-response :as http-response]
             [schema.core :as s]))
@@ -15,22 +16,23 @@
         (assoc-in [:headers "authorization"] auth-header))
     request))
 
-(defn testable-wrap-authentication
+(s/defn testable-wrap-authentication
   "wrap-autentication middleware."
-  [handler auth-service]
+  [handler
+   identity-for-token :- (s/=> s/Any s/Any)]
   (fn [request]
     (handler
      (if (:login request)
        request
        (let [auth-header (or (get-in request [:headers "authorization"])
                              (get-in request [:query-params "Authorization"]))
-             id (auth/identity-for-token auth-service auth-header)
+             id (identity-for-token auth-header)
              login (auth/login id)
              groups (auth/groups id)]
          (add-id-to-request request id login groups auth-header))))))
 
-(defn wrap-authentication [handler]
-  (testable-wrap-authentication handler @auth-service))
+(s/defn wrap-authentication [handler identity-for-token :- (s/=> s/Any s/Any)]
+  (testable-wrap-authentication handler identity-for-token))
 
 (defn wrap-authenticated [handler]
   (fn [request]
--- a/src/ctia/http/server.clj
+++ b/src/ctia/http/server.clj
@@ -155,12 +155,13 @@
     :or {access-control-allow-methods "get,post,put,patch,delete"
          send-server-version false}
     :as http-config}
-   {{:keys [get-in-config]} :ConfigService
+   {{:keys [identity-for-token]} :IAuth
+    {:keys [get-in-config]} :ConfigService
      :as services} :- APIHandlerServices]
   (doto
       (jetty/run-jetty
        (cond-> (handler/api-handler services)
-         true auth/wrap-authentication
+         true (auth/wrap-authentication identity-for-token)
 
          (:enabled jwt)
          auth-jwt/wrap-jwt-to-ctia-auth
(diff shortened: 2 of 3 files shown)


src/ctia/http/middleware/auth.clj
```diff
@@ -15,22 +16,23 @@
         (assoc-in [:headers "authorization"] auth-header))
     request))
 
-(defn testable-wrap-authentication
+(s/defn testable-wrap-authentication
   "wrap-autentication middleware."
-  [handler auth-service]
+  [handler
+   identity-for-token :- (s/=> s/Any s/Any)]
```
frenchy64, line 22: Changes second arg from `auth-service` to `identity-for-token`.

vemv (approved):

src/ctia/http/middleware/auth.clj
ereteog, line 22: > "I learnt a few days ago that s/=> has no runtime semantics
very sad
at least according to the docs
it's actually just an ifn? check haha"

[Ambrose  2020] :-)
```suggestion
   identity-for-token :- (s/pred ifn?)]
```

src/ctia/http/middleware/auth.clj
```diff
@@ -15,22 +16,23 @@
...
                              (get-in request [:query-params "Authorization"]))
-             id (auth/identity-for-token auth-service auth-header)
+             id (identity-for-token auth-header)
              login (auth/login id)
              groups (auth/groups id)]
          (add-id-to-request request id login groups auth-header))))))
 
-(defn wrap-authentication [handler]
-  (testable-wrap-authentication handler @auth-service))
+(s/defn wrap-authentication [handler identity-for-token :- (s/=> s/Any s/Any)]
```
ereteog, line 34:
```suggestion
(s/defn wrap-authentication [handler identity-for-token :- (s/pred ifn?)]
```

ereteog (approved):
lgtm

src/ctia/http/middleware/auth.clj
frenchy64: I was just stating a fact that seemed to be misunderstood by the team, I was surprised to see this interpreted as "use ifn? over s/=>". I feel the opposite :)

src/ctia/http/middleware/auth.clj
frenchy64: I'm probably not going to delete the `s/=>`s if the only reason is the semantics.

There are two `identity-for-token` functions floating about, one protocol, one service dependency. IMO deleting this one loses documentation about which one we're using.

Using `s/=>` has been keeping me sane during this port since 80% of the changes are introducing HOF's!

vemv: > I'm probably not going to delete the s/=>s if the only reason is the semantics.

I guess the reason is that they promise something that isn't delivered, and that can get out of sync. That can give us false confidence, leading us to confusion or mistakes as time passes.

Personally I'm OK with the => as a temporary tool for refactoring, or if there's a reasonable chance we'll get our own `=>` which actually performs checking.

frenchy64: > I guess the reason is that they promise something that isn't delivered

(Warning: musing + opinions)

Fair enough. From my perspective, there are many ways of implementing `s/=>`, and to learn a contracts system is to learn which they choose. For example, with spec, you should know the differences between the semantics of `s/fspec` and `s/fdef` (or ignore them at your peril). Similarly, there's an analogous difference between `s/=>` and `s/defn` (technically the [docstrings](https://github.com/plumatic/schema/blob/ddb54c87dea6926c6d73542e517b53685cc82a37/src/cljx/schema/core.cljx#L1086-L1087) warn about these semantics, but obviously we all missed them).

In spec, the interesting semantics of `s/fspec` prevents me from using it in some cases, and overrides the usefulness of its documentation. OTOH `s/=>` has a conservative implementation that you could easily imagine in the future as being enhanced to "do the right thing" or rolling our own (as you say), so it's a clearer win to me in some cases.

My intention of pointing out the semantics of `s/=>` was to help the team better understand schema, which we certainly do now. I didn't anticipate a purge of `s/=>` definitions, but that's a side-effect of the better understanding, so I think my mission is accomplished (I thought it might stimulate interest in adding semantics to `s/=>` if anything).

A somewhat relevant note: I've found it useful to combine `s/=>` with `s/fn` in the body of functions to be explicit about both the interface of a function and its checking semantics. eg.,

```clojure
(s/defn might-return-a-fn :- (s/maybe (s/=> s/Int s/Int))
  []
  (s/fn :- s/Int [i :- s/Int] (+ 1 i)))
```

vemv: Honestly I have a hard time seeing your POV, after a few attempts 😃

Would you maintain the same opinion if `=>` checked `some?` instead of `ifn?`? Or if it did no checking at all?

From the [docstring](https://github.com/plumatic/schema/blob/ddb54c87dea6926c6d73542e517b53685cc82a37/src/cljx/schema/core.cljx#L1072), which uses the word `currently` I perceive a hint that there wasn't a strong philosophy behind its current implementation; a stronger implementation surely was left in a wishlist that never materialised.

ereteog: In any case, I value the documentation aspect of `=>`.

PR #965, opened 2020-08-25
Target Record entity

agzam:
> A `TargetRecord` is a `Sighting` that has no threat or observables associated with it, it's a way of saying they saw a set of observables together as a `Target`. This is useful when attempting to determine what assets were targeted by a threat when we only have weak transient observables, like IPs or dynamically assigned hostnames.

Notes:
-----

This is a continuation of https://github.com/threatgrid/ctia/pull/939

<a name="qa">[§](#qa)</a> QA
============================

TargetRecord CRUD routes should work.

- You can create TargetRecord(s) using CTIM model example, e.g.:

        {
          "tlp": "green",
          "language": "language",
          "source_uri": "http://example.com/target-record",
          "external_references": [
              {
                  "description": "Description text",
                  "hashes": [
                      "#section1"
                  ],
                  "url": "https://ex.tld/wiki/TR042",
                  "external_id": "TR042",
                  "source_name": "source"
              }
          ],
          "title": "TR-8BA72C1B",
          "short_description": "short description",
          "targets": [
              {
                  "sensor": "process.dhcp-server",
                  "source_uri": "https://console.amp.cisco.com/computers/5229eaaa-a3f1-4a05-b961-65c8b9a28e96/trajectory?q=192.168.243.112",
                  "internal": true,
                  "observables": [
                      {
                          "type": "hostname",
                          "value": "atl.ciscothreatresponse.local"
                      },
                      {
                          "type": "ip",
                          "value": "192.168.1.204"
                      },
                      {
                          "type": "mac_address",
                          "value": "00:50:56:b8:0c:c8"
                      }
                  ],
                  "observed_time": {
                      "end_time": "2525-01-01T00:00:00Z",
                      "start_time": "2020-01-11T00:40:48Z"
                  },
                  "os": "centos linux release 7.5",
                  "type": "endpoint"
              }
          ],
          "external_ids": [
              "http://ex.tld/ctia/target-record/target-record-7d3e0f0c-6ee7-4578-8e1a-f45a750bb37b"
          ],
          "source": "cisco:ise:dhcp-server",
          "type": "target-record",
          "revision": 1,
          "schema_version": "c/ctim-schema-version",
          "description": "description"
      }

- When testing the search feature, make sure you try searching by `target-record.targets.type`,  `target-record.targets.observables.type`.


test/ctia/entity/target_record_test.clj
```diff
@@ -0,0 +1,101 @@
...
+(use-fixtures :once (join-fixtures [mth/fixture-schema-validation
+                                    helpers/fixture-properties:clean
+                                    whoami-helpers/fixture-server]))
+
+(use-fixtures :each whoami-helpers/fixture-reset-state)
+
+(defn additional-tests [_ target-record-sample]
+  (testing "GET /ctia/target-record/search"
+    ;; only when ES store
+    (when (= "es" (p/get-in-global-properties [:ctia :store :target-record]))
```
vemv, line 27: in tests, it tends to better to use when-not to when, so that one defaults to running the tests instead of skipping them

Else imagine the `[:ctia :store :target-record]` path was refactored

test/ctia/entity/target_record_test.clj
```diff
@@ -0,0 +1,101 @@
...
+      (are [term check-fn expected desc] (let [response (helpers/get
+                                                         "ctia/target-record/search"
+                                                         :query-params {"query" term}
+                                                         :headers {"Authorization" "45c1f5e3f05d0"})]
+                                           (is (= 200 (:status response)))
+                                           (is (= expected (check-fn response)) desc))
```
vemv, line 33: You can add a `true` at tail position to prevent `are` from double-failing

test/ctia/test_helpers/aggregate.clj
```diff
@@ -45,25 +46,28 @@
...
   [m ks]
   (reduce (fn [acc k]
             (cond
-              (map? acc) (get acc k)
-              (coll? acc) (seq (keep #(get % k) acc))
-              :else (reduced acc)))
+              (map? acc)  (get acc k)
+              (coll? acc) (->> acc
+                               (keep #(es-get-in % [k]))
+                               flatten
```
vemv, line 68: I tend to prefer `apply concat` given:

```
user=> (flatten #{1 2 3})
()
```

test/ctia/test_helpers/aggregate.clj
```diff
@@ -219,10 +223,15 @@
   (let [enumerable-schema (schema-enumerable-fields new-schema enumerable-fields)
         base-doc (dissoc entity-minimal :id)]
     (doall
-     (repeatedly n (fn [] (merge base-doc
-                                 (g/generate enumerable-schema
-                                             {s/Str string-generator})
-                                 (generate-date-fields date-fields)))))))
+     (repeatedly n (fn [] (deep-merge-with
+                          (fn [a b]
+                            (if (seq? a)
```
vemv, line 228: Not necessarily wrong (hard for me to tell without diving into the code), but `seq?` sometimes has an intent that would be more explicit with some other pred (`seq`, `coll?`, `list?`, `sequential?`)

test/ctia/test_helpers/aggregate.clj
agzam: but `(apply concat #{1 2 3})` is even worse, it throws `IllegalArgumentException`, I think I'm missing something here.

test/ctia/entity/target_record_test.clj
agzam: I agree with you. I adopted this snippet from previous work without much thinking; tests for `identity_assertion` and `judgement` also use the same pattern. I'm not sure if I can remove this `when` altogether. @ereteog probably can give more insight on that.

test/ctia/test_helpers/aggregate.clj
vemv: A better example:

```clj
dev=> (flatten #{[1] [2] [3]})
()
```

---

There's also the problem that `flatten` is recursive:

```clj
dev=> (apply concat [[1], [2], [[[[[[[[3]]]]]]]]])
(1 2 [[[[[[[3]]]]]]])
dev=> (flatten [[1], [2], [[[[[[[[3]]]]]]]]])
(1 2 3)
```

...99% of the times one wants to flatten just one level. Most of the time`flatten` happens to achieve the same effect (because the coll had depth 1), but if in a future the coll depth changed, one would get undesirable behavior opaquely.

test/ctia/entity/target_record_test.clj
ereteog: @vemv how would you rewrite that case? because here we only want it to run when the store is "es".

test/ctia/test_helpers/aggregate.clj
ereteog: yes here we only want to flatten one depth level. I recommend to add the attended behavior to the tests.
BTW, here it becomes complicate, I wanted a somehow generic way to test metrics, but it's really hard to simulate ES behavior. We probably hitting some limits. 
So if it becomes too hard for this PR to use the test helper, feel free to not use it and write more verbose tests.

test/ctia/entity/target_record_test.clj
vemv: > how would you rewrite that case? because here we only want it to run when the store is "es".

```clj
(when-not (#{"foo" "bar" "baz"} (p/get-in-global-properties [:ctia :store :target-record]))
  ...)
```

...of course this set could be `def`ed somewhere.

test/ctia/test_helpers/aggregate.clj
agzam: @vemv Damn, you got me to scratch my head really hard :) So, I switched from `flatten` to `apply concat` in this particular case, and it broke the tests, it took me a while to sort it out exactly why. The reason it broke the tests, apparently, sometimes the retrieved collection is already flat, and `flatten` doesn't do anything, see this:

```
(flatten ["foo" "bar"] 
;; => ["foo" "bar"])

(apply concat ["foo" "bar"]) 
;; => (\f \o \o \b \a \r)
```
It seems using `flatten` here is justified.

@ereteog 
> BTW, here it becomes complicate, I wanted a somehow generic way to test metrics, but it's really hard to simulate ES behavior.

I wanted to avoid larger refactors/restructuring at this point. I'm still getting familiar with the problem space and the codebase.

test/ctia/entity/target_record_test.clj
```diff
@@ -0,0 +1,104 @@
...
+
+        "*"
+        (fn [r] (-> r :parsed-body first :targets first :observables first :value))
+        (-> target-record-sample :targets first :observables first :value)
+        "Searching for target-record:targets:observables:value"
+
+        "*"
+        (fn [r] (-> r :parsed-body first :targets first :observables first :type))
+        (-> target-record-sample :targets first :observables first :type)
+        "Searching for target-record:targets:observables:type"))))
```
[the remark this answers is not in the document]
agzam: Yeah, I don't know what I was thinking fixed in 6a3f371154e9c3276b228ba794b4a9c27c14caab

test/ctia/entity/target_record_test.clj
agzam: I don't like this either tbh. Let me ship it as is for now. I promise you to try to make it nicer later.

test/ctia/test_helpers/aggregate.clj
```diff
@@ -149,19 +154,29 @@
...
+                  (keep #(es-get-in % parsed))
+                  flatten)
           from-str "2020-03-01T00:00:00.000Z"
           to-str "2020-10-01T00:00:00.000Z"
-          from (tf/parse from-str)
-          to (tf/parse to-str)
-          date-values (filter #(and (time/within? from to %)
-                                    (time/before? % to))
-                              (map tf/parse values))
+          date-values (vals->date-vals from-str to-str values)
```
ereteog, line 179: Since this test helper is tricky, I would add check that `date-values` is not empty:
``` clojure
(assert (ses date-values))
```

ereteog (approved):
Great PR!
Thank you very much for improving our test helpers!

ereteog:
Just one detail:
You need to add something in the QA section. This had to be tested, I would recommend to ask applying the standard CRUD, search and metric tests that are applied for other entities.

PR #976, opened 2020-09-01
TK3.8: change config "before" app starts in access-control-test

frenchy64:
Related: https://github.com/threatgrid/iroh/issues/3986

As it stands, CTIA's tests freely interleave "starting" CTIA (via fixtures) and changing its global configuration (via `p/global-properties-atom`).

With Trapperkeeper, configuration may only be changed at `app` startup, so as we move to TK some changes are needed to delay startup until all configuration changes are "in".

`access-control-test` expects to be run after CTIA has started, but it also changes the global configuration. This PR refactors it into a HOF taking fixtures so it can manage starting CTIA itself, and then adds configuration changes _before_ calling the fixtures that starts CTIA.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: TK3.8: access-control-test => HOF, add config transformers
```

### DIFF
--- a/test/ctia/test_helpers/access_control.clj
+++ b/test/ctia/test_helpers/access_control.clj
@@ -951,86 +951,110 @@
                                                   (:parsed-body player-3-entity-repost2)]})))))))
 
 (defn test-access-control-tlp-settings
-  [entity new-entity]
-  (testing "TLP Settings Enforcement"
-    (swap! (p/global-properties-atom) assoc-in
-           [:ctia :access-control]
-           {:default-tlp "amber"
-            :min-tlp "amber"})
+  [entity new-entity fixtures]
+  (helpers/with-config-transformer*
+    #(assoc-in % [:ctia :access-control] {:default-tlp "amber"
+                                          :min-tlp "amber"})
+    #(fixtures
+       (fn []
+         (testing "TLP Settings Enforcement"
+           (let [;; verify the with-config-transformer* call above--it's possible
+                 ;; a fixture could override it since we call it first
+                 get-in-config (helpers/current-get-in-config-fn)
+                 _ (assert (= (get-in-config [:ctia :access-control])
+                              {:default-tlp "amber"
+                               :min-tlp "amber"})
+                           (get-in-config [:ctia :access-control]))
+                 
+                 {status-default-tlp :status
+                  body-default-tlp :parsed-body}
+                 (post (format "ctia/%s" entity)
+                       :body (dissoc new-entity :tlp)
+                       :headers {"Authorization" "player-1-token"})
+                 {status-disallowed-tlp :status
+                  body-disallowed-tlp :parsed-body}
+                 (post (format "ctia/%s" entity)
+                       :body (assoc new-entity :tlp "white")
+                       :headers {"Authorization" "player-1-token"})]
 
-    (let [{status-default-tlp :status
-           body-default-tlp :parsed-body}
-          (post (format "ctia/%s" entity)
-                :body (dissoc new-entity :tlp)
-                :headers {"Authorization" "player-1-token"})
-          {status-disallowed-tlp :status
-           body-disallowed-tlp :parsed-body}
-          (post (format "ctia/%s" entity)
-                :body (assoc new-entity :tlp "white")
-                :headers {"Authorization" "player-1-token"})]
-
-      (is (= 201 status-default-tlp))
-      (is (= "amber" (:tlp body-default-tlp)))
+             (is (= 201 status-default-tlp))
+             (is (= "amber" (:tlp body-default-tlp)))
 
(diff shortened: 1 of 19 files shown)


test/ctia/test_helpers/core.clj
```diff
@@ -28,6 +28,17 @@
...
+(defn with-config-transformer*
+  [tf body-fn]
+  ;; TODO stub, assumed single-threaded
+  (let [a (p/global-properties-atom)
+        old @a]
+    (try
+      (swap! a tf)
+      (body-fn)
+      (finally
+        (reset! a old)))))
```
frenchy64, line 40: Will be completely rewritten when TK is introduced.

test/ctia/test_helpers/access_control.clj
```diff
@@ -951,86 +951,110 @@
...
+    #(fixtures
+       (fn []
+         (testing "TLP Settings Enforcement"
+           (let [;; verify the with-config-transformer* call above--it's possible
+                 ;; a fixture could override it since we call it first
+                 get-in-config (helpers/current-get-in-config-fn)
+                 _ (assert (= (get-in-config [:ctia :access-control])
+                              {:default-tlp "amber"
+                               :min-tlp "amber"})
+                           (get-in-config [:ctia :access-control]))
```
frenchy64, line 967: There is an unfortunate ordering issue here. Ideally, we'd call `with-config-transformer*` inside the function we pass to `fixtures`, but `fixtures` starts CTIA, so it's "too late" to do that (for reasons described in PR description).

To workaround this, we double check our `with-config-transformer*` call has not been overwritten by another fixture via this assertion.

test/ctia/test_helpers/access_control.clj
```diff
@@ -951,86 +951,110 @@
...
+
+             (is (= 400 status-disallowed-tlp))
+             (is (= "Invalid document TLP white, allowed TLPs are: amber,red"
+                    (:message body-disallowed-tlp)))))))))
+
+;; the body of this function must change the current TK config
+;; via test-access-control-tlp-settings. This is only possible by starting a new app.
+;; Since apps are started by fixtures, this function accepts a fixture and 
+;; must not be called in a context wrapped by fixture-ctia (this is dynamically enforced).
 (defn access-control-test
```
frenchy64, line 992: The changes below are mostly a reshuffling of existing code.

We redefined `fixtures` to include the setup of capabilities and such, then we separate `test-access-control-tlp-settings` from the rest of the tests because it needs to set the config before starting CTIA.

test/ctia/test_helpers/access_control.clj
```diff
@@ -951,86 +951,110 @@
...
-
-      (is (= 400 status-disallowed-tlp))
-      (is (= "Invalid document TLP white, allowed TLPs are: amber,red"
-             (:message body-disallowed-tlp))))))
-
+  [entity new-entity fixtures]
+  (helpers/with-config-transformer*
+    #(assoc-in % [:ctia :access-control] {:default-tlp "amber"
+                                          :min-tlp "amber"})
+    #(fixtures
```
vemv, line 958: wdyt of https://stuartsierra.com/2019/09/11/clojure-donts-short-fn-syntax ?

I'm aware we haven't discussed it, but for large cases like a 30-line `fixtures`, I'd say a threshold has definitely been surpassed. A large `#()` makes me visually hunt for `%` which may or not be inside the function body - awkward for a function so large.

test/ctia/test_helpers/access_control.clj
```diff
@@ -951,86 +951,110 @@
...
+                                  (whoami-helpers/set-whoami-response "player-3-token"
+                                                                      "player3"
+                                                                      "bargroup"
+                                                                      "user"))]
+                     (fixtures
+                       (fn []
+                         (setup!)
+                         (f)))))
+        
+        test-default-config #(do (test-access-control-entity-tlp-green
```
vemv, line 1030: Same

vemv (approved):
Great work! Appreciate the explanation.

test/ctia/test_helpers/access_control.clj
frenchy64: > A large #() makes me visually hunt for % which may or not be inside the function body

Good point. You also can't use a nested `#()` in those 30 lines.

> wdyt of https://stuartsierra.com/2019/09/11/clojure-donts-short-fn-syntax ?

Thanks for pointing me to it. I like the clarity of explaining the expansion of the `#()` reader macro. Personally don't really have much good to say about the rest of the post, as Stuart says in the first sentence, it seems written to those in their first 3 days of Clojure.

frenchy64: Good catch.

Issue #981, opened 2020-09-02
CRUD API routes for Asset Mapping entity


agzam:
We need basic (default) API routes so we can create, retrieve, search, and delete for AssetMapping entities.

PR #983, opened 2020-09-02
Asset Mapping  expire api

### LINKED ISSUE #977: Asset Mapping Expire API route(s)
We should be able to "expire" an Asset Mapping by setting its `:valid_time` field interval to end at the current date-time.

agzam:
> Close #977 
> Related #981 

<a name="qa">[§](#qa)</a> QA
============================

1. Create an AssetMapping

  Make sure `valid_time.end_time` is sometime in the future.

        POST http://localhost:3000/ctia/asset-mapping
        Content-type: application/json

        {
          "type": "asset-mapping",
          "asset_type": "device",
          "valid_time": {
            "start_time": "2020-01-11T00:40:48Z",
            "end_time": "2525-01-01T00:00:00Z"
          },
          "stability": "Managed",
          "schema_version": "1.0",
          "revision": 1,
          "observable": {
            "type": "ip",
            "value": "100.213.110.122"
          },
          "asset_ref": "http://localhost:3000/ctia/asset/asset-3843650f-5769-4887-a170-cb3329245321",
          "source": "cisco:unified_connect",
          "external_ids": ["8b48f2bd-b634-45b5-80a9-a09b6e78ec06"],
          "external_references": [
            {
              "source_name": "source",
              "external_id": "T1061",
              "url": "https://ex.tld/wiki/T1061",
              "hashes": [
                "#section1"
              ],
              "description": "Description text"
            }
          ],
          "source_uri": "http://example.com/asset-mapping",
          "language": "language",
          "tlp": "green",
          "specificity": "Unique",
          "confidence": "High"
        }
3. The endpoint returns an entity. Copy its "short id", e.g. `asset-mapping-a3cc8d78-8f53-477e-92b9-16a3006e5cb2`
2. Call expire endpoint using that id:

        PATCH http://localhost:3000/ctia/asset-mapping/expire/asset-mapping-a3cc8d78-8f53-477e-92b9-16a3006e5cb2

3. Read the same entity:

        GET http://localhost:3000/ctia/asset-mapping/asset-mapping-a3cc8d78-8f53-477e-92b9-16a3006e5cb2
        
4. Check `valid_time.end_time`, the value should be the date-time when it was patched

```
```

### DIFF
--- a/src/ctia/entity/asset_mapping.clj
+++ b/src/ctia/entity/asset_mapping.clj
@@ -119,19 +119,18 @@
      (if-let [updated
               (flows/patch-flow
                :services services
-               :get-fn (fn [_]
-                         (read-store :asset-mapping
-                                     ctia.store/read-record
-                                     id
-                                     identity-map
-                                     {}))
+               :get-fn (fn [_] (read-store :asset-mapping
+                                           ctia.store/read-record
+                                           id
+                                           identity-map
+                                           {}))
                :realize-fn realize-asset-mapping
                :update-fn #(write-store :asset-mapping
-                                       ctia.store/update-record
-                                       (:id %)
-                                       (assoc-in % [:valid_time :end_time] (time/internal-now))
-                                       identity-map
-                                       {})
+                                        ctia.store/update-record
+                                        (:id %)
+                                        (assoc-in % [:valid_time :end_time] (time/internal-now))
+                                        identity-map
+                                        {})
                :long-id-fn #(entities/with-long-id % get-in-config)
                :entity-type :asset-mapping
                :entity-id id
--- a/src/ctia/entity/asset_mapping.clj
+++ b/src/ctia/entity/asset_mapping.clj
@@ -1,15 +1,17 @@
 (ns ctia.entity.asset-mapping
-  (:require [ctia.domain.entities :refer [default-realize-fn]]
-            [ctia.schemas.core :refer [def-acl-schema def-stored-schema]]
-            [ctia.schemas.utils :as csu]
+  (:require [compojure.api.sweet :refer [POST routes]]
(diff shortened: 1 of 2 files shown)


ereteog:
The implementation lgtm.
However, do we really need this route, or just turning on the PATCH route with only the editable properties like `valid_time.end_time`? BTW, the update route enables us to fully erase an assert-mapping, so you should perhaps disable it if most of the fields should be immutable.

src/ctia/entity/asset_mapping.clj
```diff
@@ -97,29 +105,67 @@
...
-    :external-id-capabilities :read-asset-mapping
-    :can-aggregate?           true
-    :histogram-fields         asset-mapping-histogram-fields
-    :enumerable-fields        asset-mapping-enumerable-fields
-    }))
+(s/defn additional-routes [{{:keys [get-in-config]}          :ConfigService
+                            {:keys [read-store write-store]} :StoreService
+                            :as                              services} :- APIHandlerServices]
+  (routes
+   (PATCH "/expire/:id" []
```
gbuisson, line 112: IMHO using the `PATCH` verb here is misleading, in the REST world, `PATCH` is generally used by an API customer to supply a partial update for an entity, typically for such operations which are outside of the REST scope it is preferable to use `POST`.

ereteog:
> The implementation lgtm.
> However, do we really need this route, or just turning on the PATCH route with only the editable properties like `valid_time.end_time`? BTW, the update route enables us to fully erase an assert-mapping, so you should perhaps disable it if most of the fields should be immutable.

I synced with @gbuisson, you can forget this comment we are OK with this new route. You just need to change the http method.

ereteog (approved):
LGTM

PR #968, opened 2020-08-31
CTIA: independant ES target properties

ereteog:
> Related #threatgrid/iroh#3072

This PR introduces the ability to migrate stores across different clusters and more flexibility for the target store configuration:
- The base change is to enable configuring dedicated properties for target stores instead of building them from source stores.
- It also get rid of the full source initialization and use only necessary pure init function that do not update templates and mappings.
- It put back the command arguments for `confirm` and `restart` params which is more handy to use for OPS. In particular it enable OPS to restart a migration with the same properties file.
- Updates the doc with the full new process.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

PR #984, opened 2020-09-04
TK3.9: add service args to ESConnState

frenchy64:
**Epic** https://github.com/threatgrid/iroh/issues/3986

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: TK3.9: add service args to ESConnState
```

### DIFF
--- a/src/ctia/schemas/search_agg.clj
+++ b/src/ctia/schemas/search_agg.clj
@@ -6,22 +6,15 @@
   "Date range query, includes lowerfrom and excludes to"
   {s/Keyword
    (st/optional-keys
-    {:gte (s/conditional
-            string? s/Str
-            :else s/Inst)
-     :lt (s/conditional
-           string? s/Str
-           :else s/Inst)})})
+    {:gte s/Inst
+     :lt s/Inst})})
 
 (s/defschema SearchQuery
   "components of a search query:
    - query-string: free text search, with lucene syntax enabled"
   (st/optional-keys
    {:query-string s/Str
-    :filter-map {(s/conditional
-                   string? s/Str
-                   :else s/Keyword)
-                 s/Any}
+    :filter-map {s/Keyword s/Any}
     :date-range DateRange}))
 
 (s/defschema AggType
--- a/src/ctia/schemas/search_agg.clj
+++ b/src/ctia/schemas/search_agg.clj
@@ -6,15 +6,22 @@
   "Date range query, includes lowerfrom and excludes to"
   {s/Keyword
    (st/optional-keys
-    {:gte s/Inst
-     :lt s/Inst})})
+    {:gte (s/conditional
+            string? s/Str
+            :else s/Inst)
+     :lt (s/conditional
+           string? s/Str
+           :else s/Inst)})})
 
 (s/defschema SearchQuery
   "components of a search query:
    - query-string: free text search, with lucene syntax enabled"
   (st/optional-keys
    {:query-string s/Str
-    :filter-map {s/Keyword s/Any}
+    :filter-map {(s/conditional
+                   string? s/Str
+                   :else s/Keyword)
+                 s/Any}
     :date-range DateRange}))
 
 (s/defschema AggType
--- a/src/ctia/task/migration/store.clj
+++ b/src/ctia/task/migration/store.clj
@@ -9,7 +9,7 @@
             [clojure.set :as set]
             [clojure.string :as string]
             [clojure.tools.logging :as log]
-            [ctia.init :refer [log-properties]]
+            [ctia.init :refer [init-store-service! log-properties]]
             [ctia.lib.collection :refer [fmap]]
             [ctia.properties :refer [init!]]
             [ctia.store :refer [stores]]
@@ -27,7 +27,6 @@
 
 (def timeout (* 5 60000))
 (def es-max-retry 3)
-;;FIXME refactor to local argument
 (defonce migration-es-conn (atom nil))
 
 (s/defschema MigrationStoreServices
@@ -59,7 +58,8 @@
                    :started em/ts
                    :completed em/ts}}})
 
-(s/defn migration-mapping [{{:keys [deref-stores]} :StoreService} :- MigrationStoreServices]
+(s/defn migration-mapping
+  [{{:keys [deref-stores]} :StoreService} :- MigrationStoreServices]
   {"migration"
    {:dynamic false
     :properties
@@ -681,6 +681,13 @@ when confirm? is true, it stores this state and creates the target indices."
 (s/defn setup!
   "setup store service"
   [services :- MigrationStoreServices]
+  ;; TODO will be rolled into trapperkeeper bootstrap
+  ;; START global services setup
+  (log/info "starting CTIA Stores...")
+  (init!)
+  (log-properties)
+  (init-store-service!)
+  ;; END global services setup
   (reset! migration-es-conn
           (-> (migration-store-properties services)
               (init-store-conn (MigrationStoreServices->ESConnServices
--- a/src/ctia/task/migration/store.clj
+++ b/src/ctia/task/migration/store.clj
@@ -5,35 +5,62 @@
             [clj-momo.lib.es.document :as es-doc]
             [clj-momo.lib.es.index :as es-index]
             [clj-momo.lib.es.query :as es-query]
-            [clj-momo.lib.es.schemas :refer [ESConn ESQuery]]
+            [clj-momo.lib.es.schemas :refer [ESConn ESQuery Refresh]]
             [clojure.set :as set]
             [clojure.string :as string]
             [clojure.tools.logging :as log]
-            [ctia.init :refer [init-store-service! log-properties]]
+            [ctia.init :refer [log-properties]]
             [ctia.lib.collection :refer [fmap]]
-            [ctia.properties :refer [init!]]
-            [ctia.store :refer [stores]]
+            [ctia.properties :as p]
+            [ctia.store :as store]
             [ctia.stores.es.crud :as crud]
-            [ctia.stores.es.init
-             :refer
-             [get-store-properties init-es-conn! init-store-conn StoreProperties]]
+            [ctia.stores.es.init :as es.init]
             [ctia.stores.es.mapping :as em]
             [ctia.stores.es.schemas :refer [ESConnServices ESConnState]]
             [ctia.stores.es.store :as es-store :refer [StoreMap]]
             [ctia.task.rollover :refer [rollover-store]]
+            [ctia.task.migration.migrations :refer [available-migrations]]
             [ctim.domain.id :refer [long-id->id]]
             [schema-tools.core :as st]
             [schema.core :as s]))
 
+(s/defschema ESStoreProperties
+  (st/optional-keys
+   {:host s/Str
+    :port s/Int
+    :transport s/Keyword
+    :indexname s/Str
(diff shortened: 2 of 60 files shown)


src/ctia/task/migration/store.clj
```diff
@@ -634,16 +675,21 @@ when confirm? is true, it stores this state and creates the target indices."
...
-  "init properties, start CTIA and its store service"
-  []
+                          es-conn
+                          services))
+
+(s/defn setup!
+  "setup store service"
+  [services :- MigrationStoreServices]
+  ;; TODO will be rolled into trapperkeeper bootstrap
+  ;; START global services setup
```
ereteog, line 685: So the migration will only use the configuration services, and should now avoid triggering `init--store-services!` that updates source store. This is handy when starting CTIA but should be avoided (and could fail) for a migration.
https://github.com/threatgrid/ctia/pull/968/files#diff-0e5ad98d51479717941d6ba505d26839L640

test/ctia/task/migration/migrate_es_stores_test.clj
```diff
@@ -74,10 +71,25 @@
...
+                                 es-helpers/fixture-delete-store-indexes
+                                 fixture-clean-migration])]
+    (helpers/with-config-transformer*
+      config-transformer
+      #(fixtures body-fn))))
+
+(defmacro with-each-fixtures
+  "Primarily for check-migration-params-test."
+  [config-transformer & body]
+  `(with-each-fixtures* ~config-transformer #(do ~@body)))
```
ereteog, line 92: It's a nice clarification of the context of execution of each test. I am wondering if we should not go further and explicitly pass the configuration instead of a transformer in order to clarify the context of execution of a migration test (and the lack of clarity of this context is totally my bad here ;-) )

ereteog (commented):
An early review, the PR title dragged me into it :-)

src/ctia/task/migration/store.clj
```diff
@@ -25,7 +27,19 @@
...
 (def es-max-retry 3)
-(def migration-es-conn (atom nil))
+(defonce migration-es-conn (atom nil))
+
+(s/defschema MigrationStoreServices
+  {:ConfigService {:get-in-config (s/=>* s/Any
+                                         [(s/named s/Any 'path)]
+                                         [(s/named s/Any 'path)
+                                          (s/named s/Any 'default)])}
+   :StoreService {:deref-stores (s/=> s/Any)}})
```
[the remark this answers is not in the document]
frenchy64: > I would hide the atom detail here, and just use something like get-store, wdyt?

Yes, I had the same thought. Since you're interested, I'll see how feasible it is. I was worried that enough places used `stores` with deref/reset!/swap! for it to be too much work to abstract over, but I never checked.

> FYI I am currently removing the use of stores atom in this PR: #968. It's initialization triggers template and mapping updates in the source, which is unnecessary (and not desirable) during the migration.

Ah ok. I'll keep that in mind, might be able to avoid the merge conflict.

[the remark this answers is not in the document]
frenchy64: It was just a matter of searching for `stores-atom` in the big PR. `reset!` is only used to initialize the stores. Should be easy to refactor.

test/ctia/task/migration/migrate_es_stores_test.clj
frenchy64: It's kind of difficult because CTIA is usually started by a test fixture. That's what motivated this strange "transformer" helper, as well as not having to rewrite existing config "transformers" that were already present from `with-redefs`.

I think once we move to `with-app-with-config` (if ever) then we should totally pass the entire map. Otherwise, I think this approach is more composable (eg., who wins if we bind the entire config twice?).

src/ctia/schemas/search_agg.clj
```diff
@@ -6,15 +6,22 @@
...
   {s/Keyword
    (st/optional-keys
-    {:gte s/Inst
-     :lt s/Inst})})
+    {:gte (s/conditional
+            string? s/Str
+            :else s/Inst)
+     :lt (s/conditional
+           string? s/Str
+           :else s/Inst)})})
```
frenchy64, line 14: Changes in this file were needed after adding schema fixtures to `ctia.stores.es.crud-test` (also in this PR).

test/ctia/test_helpers/core.clj
```diff
@@ -27,6 +27,21 @@
...
+          "app must be the result of get-current-app")
+  (case svc-kw
+    :ConfigService {:get-config p/get-global-properties
+                    :get-in-config p/get-in-global-properties}
+    :StoreService {:all-stores (fn [] @store/stores)
+                   :read-store store/read-store
+                   :write-store store/write-store}
+    (throw (ex-info (str "No service: " svc-kw)
+                    {:app app
+                     :service svc-kw}))))
```
frenchy64, line 43: This has a general implementation for TK that I discovered how to write recently.

```clojure
(defn get-service-map [app svc-kw]
  {:pre [(keyword? svc-kw)]}
  (let [graph (app/service-graph app)
        m (svc-kw graph)]
    (assert (map? m) (str "No service " svc-kw ", found " (keys graph)))
    m))
```

src/ctia/task/migration/store.clj
frenchy64: Ok, I'm hesitant to touch this code in this PR, so I'd prefer for one of us to resolve a merge conflict. If you merge #968 I can merge this PR back on master.

src/ctia/task/migration/store.clj
```diff
@@ -54,19 +56,31 @@
...
-  [prefix store-key]
-  (let [migration-default-es-props (dissoc (p/get-in-global-properties [:ctia :migration :store :es :default])
-                                        :indexname)]
+  [prefix store-key get-in-config]
+  (let [migration-default-es-props (-> (get-in-config [:ctia :migration :store :es :default])
+                                       (dissoc :indexname))]
     (cond-> (es.init/get-store-properties store-key)
       prefix (update :indexname
                      #(prefixed-index % prefix))
-      :always (-> (into migration-default-es-props)
```
frenchy64, line 68: This `into` above seems to do nothing, removed it.

src/ctia/task/migration/store.clj
```diff
@@ -188,15 +203,19 @@
...
-          :mapping :type :settings))
-
-(defn bulk-metas
+  [conn-state :- StoreMap
+   services :- MigrationStoreServices]
+  (-> conn-state 
+      (set/rename-keys {:indexname :index})
+      (dissoc :mapping :type :settings)
+      (assoc :services (MigrationStoreServices->ESConnServices
+                         services))))
```
frenchy64, line 212: The above change adds the `:services` key, otherwise is a refactoring to use `->`.

src/ctia/task/migration/store.clj
```diff
@@ -230,23 +250,23 @@
         {modified true not-modified false} (group-by #(search-real-index? aliased %)
                                                      with-metas)
         modified-by-ids (fmap first (group-by :id modified))
-        bulk-metas-res (->> (map :id modified)
-                            (bulk-metas store-map))
+        bulk-metas-res (bulk-metas store-map (map :id modified) services)
```
frenchy64, line 253: `->>` doesn't make work with the new `services` arg.

src/ctia/task/migration/store.clj
frenchy64: Deleted after merge.

src/ctia/task/migration/store.clj
```diff
@@ -655,15 +693,20 @@ when confirm? is true, it stores this state and creates the target indices."
...
-  (->> (migration-store-properties)
-       es.init/init-store-conn
-       :conn
-       (reset! migration-es-conn)))
+  ;; END global services setup
+  (reset! migration-es-conn
+          (-> (migration-store-properties services)
+              (es.init/init-store-conn (MigrationStoreServices->ESConnServices
+                                         services))
+              :conn)))
```
frenchy64, line 712: Refactored `->>` to `->` to accomodate service args.

test/ctia/stores/es/init_test.clj
```diff
@@ -32,8 +34,9 @@
 
 (deftest init-store-conn-test
   (testing "init store conn should return a proper conn state with unaliased conf"
-    (let [{:keys [index props config conn]}
-          (sut/init-store-conn props-not-aliased)]
+    (let [services (->ESConnServices)
```
frenchy64, line 37: There will be no `app` in this namespace because there are no `fixture-ctia` fixtures to start CTIA. Thus we use `->ESConnServices` over `app->ESConServices`.

ereteog (commented):
The `stores` atom is not used anymore in the migration, you can remove the `StoreService` from the `MigrationStoreServices` (and thus probably remove it and directly use `get-in-config`).

src/ctia/task/migration/store.clj
frenchy64: They definitely fail! Reverting.

frenchy64:
@ereteog thanks! I'll keep the `MigrationStoreServices` as a map of services instead of passing just `get-in-config` because it must at least contain the services of `ESConnServices` (even though `MigrationStoreServices->ESConnServices` is currently just `identity`, it'll make it easier to grow).

src/ctia/schemas/search_agg.clj
frenchy64: Will do.

test/ctia/stores/es/crud_test.clj
```diff
@@ -277,7 +290,7 @@
          (sut/make-aggregation {:agg-type :cardinality
                                 :aggregate-on "observable.value"}))))
 
-(defn generete-sightings
+(defn generate-sightings
```
frenchy64, line 293: Fix typo in name.

test/ctia/stores/es/crud_test.clj
```diff
@@ -202,19 +212,22 @@
...
             es-query-string-no-op {:query_string {:query query-string}}
-            date-range {:created {:gte "2020-04-01T00:00:00.000Z"
-                                  :lt "2020-05-01T00:00:00.000Z"}}
+            date-range {:created {:gte #inst "2020-04-01T00:00:00.000Z"
+                                  :lt #inst "2020-05-01T00:00:00.000Z"}}
             es-date-range {:range date-range}
-            filter-map {"disposition" 2
-                        "observable.type" "domain"}
+            filter-map {:disposition 2
+                        :observable.type "domain"}
```
frenchy64, line 230: Update these tests to conform to existing schemas (this PR enables schema checking for this namespace).

test/ctia/stores/es/crud_test.clj
```diff
@@ -316,15 +329,18 @@
...
                             (update :props assoc :default_operator "AND"))
           _ (create-fn es-conn-state
                        search-metrics-entities
                        ident
                        {:refresh "true"})
           query-string "title1"
-          date-range {:created {:gte timestamp-1
-                                :lt timestamp-2}}
+          date-range {:created {:gte (inst/read-instant-date timestamp-1)
+                                :lt (inst/read-instant-date timestamp-2)}}
```
frenchy64, line 343: Suggestions welcome for alternative Date constructors.

ereteog (approved):
Nice!
My comment on the `migration-es-conn` is up to you.
You just need to fix the `run-migration` and for me it's ready to be merged :-).

PR #985, opened 2020-09-08
fix indicator search by tags

ereteog:
This PR fixes Indicator search query schema. `tags` filter was defined as `s/Int` instead of `s/Str`.

<a name="qa">[§](#qa)</a> QA
============================

check that indicator search route properly filters documents by `tags` field.


frenchy64 (approved):
Nice, thanks!

PR #990, opened 2020-09-15
TK3.12: GraphQL realize-fn

frenchy64:
**Epic**: https://github.com/threatgrid/iroh/issues/3986

This PR broadens the definition of a "realize-fn" to be able to return another `fn` that accepts services.

Two new "services" are introduced (stubbed, really): `GraphQLService` and `GraphQLNamedTypeRegistryService`. The former uses the latter during initialization, so I decided to split them.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: TK3.12: GraphQL realize-fn
```

### DIFF
--- a/src/ctia/domain/entities.clj
+++ b/src/ctia/domain/entities.clj
@@ -1,8 +1,12 @@
 (ns ctia.domain.entities
   (:require [clj-momo.lib.time :as time]
             [ctia.domain.access-control :refer [properties-default-tlp]]
-            [ctia.properties :as p :refer [get-http-show]]
-            [ctia.schemas.core :as ctia-schemas :refer [TempIDs]]
+            [ctia.properties :refer [get-http-show]]
+            [ctia.schemas.core :as ctia-schemas
+             :refer [GraphQLRuntimeOptions
+                     MaybeDelayedRealizeFn
+                     MaybeDelayedRealizeFnResult
+                     TempIDs]]
             [ctim.domain.id :as id]
             [ctim.schemas.common :refer [ctim-schema-version]]
             [schema.core :as s]))
@@ -29,8 +33,14 @@
                     (:end_time prev-valid-time)
                     time/default-expire-date)}})
 
-(defn default-realize-fn [type-name Model StoredModel]
-  (s/fn default-realize :- StoredModel
+(s/defn default-realize-fn
+  ;; commented since StoredModel is a value argument.
+  ;; checking is propagated to the body via s/fn
+  #_#_:- (MaybeDelayedRealizeFn StoredModel)
+  [type-name
+   Model :- (s/protocol s/Schema)
+   StoredModel  :- (s/protocol s/Schema)]
+  (s/fn default-realize :- (MaybeDelayedRealizeFnResult StoredModel)
     ([new-object :- Model
       id :- s/Str
       tempids :- (s/maybe TempIDs)
@@ -43,6 +53,10 @@
       owner :- s/Str
       groups :- [s/Str]
       prev-object :- (s/maybe StoredModel)]
+    (s/fn :- StoredModel
+     [{{{:keys [get-in-config]} :ConfigService}
+       :services
+       :as _rt-opt_} :- GraphQLRuntimeOptions]
      (let [now (time/now)]
        (merge new-object
               {:id id
@@ -54,11 +68,11 @@
                :modified now
                :timestamp (or (:timestamp new-object) now)
                :tlp (:tlp new-object
-                          (:tlp prev-object (properties-default-tlp p/get-in-global-properties)))}
+                          (:tlp prev-object (properties-default-tlp get-in-config)))}
               (when (contains-key? Model :valid_time)
                 (make-valid-time (:valid_time prev-object)
                                  (:valid_time new-object)
-                                 now)))))))
+                                 now))))))))
 
 (defn short-id->long-id [id get-in-config]
   (id/short-id->long-id id #(get-http-show get-in-config)))
--- a/src/ctia/domain/entities.clj
+++ b/src/ctia/domain/entities.clj
@@ -54,9 +54,7 @@
       groups :- [s/Str]
       prev-object :- (s/maybe StoredModel)]
     (s/fn :- StoredModel
-     [{{{:keys [get-in-config]} :ConfigService}
-       :services
-       :as _rt-opt_} :- GraphQLRuntimeOptions]
+     [{{{:keys [get-in-config]} :ConfigService} :services} :- GraphQLRuntimeOptions]
      (let [now (time/now)]
        (merge new-object
               {:id id
--- a/src/ctia/domain/entities.clj
+++ b/src/ctia/domain/entities.clj
@@ -34,7 +34,7 @@
                     time/default-expire-date)}})
 
 (s/defn default-realize-fn
-  ;; commented since StoredModel is a value argument.
+  ;; commented since StoredModel is a parameter and not in scope here,
   ;; checking is propagated to the body via s/fn
   #_#_:- (MaybeDelayedRealizeFn StoredModel)
   [type-name
--- a/src/ctia/domain/entities.clj
+++ b/src/ctia/domain/entities.clj
@@ -5,9 +5,10 @@
             [ctia.schemas.core :as ctia-schemas
              :refer [GraphQLRuntimeOptions
                      MaybeDelayedRealizeFn
-                     MaybeDelayedRealizeFnResult
+                     RealizeFnResult
                      TempIDs]]
             [ctim.domain.id :as id]
+            [ctia.graphql.delayed :as delayed]
             [ctim.schemas.common :refer [ctim-schema-version]]
             [schema.core :as s]))
 
@@ -40,7 +41,7 @@
   [type-name
    Model :- (s/protocol s/Schema)
    StoredModel  :- (s/protocol s/Schema)]
-  (s/fn default-realize :- (MaybeDelayedRealizeFnResult StoredModel)
+  (s/fn default-realize :- (RealizeFnResult StoredModel)
     ([new-object :- Model
       id :- s/Str
       tempids :- (s/maybe TempIDs)
@@ -53,7 +54,7 @@
       owner :- s/Str
       groups :- [s/Str]
       prev-object :- (s/maybe StoredModel)]
-    (s/fn :- StoredModel
+    (delayed/fn :- StoredModel
      [{{{:keys [get-in-config]} :ConfigService} :services} :- GraphQLRuntimeOptions]
      (let [now (time/now)]
        (merge new-object
--- a/src/ctia/domain/entities.clj
+++ b/src/ctia/domain/entities.clj
@@ -4,7 +4,7 @@
             [ctia.properties :refer [get-http-show]]
             [ctia.schemas.core :as ctia-schemas
              :refer [GraphQLRuntimeContext
-                     MaybeDelayedRealizeFn
+                     RealizeFn
                      RealizeFnResult
                      TempIDs]]
             [ctim.domain.id :as id]
@@ -37,7 +37,7 @@
 (s/defn default-realize-fn
   ;; commented since StoredModel is a parameter and not in scope here,
   ;; checking is propagated to the body via s/fn
-  #_#_:- (MaybeDelayedRealizeFn StoredModel)
+  #_#_:- (RealizeFn StoredModel)
   [type-name
    Model :- (s/protocol s/Schema)
    StoredModel  :- (s/protocol s/Schema)]
--- a/src/ctia/domain/entities.clj
+++ b/src/ctia/domain/entities.clj
@@ -3,7 +3,7 @@
             [ctia.domain.access-control :refer [properties-default-tlp]]
             [ctia.properties :refer [get-http-show]]
             [ctia.schemas.core :as ctia-schemas
-             :refer [GraphQLRuntimeOptions
+             :refer [GraphQLRuntimeContext
                      MaybeDelayedRealizeFn
                      RealizeFnResult
                      TempIDs]]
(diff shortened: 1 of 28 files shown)


src/ctia/schemas/core.clj
```diff
@@ -38,6 +43,144 @@
...
+    [:IEncryption [:decrypt :encrypt]]))
+
+(s/defschema GraphQLRuntimeOptions
+  "A map of options to resolve a DelayedGraphQLValue"
+  {:services RealizeFnServices})
+
+(defn delayed-graphql-value?
+  "A flat predicate deciding if the argument is delayed."
+  [v]
+  (fn? v))
```
frenchy64, line 77: I wasn't comfortable using anything but a `clojure.lang.Fn` check here. A realize-fn seems to return just about anything except a `fn`.

src/ctia/schemas/core.clj
```diff
@@ -38,6 +43,144 @@
...
+  (select-keys-in
+    services
+    [:ConfigService [:get-in-config]]
+    [:StoreService [:read-store]]
+    [:GraphQLNamedTypeRegistryService [:get-or-update-named-type-registry]]
+    [:IEncryption [:decrypt :encrypt]]))
+
+(s/defschema GraphQLRuntimeOptions
+  "A map of options to resolve a DelayedGraphQLValue"
+  {:services RealizeFnServices})
```
frenchy64, line 72: I made this an extensible map argument instead of just a services arg since we can only do this "delay by returning function" trick once, and if we need additional non-service values it'll come in really handy.

src/ctia/domain/entities.clj
```diff
@@ -29,8 +33,14 @@
                     (:end_time prev-valid-time)
                     time/default-expire-date)}})
 
-(defn default-realize-fn [type-name Model StoredModel]
-  (s/fn default-realize :- StoredModel
+(s/defn default-realize-fn
+  ;; commented since StoredModel is a parameter and not in scope here,
+  ;; checking is propagated to the body via s/fn
+  #_#_:- (MaybeDelayedRealizeFn StoredModel)
```
frenchy64, line 39: This fails to compile, as explained, but I found it useful documentation.

src/ctia/graphql/schemas.clj
```diff
@@ -155,8 +160,14 @@
                               p/connection-arguments)
                  :resolve (res/search-entity-resolver :weakness)}}))
 
-(def schema (g/new-schema QueryType))
-(def graphql (g/new-graphql schema))
+(s/def schema :- (MaybeDelayedGraphQLValue GraphQLSchema)
+  (g/new-schema QueryType))
+(s/def graphql :- (MaybeDelayedGraphQLValue GraphQL)
+  (g/new-graphql schema))
```
frenchy64, line 166: Part of the point of introducing delayed graphql values is so vars like the above don't need to be changed. They do, however, change their schema. There are dozens of vars that I didn't explicitly add `s/def`'s for because they're not "directly" used, just stuffed into a global map somewhere---it would probably be too noisy to add schemas.

I think the above `graphql` and `schema` values are important enough to annotate though, since they're used directly to as an entry point to building the GraphQL "graph".

src/ctia/schemas/graphql/helpers.clj
```diff
@@ -258,25 +283,28 @@
...
+   rt-opt :- GraphQLRuntimeOptions]
   (let [builder
         (-> (GraphQLArgument/newArgument)
             (.name arg-name)
-            (.type arg-type)
+            (.type (-> arg-type (resolve-with-rt-opt rt-opt)))
             (.description (or arg-description "")))]
     (when (some? arg-default-value)
       (.defaultValue builder arg-default-value))
     (.build builder)))
```
frenchy64, line 302: `s/defn` doesn't support type hints on the argv unfortunately.

test/ctia/entity/entities_test.clj
```diff
@@ -1,23 +1,32 @@
...
-  (let [properties [:id :type :owner :groups :schema_version
+  (let [realize-fn-services (realize-fn-global-services)
+        properties [:id :type :owner :groups :schema_version
                     :created :modified :timestamp :tlp]
         ;; properties to dissoc to get a valid entity when
         ;; using the spec generator
         to-dissoc [:disposition]]
     (doseq [[_ {:keys [realize-fn] :as entity}] sut/entities]
-      (when realize-fn
+      (when-some [realize-fn (some-> realize-fn
```
vemv, line 27: Any way to strengthen this section so that we can ensure that tests are being run at all?

e.g. the `{:keys [realize-fn] :as entity}` destructuring may become stale at some point and the `when-some` would never pass

vemv (approved):
LGTM!

src/ctia/schemas/core.clj
```diff
@@ -38,6 +43,144 @@
...
+
+(s/defn resolve-with-rt-opt :- GraphQLValue
+  "Resolve a MaybeDelayedGraphQLValue value, if needed, using given runtime options."
+  [maybe-fn :- AnyMaybeDelayedGraphQLValue
+   rt-opt :- GraphQLRuntimeOptions]
+  (if (delayed-graphql-value? maybe-fn)
+    (maybe-fn rt-opt)
+    maybe-fn))
+
+(s/defn MaybeDelayedRealizeFn->RealizeFn
```
yogsototh, line 174: I think this fn should be lowercase, it doesn't appear to return a schema but really simply apply args to a function.
It was surprising to see its usage later in the code.

src/ctia/flows/crud.clj
```diff
@@ -141,15 +147,20 @@
...
            flow-type
            identity
            tempids
            prev-entity
            realize-fn] :as fm} :- FlowMap]
   (let [login (auth/login identity)
-        groups (auth/groups identity)]
+        groups (auth/groups identity)
+        realize-fn (MaybeDelayedRealizeFn->RealizeFn
+                     realize-fn
```
yogsototh, line 161: I'm generally not fan of reusing symbols in a let.
This is generally confusing.
But I see why you did it in this case.

src/ctia/schemas/core.clj
yogsototh: This look really dangerous. But considering the CTIA code is quite stable now this should be fine.

Another option would be to be extra protective and have a ns that deal with "delayed" objects.
A record, or simply `{:delayed f}`.
So the check will be a lot easier to really trust.
So instead of replacing all `(defn foo [& args] body)` by `(defn foo [& args] (fn [& args] body))`
you could instead return `(defn foo [& args] {:ctia.delayed/delayed (fn [& args] body))})`
so the `delayed-graphql-value?` would probably be a lot safer.

src/ctia/schemas/core.clj
yogsototh: So `GraphQLRuntimeOptions` contains all the services graph.
And It could probably contain other things added during the init.
So I think it looks pretty similar of the idea of ServiceContext to me because it looks used the same way.

So maybe `GraphQLRuntimeContext` because with `Options` it doesn't sound like the Options contains handlers.

src/ctia/schemas/core.clj
```diff
@@ -38,6 +43,144 @@
...
+  "A flat predicate deciding if the argument is delayed."
+  [v]
+  (fn? v))
+
+(defn resolved-graphql-value?
+  "A flat predicate deciding if the argument is not delayed."
+  [v]
+  (not (delayed-graphql-value? v)))
+
+(s/defn MaybeDelayedRealizeFnResult
```
yogsototh, line 84: I was really confused with or use of `Maybe` because in Haskell/Purescript this is about nullable types.
Here what you have is closer to an `Either`. More precisely `Either DelatedRealizeFn Result`.
So I'll let you be the judge if you could switch to Either instead of `Maybe`.

yogsototh (approved):
LGTM! 

Still notice it was very difficult to follow.
By using an optional level of indirection this makes things a lot harder to read and to follow.
Hopefully, schemas were quite helpful but even with those it is still hard to follow.

The old `realize-fn` became continuations for the new one.
Another layer of confusion at least to me, is that you kept the old name of the function while changing their type from a `fn` returning something to a continuation.
We generally tried to use the suffix `-fn` or name those functions `gen-xxx-fn` when those functions return a function.

So that's pretty smart, but I'm afraid it might make the overall code quite confusing during debugging sessions. So that's not at all a blocker but really something to consider.

src/ctia/schemas/core.clj
frenchy64: I had the same thought overnight too. I think I'll do something like you suggest.

src/ctia/schemas/core.clj
frenchy64: Ah yes!

test/ctia/entity/entities_test.clj
frenchy64: Hmm I can't think of any way of strengthening this. `:realize-fn` an optional entry of an `ctia.schemas.core/Entity`, which `sut/entities` is asserted to have `Entity` keys in the `src`.

vemv: Yeah it seems a bit hard tbh

Maybe sth like...?

```clj
(let [corpus (filter has-realize-fn? sut/entities)]
  (assert (seq corpus))
  (doseq [i corpus]
    ))
```

src/ctia/flows/crud.clj
frenchy64: I'll remove the surrounding destructuring so it isn't shadowed.

src/ctia/schemas/core.clj
frenchy64: Thanks I'll find better names.

test/ctia/entity/entities_test.clj
frenchy64: Ah yes, I like that better thanks!

src/ctia/schemas/core.clj
frenchy64: This was in the spirit of the `->Foo` and `map->Foo` constructor functions, that name the types being coerced. I understand your confusion, but I'm not sure what other behavior you were expecting here, could you explain?

src/ctia/lib/utils.clj
```diff
@@ -134,3 +134,33 @@
...
+                  _ (assert (vector? keys) (pr-str keys))]
+              (-> out
+                  (update-in prefix
+                             (fn [old]
+                               (assert (nil? old) (str "Repeated prefix " prefix))
+                               (-> m
+                                   (get-in prefix)
+                                   (select-keys keys)))))))
+          {}
+          paths))
```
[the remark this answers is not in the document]
frenchy64: Thanks, I'll rename it to `service-subgraph` (since it slices a `service-graph`) and change its interface to just accept a key instead of a path.

```clojure
  (service-subgraph
    services
    :ConfigService [:get-in-config]
    :StoreService [:read-store]
    :GraphQLNamedTypeRegistryService [:get-or-update-named-type-registry]
    :IEncryption [:decrypt :encrypt])
```

src/ctia/schemas/core.clj
frenchy64: I introduced a `ctia.graphql.delayed` namespace that encapsulates a simple wrapper datatype.

test/ctia/entity/entities_test.clj
frenchy64: Done. I also collapsed the nested `doseq`s.

src/ctia/lib/utils.clj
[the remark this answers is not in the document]
frenchy64: Done.

src/ctia/lib/utils.clj
```diff
@@ -134,3 +134,40 @@
...
+  "
+  [graph & selectors]
+  {:pre [(map? graph)]}
+  (reduce (fn [out [service-kw fn-kws :as sel]]
+            (assert (keyword? service-kw)
+                    (pr-str service-kw))
+            (assert (every? keyword? fn-kws)
+                    (pr-str fn-kws))
+            (assert (= 2 (count sel))
+                    (str "Uneven number of selectors: "
```
vemv, line 162: Doesn't check evenness

frenchy64: Doesn't it? This checks if each element of a `(partition 2)` is of even length.

frenchy64: Ah now I see why, whoops. Didn't expect `partition` to drop elements.

frenchy64: Confused it with `partition-all`.

ereteog (approved):
LGTM

src/ctia/lib/utils.clj
```diff
@@ -134,3 +134,38 @@
...
+    :MissingService [:m1])
+  ;=> {:ConfigService {:get-config <...>}
+  ;    :FooService {:f2 <...>}}
+  "
+  [graph & selectors]
+  {:pre [(map? graph)]}
+  (reduce (fn [out [service-kw & [fn-kws :as is-even]]]
+            (assert is-even
+                    (str "Uneven number of selectors: "
+                         (count selectors)))
```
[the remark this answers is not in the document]
frenchy64: I'll add comments in the code. Since we're reducing over a `(partition-all 2)` each partition is either size 1 or 2. All but maybe the last one will be size 2.

This avoids `count` which would retraverse sequences.

FWIW there are tests that ensure this works, which is what enabled me to refactor this away from `count` more confidently via TDD.

src/ctia/schemas/core.clj
frenchy64: Renamed to `lift-realize-fn-with-context`.

src/ctia/schemas/core.clj
frenchy64: Simplifed to `RealizeFn` and `ResolvedRealizeFn`.

PR #993, opened 2020-09-18
Fix pagination

ereteog:
> Close #threatgrid/iroh/issues/4192

The inconsistent results in feeds are due to the `all-pages` function that performs a `search_after` based pagination using the default `sort_by` parameter. However this default value is base on `_doc` field which is unique per shard but not per index! Indeed this is a problem that can occur for any `search_after` pagination that will use the default `sort_by` value. 
This PR changes the default `sort_by` value to `_doc,id` that will first sort by `_doc` to ensure the order then by `id` for disambiguation. It also generalizes tests with multiple shards to better reflect production properties.

<a name="qa">[§](#qa)</a> QA
============================

1. Create feed data with a single bulk that contains all data of a "big" feed:
- 1 indicator
- 350 judgements
- 350 relationships linking every created judgments to the created indicator (judgement id as `source_ref` and indicator id as `target_ref`).
2. Create the feed instance for the previous indicator.
3. make 10 calls to corresponding view `/ctia/feed/{feed-id}/view?s={feed-share-token}` and check that the responses are always the same.

```
intern: fixed edge case pagination bug.
```

### DIFF
--- a/test/ctia/entity/event_test.clj
+++ b/test/ctia/entity/event_test.clj
@@ -455,17 +455,17 @@
                         q (uri/uri-encode
                             (format "entity.id:\"%s\"" initial-id))
                         results (map :fields
-                                     (:parsed-body (get (str "ctia/event/search?query=" q)
+                                     (:parsed-body (get (str "ctia/event/search?sort_by=timestamp&query=" q)
                                                         :content-type :json
                                                         :headers {"Authorization" "user1"})))]
-                    (is (= '[nil
-                             [{:field :assignees
-                               :action "added"
-                               :change {:after ["1"]}}]
-                             [{:field :assignees
-                               :action "modified"
-                               :change {:before ["1"], :after ["1" "2"]}}]
-                             [{:field :assignees
-                               :action "deleted"
-                               :change {:before ["1" "2"]}}]]
+                    (is (= [nil
+                            [{:field :assignees
+                              :action "added"
+                              :change {:after ["1"]}}]
+                            [{:field :assignees
+                              :action "modified"
+                              :change {:before ["1"], :after ["1" "2"]}}]
+                            [{:field :assignees
+                              :action "deleted"
+                              :change {:before ["1" "2"]}}]]
                            results)))))))))))))
--- a/test/ctia/entity/feed_test.clj
+++ b/test/ctia/entity/feed_test.clj
@@ -1,5 +1,6 @@
 (ns ctia.entity.feed-test
   (:require
+   [ctia.entity.feed :as sut]
    [clojure.string :as string]
    [clj-momo.test-helpers.core :as mth]
    [clj-http.client :as client]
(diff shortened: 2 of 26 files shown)


test/ctia/entity/event_test.clj
```diff
@@ -455,17 +455,17 @@
                         q (uri/uri-encode
                             (format "entity.id:\"%s\"" initial-id))
                         results (map :fields
-                                     (:parsed-body (get (str "ctia/event/search?query=" q)
+                                     (:parsed-body (get (str "ctia/event/search?sort_by=timestamp&query=" q)
```
ereteog, line 458: added a sort to ensure result order.

test/ctia/entity/feed_test.clj
```diff
@@ -252,13 +171,16 @@
                          :feed_view_url)))
 
           (let [feed-view-url (:feed_view_url updated-feed)
-                response (client/get feed-view-url
-                                     {:as :json})
+                response  (with-redefs [sut/fetch-limit 19] ;; < |judgements| and not a multipe of nb shards (5).
```
ereteog, line 174: if we revert back the default sort by to `_doc` this test fails because for each `_doc` value, we have 5 (number of shards) instances that share this value. Thus when sorting on `_doc` with a page size that is not a multiple of 5, we will not return the 5 documents for the last page value and then skip the missing documents with `search_after` at next page call.

test/ctia/entity/feed_test.clj
```diff
@@ -232,7 +149,9 @@
           response-body-txt-wrong-secret (:body response-txt-wrong-secret)]
 
       (is (= 200 (:status response-txt)))
-      (is (= "187.75.16.75\n187.75.16.76\n187.75.16.77"
+      (is (= (->> (map #(-> % :observable :value) judgements)
+                  sort
+                  (string/join "\n"))
```
vemv, line 154: here you can assert that the computed value is not `string/blank?`

test/ctia/test_helpers/aggregate.clj
```diff
@@ -251,15 +255,23 @@
...
-                          constantly)]
-      (hc/post-bulk {plural docs})
-      (doseq [field enumerable-fields]
-        (test-cardinality docs entity field)
-        (test-topn docs entity field 3))
-      (doseq [field date-fields]
-        (test-histogram docs entity field :day)
-        (test-histogram docs entity field :month)))))
+  ;; enforce 1 shard to avoid ES terms approximation used by topn which is not simulated here by the manual aggregation here.
+  ;; see ES details: https://www.elastic.co/guide/en/elasticsearch/reference/current/search-aggregations-bucket-terms-aggregation.html#search-aggregations-bucket-terms-aggregation-approximate-counts
```
vemv, line 259: `s/current/7.9` for a more permanent link

test/ctia/entity/feed_test.clj
vemv: It is possible (although not a particularly common technique) to `with-redefs [default-sort-field :_doc]`, and run some "negative" tests, to prove that the tests fail with the old value.

That way, we gain confidence:

* that the tests are passing for the precise reason we believe they are
  * Maybe something else is performing the desired fix
* that the test is valid
  * maybe it would pass in all cases

A handy trick:

```clj
(is (cond-> (= (count judgements)
               (count (:judgements response-body)))
      negative? not))
```

vemv (approved):
lgtm!

some minor suggestions

NeelSaini:
Verified in rel v1.58

PR #994, opened 2020-09-21
Use fake TK app over global state several tests

frenchy64:
**Epic** https://github.com/threatgrid/iroh/issues/3986

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: Use fake TK app over global state several tests
```

### DIFF
--- a/test/ctia/bundle/routes_test.clj
+++ b/test/ctia/bundle/routes_test.clj
@@ -10,7 +10,7 @@
              [test :as t :refer [deftest is join-fixtures testing use-fixtures]]]
             [ctia.bulk.core :as bulk]
             [ctia.bundle.core :as core]
-            [ctia.store :as store :refer [stores]]
+            [ctia.store-service :as store-svc]
             [ctia.properties :as p]
             [ctia.auth.capabilities :refer [all-capabilities]]
             [ctia.test-helpers
@@ -211,7 +211,10 @@
                                          "foouser"
                                          "foogroup"
                                          "user")
-     (let [indicators [(mk-indicator 0)
+     (let [app (helpers/get-current-app)
+           {:keys [all-stores]} (helpers/get-service-map app :StoreService)
+
+           indicators [(mk-indicator 0)
                        (mk-indicator 1)]
            sightings [(mk-sighting 0)
                       (mk-sighting 1)]
@@ -366,7 +369,7 @@
                        (map :result (:results bundle-result-update)))
                "All existing entities are not updated")))
        (testing "Partial results with errors"
-         (let [indicator-store-state (-> @stores :indicator first :state)
+         (let [indicator-store-state (-> (all-stores) :indicator first :state)
                indexname (:index indicator-store-state)
                ;; close indicator index to produce ES errors on that store
                _ (es-index/close! (:conn indicator-store-state) indexname)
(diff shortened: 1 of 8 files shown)


test/ctia/bundle/routes_test.clj
```diff
@@ -908,17 +915,17 @@
...
+  (helpers/with-config-transformer*
+    #(-> %
+         (assoc-in [:ctia :access-control :min-tlp] tlp)
+         (assoc-in [:ctia :access-control :default-tlp] tlp))
+    f))
 
 (deftest bundle-tlp-test
-  (test-for-each-store
-   (fn []
+ (with-tlp-property-setting "amber"
```
frenchy64, line 925: We must bring `with-tlp-property-setting` outside `test-for-each-store` so the former can change the config via `with-config-transformer*`  before the latter starts an app.

test/ctia/test_helpers/core.clj
```diff
@@ -194,7 +194,11 @@
 
 (defn build-get-in-config-fn []
   ;; stub implementation until trapperkeeper bootstrap
```
frenchy64, line 196: A slightly more useful stub which is closer to the final impl.

test/ctia/bundle/routes_test.clj
```diff
@@ -1002,4 +1009,4 @@
          (is (= 201 (:status incident-post-res)))
          (is (= 201 (:status link-res)))
          (is (= 200 (:status bundle-get-res)))
-         (is seq (-> bundle-get-res :parsed-body :casebooks)))))))
+         (is (seq (-> bundle-get-res :parsed-body :casebooks))))))))
```
vemv, line 1012: :( thanks for this one

I checked that there are no more occurences: `rg --pcre2 '\(is\s(?!\().{1}'`

test/ctia/entity/asset_test.clj
```diff
@@ -22,8 +21,10 @@
 
 (defn additional-tests [asset-id asset-sample]
   (testing "GET /ctia/asset/search"
+   (let [app (helpers/get-current-app)
+         {:keys [get-in-config]} (helpers/get-service-map app :ConfigService)]
     ;; only when ES store
-    (when (= "es" (p/get-in-global-properties [:ctia :store :asset]))
+    (when (= "es" (get-in-config [:ctia :store :asset]))
```
vemv, line 27: Maybe https://github.com/threatgrid/ctia/pull/965#discussion_r477188465

vemv (approved):
lgtm, btw :)

test/ctia/entity/asset_test.clj
ereteog: indeed there is no other store implemented yet, so we could simply replace all the patterns`(when (= "es" (get-in-config [:ctia :store store-kw])) body)` by  `body`.

ereteog (approved):
LGTM

test/ctia/entity/asset_test.clj
ereteog: And if we need to exclude a potential new store later we could use the `when-not` pattern. WDYT?

yogsototh (approved):
LGTM

test/ctia/entity/asset_test.clj
vemv: > And if we need to exclude a potential new store later we could use the when-not pattern. WDYT?

sure!

PR #996, opened 2020-09-23
Port ctia.properties and IEncryption to trapperkeeper

frenchy64:
> **Epic** https://github.com/threatgrid/iroh/issues/3986

Finally deletes `get-in-global-properties` and friends!

Note: the build times on CTIA have been fluctuating between 25-40 minutes over the last few days. This PR has very slow build times today but it may have more to do with Travis that this code. At least, I have no idea what I did if it is (and would like to know!).

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: Port ctia.properties and IEncryption to trapperkeeper
```

### DIFF
--- a/src/ctia/encryption/default.clj
+++ b/src/ctia/encryption/default.clj
@@ -23,9 +23,9 @@
   (init [this context]
     (log/info "Loading Encryption Key")
     (let [secret (encryption-key (get-in-config [:ctia :encryption]))]
-      (into context
-              {:encrypt-fn #(encrypt-as-base64 % secret)
-               :decrypt-fn #(decrypt-from-base64 % secret)})))
+      (assoc context
+             :encrypt-fn #(encrypt-as-base64 % secret)
+             :decrypt-fn #(decrypt-from-base64 % secret))))
   (decrypt [this src]
     (let [{:keys [decrypt-fn]} (service-context this)]
       (decrypt-fn src)))
--- a/src/ctia/encryption/default.clj
+++ b/src/ctia/encryption/default.clj
@@ -4,7 +4,9 @@
    [clojure.string :as string]
    [ctia.encryption :refer [IEncryption]]
    [lock-key.core :refer [decrypt-from-base64
-                          encrypt-as-base64]]))
+                          encrypt-as-base64]]
+   [puppetlabs.trapperkeeper.core :as tk]
+   [puppetlabs.trapperkeeper.services :refer [service-context]]))
 
 (defn encryption-key
   [{:keys [secret key]}]
@@ -15,17 +17,18 @@
               slurp
               string/trim)))
 
-(defrecord EncryptionService [state]
+(tk/defservice default-encryption-service
   IEncryption
-  (init [this props]
+  [[:ConfigService get-in-config]]
+  (init [this context]
     (log/info "Loading Encryption Key")
-    (let [secret (encryption-key props)]
-      (reset! (:state this)
+    (let [secret (encryption-key (get-in-config [:ctia :encryption]))]
+      (into context
               {:encrypt-fn #(encrypt-as-base64 % secret)
                :decrypt-fn #(decrypt-from-base64 % secret)})))
-  (decrypt [{:keys [state]} src]
-    (let [{:keys [decrypt-fn]} @state]
+  (decrypt [this src]
+    (let [{:keys [decrypt-fn]} (service-context this)]
       (decrypt-fn src)))
-  (encrypt [{:keys [state]} src]
-    (let [{:keys [encrypt-fn]} @state]
+  (encrypt [this src]
+    (let [{:keys [encrypt-fn]} (service-context this)]
       (encrypt-fn src))))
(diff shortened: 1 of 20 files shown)


src/ctia/init.clj
```diff
@@ -103,16 +96,17 @@
 (defn start-ctia!
   "Does the heavy lifting for ctia.main (ie entry point that isn't a class).
   Returns the Trapperkeeper app."
```
frenchy64, line 98: Added an extra arity for an explicit config, for easy test stubbing of config. This will eventually be removed.

test/ctia/test_helpers/core.clj
```diff
@@ -30,45 +30,38 @@
              [utils :as fu]]
             [puppetlabs.trapperkeeper.app :as app]))
 
+(def ^:dynamic ^:private *current-app*)
 (def ^:dynamic ^:private *config-transformers* [])
 
 (defn get-service-map [app svc-kw]
```
frenchy64, line 36: Ditched `::global-app` stub for TK app, now all TK apps are real.

src/ctia/init.clj
```diff
@@ -30,44 +30,27 @@
    [puppetlabs.trapperkeeper.core :as tk]
    [puppetlabs.trapperkeeper.app :as app]))
 
-(defn init-encryption-service! []
```
frenchy64, line 33: Moved into `default-services`.

src/ctia/properties.clj
```diff
@@ -231,10 +210,19 @@
 (def configurable-properties
   (mls/keys PropertiesSchema))
 
-(def init! (mp/build-init-fn files
```
frenchy64, line 234: This logic is moved almost-verbatim to `build-init-config`.

test/ctia/encryption/default_test.clj
```diff
@@ -1,23 +1,29 @@
...
-             {:state (atom nil)})]
-    (testing "service init"
-      (try (enc/init rec {})
-           (catch Throwable e
-             (is (= (.getMessage e)
-                    "Assert failed: no secret or key filepath provided\n(or secret (:filepath key))"))))
-      (enc/init rec {:key {:filepath key-path}})
+  (let [key-path "resources/cert/ctia-encryption.key"]
+    (testing "failed service init"
+      (assert
```
vemv, line 11: Not sure if I like the assert wrapping

Replacing `false` with `(assert false)` would be equivalent, while reducing nesting and avoiding the `(assert (is ...))` path which seems to rely on the `is` ret val (which I don't think is an API)

test/ctia/test_helpers/core.clj
```diff
@@ -86,6 +85,8 @@
   "For use in a test fixture to dynamically transform a Trapperkeeper
   config before creating an app."
   [tf body-fn]
+  (assert (not (thread-bound? #'*current-app*))
+          "Cannot transform config after TK app has started!")
```
vemv, line 89: 👏great safeguarding of invariants

src/ctia/properties.clj
```diff
@@ -231,10 +210,19 @@
...
-                             PropertiesSchema
-                             ;; TOP-LEVEL STATE!!
-                             (global-properties-atom)))
+(defn build-init-config
+  "Returns a (nested keyword) config map from (merged left-to-right):
+
+  1. [[files]] merged left-to-right
```
vemv, line 216: While `[[foo]]` is also recognised by certain doc tools and IDEs, `#'foo` is less ambiguous since it makes clear that it's a var (and not an arg, or other code reference).

That decreased ambiguity enables stronger tooling (e.g. a refactoring or grep tool) and also eases repl interactions (one can send #'foo to the repl and e.g. `(-> *1 meta :doc)`)

vemv (approved):
LGTM!

src/ctia/encryption/default.clj
```diff
@@ -15,17 +17,18 @@
               slurp
               string/trim)))
 
-(defrecord EncryptionService [state]
+(tk/defservice default-encryption-service
```
yogsototh, line 20: In IROH we try to declare the protocol along the defservice.
I think it would be nice to merge the two ns in another PR.

Regarding the name of the service, I think I would name it `encryption-service` and distinct between other service implementing the same protocol via the namespace.

src/ctia/task/update_mapping.clj
```diff
@@ -32,24 +32,30 @@
...
+      (pp/pprint options)
+      (let [config (doto (p/build-init-config)
+                     init/log-properties)
+            all-stores (fn [] @store/stores)]
+        (init/init-store-service! (partial get-in config))
+        (->> (:stores options)
+             (select-keys (all-stores))
+             update-mapping-stores!)
+        (log/info "Completed update-mapping task")
+        (System/exit 0)))
```
yogsototh, line 58: Maybe you could add a catch here to try print the stacktrace of the error.
I'm afraid that `finally` could hide the uncaught exception.

yogsototh (approved):
LGTM!

src/ctia/encryption/default.clj
frenchy64: Thanks, I'll make a note. I think it'll be easy to make a sweep to use standard TK conventions after the dust has settled .

test/ctia/encryption/default_test.clj
frenchy64: Replaced with `thrown-with-msg?`.

PR #1001, opened 2020-09-25
Construct test config thread-locally

frenchy64:
Related https://github.com/threatgrid/iroh/issues/3986

This fixes a big blocker to parallelizing tests since `with-properties` sets global System properties which would make any two tests conflict. The approach is to implement `with-properties` using thread-local bindings instead, and to extract some of the original implementation for constructing the config map from `clj-momo`.

Eventually we will probably prefer merging config maps explicitly as in our other projects.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: Construct test config thread-locally
```

### DIFF
--- a/test/ctia/test_helpers/core.clj
+++ b/test/ctia/test_helpers/core.clj
@@ -29,10 +29,13 @@
             [schema.core :as s]))
 
 (def ^:dynamic ^:private *current-app*)
-;; an even-sized vector of property key-val flattened pairs (like the
-;; first argument to #'with-properties) that will be
-;; used to override the default properties.
-(def ^:dynamic ^:private *properties-overrides* [])
+(def
+  ^:dynamic ^:private
+  *properties-overrides*
+  "An even-sized vector of property key-val flattened pairs (like the
+  first argument to #'with-properties) that will be
+  used to override the default properties."
+  [])
 (def ^:dynamic ^:private *config-transformers* [])
 
 (defn get-service-map [app svc-kw]
@@ -179,7 +182,7 @@
                   log/*logger-factory* lf]
       (f))))
 
-(defn- split-property-to-keywords [prop]
+(defn split-property-to-keywords [prop]
   (map keyword (str/split prop #"\.")))
 
 (defn build-transformed-init-config
--- a/test/ctia/test_helpers/core.clj
+++ b/test/ctia/test_helpers/core.clj
@@ -71,33 +71,36 @@
      (fn [] (do ~@sexprs))))
 
 (defn fixture-properties:clean [f]
-  ;; Override any properties that are in the default properties file
-  ;; yet are unsafe/undesirable for tests
-  (with-properties ["ctia.auth.type"                            "allow-all"
-                    "ctia.access-control.default-tlp"           "green"
-                    "ctia.access-control.min-tlp"               "white"
-                    "ctia.access-control.max-record-visibility" "everyone"
-                    "ctia.encryption.key.filepath"              "resources/cert/ctia-encryption.key"
-                    "ctia.events.enabled"                        true
-                    "ctia.events.log"                            false
-                    "ctia.http.dev-reload"                       false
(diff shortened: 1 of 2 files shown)


test/ctia/test_helpers/core.clj
```diff
@@ -25,9 +25,14 @@
...
              [utils :as fu]]
-            [puppetlabs.trapperkeeper.app :as app]))
+            [puppetlabs.trapperkeeper.app :as app]
+            [schema.core :as s]))
 
 (def ^:dynamic ^:private *current-app*)
+;; an even-sized vector of property key-val flattened pairs (like the
+;; first argument to #'with-properties) that will be
+;; used to override the default properties.
+(def ^:dynamic ^:private *properties-overrides* [])
```
vemv, line 35: the comment could be a `def` docstring as well

test/ctia/test_helpers/core.clj
```diff
@@ -47,44 +51,56 @@
   (binding [*config-transformers* (conj *config-transformers* tf)]
     (body-fn)))
 
-(def with-properties-vec
-  (mth/build-with-properties-vec-fn PropertiesSchema))
+(s/defn with-properties*
+  [properties-vec :- (s/pred vector?)
```
vemv, line 55: `vector?` could be refined to assert the properties in the docstring above (`even-sized [...]`)

test/ctia/test_helpers/core.clj
```diff
@@ -47,44 +51,56 @@
...
+  (assert (even? (count properties-vec))
+          (str "Even count required for properties-vec, actual " (count properties-vec)))
+  (binding [*properties-overrides* (into *properties-overrides* properties-vec)]
+    (f)))
 
-(defmacro with-properties [properties-vec & sexprs]
-  `(with-properties-vec ~properties-vec
+(defmacro with-properties
+  "Simulates setting the specified System properties to configure
+  the creation of a TK app, except thread-locally.
```
vemv, line 66: `except` -> `except that it is` or so

(else one could think you're saying the opposite)

test/ctia/test_helpers/core_test.clj
```diff
@@ -0,0 +1,67 @@
...
+  (testing "properties are corced to PropertiesSchema"
+    (is (thrown-with-msg?
+          ExceptionInfo
+          #"Value cannot be coerced to match schema: \{\"obviously.wrong.path\" disallowed-key\}"
+          (sut/with-properties
+            ["obviously.wrong.path" "val"]
+            (sut/build-transformed-init-config)))))
+  (testing "with-properties overrides test properties without also setting System properties"
+    (sut/with-properties
+      ["ctia.auth.type" "foobar"]
```
vemv, line 57: these repeated strings are better extracted to a `let` to ensure the test will remain valid at all times

test/ctia/test_helpers/core_test.clj
```diff
@@ -0,0 +1,67 @@
...
+                     [:ctia :auth :type])
+             :foobar))))
+  (testing "overrides using with-config-transformer*"
+    (sut/with-config-transformer*
+      #(assoc-in % [:ctia :auth :type] :foobar)
+      #(is (= (get-in (sut/build-transformed-init-config)
+                      [:ctia :auth :type])
+              :foobar))))
+  (testing "overrides using with-properties and with-config-transformer*"
+    (testing "both applied with disjoint paths"
```
vemv, line 30: I have trouble understanding what `both applied with disjoint paths` means, probably it can be rewritten to feel more like a complete sentence

vemv (approved):
LGTM!

test/ctia/test_helpers/core.clj
frenchy64: Yes, evenness is asserted a few lines down.

frenchy64:
Thanks Victor!

PR #1007, opened 2020-10-01
Replace `get-free-port` with explicit port 0 config

frenchy64:
**Epic**: https://github.com/threatgrid/iroh/issues/3985

This removes uses of functions like `get-free-port` that are prone to races and simply passes port 0 to Jetty during the tests.

Due to some tricky juggling "tying the knot" (eg., providing the routes with the port and GraphQL before the server has even started), I merged GraphQLService into the server service. See `ctia.http.server-service-core/start` for the tricky parts.

Most of the other changes involve changing service graph arguments and encapsulating the port configuration via `p/get-http-show`.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: Replace `get-free-port` with explicit port 0 config
```


src/ctia/http/server_service_core.clj
```diff
@@ -1,18 +1,57 @@
...
+                               (st/dissoc-in [:CTIAHTTPServerService :get-graphql]))]
+  (let [_ (log/info "Starting HTTP server...")
+        [server graphql] (let [graphql-prm (promise)
+                               server-prm (promise)
+                               ;; tie the knot for CTIAHTTPServerService self-recursion
+                               services (-> services
+                                            (assoc-in [:CTIAHTTPServerService :get-port]
+                                                      #(server->port @server-prm))
+                                            (assoc-in [:CTIAHTTPServerService :get-graphql]
+                                                      #(deref graphql-prm)))]
```
frenchy64, line 36: This `:get-graphql` method construction is mostly the reason why the GraphQLService was merged into the server service.

src/ctia/http/server_service.clj
```diff
@@ -6,32 +6,35 @@
...
-                                     :StoreService (-> StoreService 
-                                                       (select-keys [:read-store
-                                                                     :write-store])
-                                                       store-svc/lift-store-service-fns)
-                                     :IAuth IAuth
-                                     :GraphQLService GraphQLService
-                                     :GraphQLNamedTypeRegistryService GraphQLNamedTypeRegistryService
-                                     :IEncryption IEncryption}))
-  (stop [this context] (core/stop context)))
+  (start [_ context] (core/start context
```
vemv, line 21: Can you add a newline after `[_ context]`?

Fairly handy to be able to distinguish between definitions and invocations at a glance.

src/ctia/http/server_service_core.clj
```diff
@@ -1,18 +1,57 @@
...
+    (assoc context
+           :server server
+           :graphql graphql)))
 
 (defn stop [{:keys [^Server server] :as context}]
   (some-> server .stop)
-  (dissoc context :server))
+  (dissoc context :server :graphql))
+
+(s/defn get-port :- s/Int
```
vemv, line 53: > s/Int

I'd reuse the `constrained` version that is used above

...same for other places, the port schema seems often repeated inline

vemv (approved):
lgtm!

src/ctia/http/server_service_core.clj
yogsototh: That's very nice!

yogsototh (approved):
LGTM!

src/ctia/http/server_service_core.clj
```diff
@@ -1,18 +1,57 @@
...
-               {:keys [port] :as http-config}
-               services :- APIHandlerServices]
-  (log/info (str "Starting HTTP server on port " port))
-  (assoc context
-         :services services
-         :server (new-jetty-instance http-config services)))
+               http-config
+               services :- (-> APIHandlerServices
+                               (st/dissoc-in [:CTIAHTTPServerService :get-port])
+                               (st/dissoc-in [:CTIAHTTPServerService :get-graphql]))]
```
ereteog, line 27: since `CTIAHTTPServerService` only contains what is dissoced, I would simply dissoc `:CTIAHTTPServerService` or optionalize it's keys.

doc/properties.org
```diff
@@ -60,7 +60,7 @@
 | Property              | Description                                           | Possible values |
 |-----------------------+-------------------------------------------------------+-----------------|
 | ctia.http.enabled     | enable the http server                                | =true= =false=  |
-| ctia.http.port        | set the listening port                                | number          |
+| ctia.http.port        | set the listening port. 0 for any free port.          | number          |
```
ereteog, line 63: Is it desirable in a production environment?

ereteog (commented):
I am pretty sure that I missed the main point, but I still have this question :-):
Why not having a TK dependency of `GraphQLService` over the `CTIAHTTPServerService`? 
Isn't it possible to deal with the `init` and `start` functions to handle this issue?
In addition, the initial problem seems only a test issue and I tend to avoid refactoring the code for tackling a test issue.
Finally, I think that GraphQL deserves its own service.
WDYT?

doc/properties.org
frenchy64: Probably doesn't make much sense, I'll make a note.

src/ctia/graphql_service.clj
```diff
@@ -1,27 +0,0 @@
...
-
-(defprotocol GraphQLService
-  (get-graphql [this] "Returns an instance of graphql.GraphQL"))
-
-(tk/defservice graphql-service
-  GraphQLService
-  [ConfigService
-   StoreService
-   GraphQLNamedTypeRegistryService
-   IEncryption]
```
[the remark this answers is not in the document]
frenchy64: As we discussed today, now there's a cycle between `CTIAHTTPServerService` and `GraphQLService`, since `CTIAHTTPServerService` already uses `GraphQLService`, and now `GraphQLService` uses `CTIAHTTPServerService` to get the current port instead of `ConfigService`.

Alternatively we could create a third service that just handles the dynamic configuration of the port, which `CTIAHTTPServerService` would set and `GraphQLService` would read.

src/ctia/bundle/core.clj
```diff
@@ -94,10 +94,14 @@
         (recur remaining-ext-ids acc-entities)
         acc-entities))))
 
+(s/defschema FindByExternalIdsServices
+  {:StoreService {:read-store (s/pred ifn?)
+                  s/Keyword s/Any}
+   s/Keyword s/Any})
```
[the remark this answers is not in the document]
frenchy64: I'll put it in `ctia.bundle.schemas`. I find `open-schema` isn't precise enough to convey my intention here, eg., if there's a map schema in a `s/=>`, it'll change those too.

src/ctia/http/server_service_core.clj
frenchy64: Factored out into `ctia.schemas.core/Port`.

src/ctia/bundle/core.clj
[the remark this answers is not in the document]
ereteog: I agree. But this is not the case here, is it?

ereteog (approved):
LGTM!

> Alternatively we could create a third service that just handles the dynamic configuration of the port, which CTIAHTTPServerService would set and GraphQLService would read.

Worth a try, I think that would clarify everything.

frenchy64:
> Worth a try, I think that would clarify everything.

Ok will do.

frenchy64:
@ereteog I tried creating a third service. It brought up a lot more questions than I expected that is not worth the benefits of this feature. I wanted the port to be immutable (ie., a promise, not an atom), and the extra logic needed to avoid deadlocks made me feel very uneasy.

I could go either way on this feature with the current implementation. It would be really handy to have 100% reliable parallel tests on the same machine, but it's unfortunate that GraphQLService is deleted. I'm probably leaning towards merging as-is.

Thoughts?

ereteog:
@frenchy64 it is already nice, you can merge it as it is. Thanks for trying!

PR #1008, opened 2020-10-02
fix ductile protocol parameter

ereteog:
> Related #threatgrid/iroh/issues/4216
> Related #threatgrid/iroh/issues/3072

This PR fixes what I broke :-).
Ductile uses a parameter `protocol` instead of `transport` to precise the uri protocol in the ESConn.
I was not used in CTIA tests neither in any of our deployed instances that use the default value (`http`). Thus this bug did not appear when I introduced ductile. 

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

### DIFF
--- a/resources/ctia-default.properties
+++ b/resources/ctia-default.properties
@@ -63,7 +63,7 @@ ctia.store.es.default.host=127.0.0.1
 ctia.store.es.default.port=9200
 #ctia.store.es.default.host=es-storage.int.iroh.site
 #ctia.store.es.default.port=443
-#ctia.store.es.default.transport=https
+#ctia.store.es.default.protocol=https
 ctia.store.es.default.indexname=ctia_default
 ctia.store.es.default.replicas=1
 ctia.store.es.default.refresh_interval=1s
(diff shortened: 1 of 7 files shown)


test/ctia/task/migration/store_test.clj
```diff
@@ -968,7 +968,7 @@
 (deftest target-store-properties-test
   (let [default-es-props {:host "localhost"
                           :port 9200
-                          :transport "http"
+                          :protocol "http"
```
yogsototh, line 971: 🔵  This is declared as `s/Keyword` but you use a string here.

src/ctia/task/migration/store.clj
```diff
@@ -30,7 +30,7 @@
   (st/optional-keys
    {:host s/Str
     :port s/Int
-    :transport s/Keyword
+    :protocol s/Keyword
```
yogsototh, line 33: Sounds safer to use an enum here.

```suggestion
    :protocol (s/enum :http :https)
```

resources/ctia-default.properties
```diff
@@ -159,13 +159,13 @@ ctia.log.riemann.service-prefix=Dev CTIA
 ctia.migration.migration-id=migration-test
 ctia.migration.prefix=1.2.0
 ctia.migration.migrations=identity
-ctia.migration.store-keys=malware,tool
+ctia.migration.store-keys==malware,tool
```
yogsototh, line 162: 🔴 

```suggestion
ctia.migration.store-keys=malware,tool
```

yogsototh (changes requested):
You made a typo and it could break the deployed env.

vemv (approved):

yogsototh (approved):
LGTM!

PR #1013, opened 2020-10-16
Es7 init

ereteog:
> Close #https://github.com/threatgrid/iroh/issues/4220
> Related #https://github.com/threatgrid/iroh/issues/4217
> Related #https://github.com/threatgrid/iroh/issues/4288

This PR enables CTIA to run with either Elasticsearch 5 or 7. it impacts most store operations including:
- mapping
- index initialization
- CRUD
- bulk
switching between ES 5 and ES 7 is possible throw the configuration by simply setting the parameter `ctia.store.default.es.version`. It is also configurable per store like other ES parameters and indeed we could migrate iteratively stores and have stores on ES5 and some stores on ES7 while performing the migration.

⚠️ I must disable the mapping update. I prepared the switch to ES7 which requires mapping changes of existing fields. Once this PR will be merged it will be not possible to add new fields mapping through CTIA task. If anyone needs to introduce new fields before the migration, I can update mappings manually if you tell me a nice joke.
Updating an existing field mapping with `include_in_all` does not make the mapping update fail like any other update of existing field mapping. We need to disable it until ES7 migration, and have a simple control on this. This PR includes the ability to disable mapping/setting update directly from the properties. The default value is false for a better control on the indices life cycle.

For compatibility reason it impacts the migration task and the removal of `include_in_all` which are also mostly handled here (with some few details that will be handled in coming PR dedicated to this).
The compatibility between the 2 versions are mostly delegated to our new ES library, ductile. Depending on the queried ES version, which is now part of the connection fields, ductile builds the proper uri/parameters. I kept the corresponding PR opened: https://github.com/threatgrid/ductile/pull/5.

Since I tried to delegate as far as possible the compatibility to ductile, most of the changes are in the tests to enable testing both ES 5 and ES 7. An helper macro slightly derived from the one used in ductile is used for testing both version. The `Hide whitespace changes` should protect your eyes :-).

As a reminder the migration process is as follow, for each ENV:
- let CTIA runs on current Elasticsearch cluster and indices
- start migration task (4 billion documents on NAM private CTIA, should take 1 week)
- when migration task is complete restart on new cluster / indices
As mentioned this process could be decompose for each ENV / store.

<a name="qa">[§](#qa)</a> QA
============================

This is a sensitive PR that change the core of CTIA. It must not modify any behavior on current ES5 cluster.

```
intern: Elasticsearch can now run on ES7.
```


src/ctia/stores/es/mapping.clj
```diff
@@ -9,59 +9,44 @@
 ;; even in different entities.
 
 (def float-type
-  {:type "float"
-   :include_in_all false})
```
ereteog, line 13: The removal of `include_in_all` is mandatory for ES7 init.
In ES7, the default behavior is to search in every fields if none is specified, but unlinke the former `_all` (removed in ES7) this search preserve each type and analyser which offers more precise search.
A deeper analysis of this impact will be covered when tackling https://github.com/threatgrid/iroh/issues/4288.
In any case, this change will not affect existing indices and I disable in this PR the mapping update which would failed due to this change.

test/ctia/stores/es/mapping_test.clj
```diff
@@ -4,229 +4,185 @@
              [document :as doc]
              [conn :as conn]
              [index :as index]]
-            [clj-momo.lib.es.document :refer [bulk-create-doc]]
+            [ctia.test-helpers.es :as es-helpers]
             [clojure.test :refer [deftest testing is]]))
 
 (deftest mapping-test
```
ereteog, line 10: Some context here, this is not trivial do not hesitate for any precision:
The `_all` field was removed in ES7. Instead, when no field is supplied for query_string queries in ES7, the query is applied on every fields and preserve existing mapping. This is a breaking change with our former use of `_all` field that included only some fields but analyzed them as text. This new behavior enable what was impossible: query every fields while respecting their own analyzers. Thus a keyword value is no more analyzed and a text when searching on every fields and will only be matched as if we filter only on it's value (which takes advantage of term queries: exact match ignoring case).

frenchy64 (approved):
Nice! Some minor comments about the tests, thanks for all the explanations and documentation!

test/ctia/stores/es/mapping_test.clj
```diff
@@ -4,229 +4,185 @@
...
+         (is (= #{"doc1" "doc2"}
+                (->> res-range
+                     (map :id)
+                     set)))))
+     (testing "scalar type should be properly sorted"
+       (test-sort "float" '("doc0" "doc1" "doc2"))
+       (test-sort "long" '("doc0" "doc1" "doc2"))
+       (test-sort "integer" '("doc0" "doc1" "doc2"))
+       (test-sort "boolean,id" '("doc1" "doc0" "doc2"))
+       (test-sort "table.row_count" '("doc0" "doc1" "doc2"))))))
```
agzam, line 188: Looks like a good candidate for `are`

agzam (commented):
That's a lot of work @ereteog ! Awesome. I haven't done in-depth evaluation, things looking good in GitHub diffs. Let me know if you need help testing it out.
Eventually we will move everything to v7 and would be able to remove v5, right? ES can be a memory hog, would be nice if we don't have to run an extra container.

agzam (approved):

ereteog:
@agzam yes, after the migration we will totally drop ES5. However, for now we need to live with both in tests related to ES. This PR is the biggest part of the ES7 milestone, so we should not live for long with both containers and it's easy to disable any of them when not developing on directly on the ES store behavior.

test/ctia/stores/es/mapping_test.clj
ereteog: I agree, but indeed test-sort performs 2 `is` testing ascending and descending order option for `search-docs`(it is also used above on other token and text mapping type).

test/ctia/entity/entities_test.clj
```diff
@@ -15,7 +17,8 @@
 #_
 (use-fixtures :once mth/fixture-schema-validation)
 
-(use-fixtures :each test-helpers/fixture-ctia-fast)
+(use-fixtures :once (join-fixtures [es-helpers/fixture-properties:es-store
```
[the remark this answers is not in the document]
ereteog: When I changed `:each` => `:once` it's because they only launch CTIA and/or only init some properties. The tests do not change these states, and with the `for-each-es-version`macro, the cleaning is delegated to the tests that really require it (while I doubled ES tests :-) )
So the purpose here is to limit as far as possible the actions only when it's necessary. With `:each`, in big ns tests we had fixtures applied for many tests that indeed do not require them.

> "it seems more robust to use :each for fixture-ctia"

why?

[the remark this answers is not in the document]
frenchy64: > why?

In brief, IMO it encourages test isolation, which IMO makes our test suite more robust since it is independent of test ordering (or potential parallelism).

My reasoning is that it's possible that tests can interact with stateful `:once` fixtures since we're sharing a Trapperkeeper app with several deftests. Isolating tests from each other with `:each` means we can be more confident that our tests aren't passing by coincidence. We can also be more confident that trying different test runners or different test splitting techniques or refactoring tests will not break the test suite.

test/ctia/stores/es/mapping_test.clj
frenchy64: Very minor opinion, no action asked: since `test-sort` is a function, `are` won't improve error messages. I might use `doseq` to save the redundant data and open the door to generate more cases in the future.

```clojure
(doseq [[s r] (concat
                (map vector
                     ["float"
                      "long"
                      "integer"
                      "table.row_count"]
                     (repeat '("doc0" "doc1" "doc2")))
                ;; could generate combinations of cases here
                ...
                ;; could hardcode more cases here
                ...)]
  (test-sort s r))
```

test/ctia/test_helpers/es.clj
```diff
@@ -226,3 +228,34 @@
                         (:docs.count entry))}))
          (into {})
          walk/keywordize-keys)))
+
+(defmacro for-each-es-version [msg versions clean & body]
```
vemv, line 232: :style/indent should be 3

Also please correct the argv/docstring/attrmap order

hisans:
QA Verified in 1.62 Release

PR #1015, opened 2020-10-19
Implement go/start/stop for repl dev

frenchy64:
**Epic** https://github.com/threatgrid/iroh/issues/3985

As requested by @agzam, implements `start`, `stop` and `go` for CTIA.

Some extra ceremony to prevent parallel app startup in the presence of namespace reloading, using the "define isolated var pattern" I picked up from Victor.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: Implement go/start/stop for repl dev
```

### DIFF
--- a/dev/user.clj
+++ b/dev/user.clj
@@ -22,7 +22,7 @@
 (s/validate
   {:app s/Any ;; nil or a running App
    ;; coordination mechanism for #'serially-alter-app
-   :semaphore java.util.concurrent.Semaphore}
+   :lock java.util.concurrent.locks.Lock}
   (-> (create-ns 'ctia.repl.no-reload)
       (intern 'system-state)
       (alter-var-root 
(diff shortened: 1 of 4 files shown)


agzam:
Whoa, looks great. I'll give it a try. Thanks you @frenchy64

frenchy64:
Thanks @agzam, I made a few tweaks. Should be ready now.

frenchy64:
@ereteog you might also find this handy.

ereteog (approved):
Very Nice! Thanks!

dev/user.clj
```diff
@@ -2,7 +2,95 @@
...
+  (-> (create-ns 'ctia.repl.no-reload)
+      (intern 'system-state)
+      (alter-var-root 
+        #(if (map? %)
+           ;; set root binding exactly once
+           %
+           {:app nil
+            :semaphore (java.util.concurrent.Semaphore. 1)}))))
+
+;; avoids #'locking and private vars to be robust in the presence of code reloading
```
vemv, line 35: Why avoid #'locking?

If the goal is avoiding indefinite locking, `tryLock` from https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/Lock.html might help.

Semaphores may be considered less intuitive in comparison

dev/user.clj
```diff
@@ -2,7 +2,95 @@
...
+           (throw (ex-info "Lifecycle management parallelism!"
+                           {})))
+         (finally
+           (when has-lock
+             (.release s))))))
+
+(defn current-app
+  "Returns the current app, or nil."
+  []
+  (:app ctia.repl.no-reload/system-state))
```
vemv, line 59: `ctia.repl.no-reload/*` references might be slightly fragile? Because that ns cannot be required. Might confuse some tooling.

I'd favor `resolve` (maybe DRYed behind a defn)

dev/user.clj
frenchy64: >  Why avoid #'locking?

I wanted to try and use a coordination mechanism that wasn't tied to any particular bytecode. In this case, if we redefine `serially-alter-app` in the middle of calling it, both definitions will share the same lock.

I didn't fully investigate this, but my understanding is that `locking` generates similar bytecode to Java's `synchronized` methods, and thus two separately compiled `locking` statements cannot coordinate. Does that sound right?

> tryLock from https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/Lock.html might help.

Ah, thanks! I couldn't find the `Lock` interface (was thinking "flag" for my search terms).

dev/user.clj
frenchy64: Great point, encapsulated.

dev/user.clj
vemv: > Does that sound right?

It sounds very plausible to me :) good anyway to use the safest possible approach, i.e. Lock is future-proof, as this ns could evolve.

PR #1017, opened 2020-10-21
Hide contexts based on feature flag

### LINKED ISSUE #1016: Feat: Disable routes and hide Swagger Contexts for selected entities
We need to be able to completely disable http routes associated with selected Entities and hide their contexts and endpoints in Swagger using feature-flags or some other configurable mechanism, so we could enable things for testing but disable them in production.

agzam:
> **Epic** #
> Close #1016
> Related https://github.com/threatgrid/response/issues/389

It's a WIP PR, intended for collecting early feedback.
Missing pieces:

- [  ] tests
- [  ] documentation & docstrings

for testing locally: 
- add something like `ctia.features.disable=asset,actor,sighting` to `resources/ctia-default.properties` and contexts for Asset, Actor and Sighting should not be visible in Swagger and all the corresponding routes should return 404.
 

<a name="qa">[§](#qa)</a> QA
============================

TODO

```
```

### DIFF
--- a/src/ctia/http/handler.clj
+++ b/src/ctia/http/handler.clj
@@ -5,6 +5,7 @@
             [ctia.entity.feed :refer [feed-view-routes]]
             [ctia.entity.relationship :refer [incident-link-route]]
             [ctia.schemas.core :refer [APIHandlerServices]]
+            [compojure.core]
             [compojure.api
              [core :refer [middleware]]
              [routes :as api-routes]
@@ -33,7 +34,8 @@
             [ctia.status.routes :refer [status-routes]]
             [ring.middleware.not-modified :refer [wrap-not-modified]]
             [ring.util.http-response :refer [ok]]
-            [schema.core :as s]))
+            [schema.core :as s]
+            [puppetlabs.trapperkeeper.app :as tk-app]))
 
 (def api-description
   "A Threat Intelligence API service
@@ -78,6 +80,11 @@
     (assert services->routes (str "Missing :services->routes for " (:entity entity)))
     (services->routes services-map)))
 
+(defn- entity-enabled?
+  [entity
+   {{:keys [enabled?]} :FeaturesService}]
+  (enabled? entity))
+
 (defmacro entity-routes
   [entities services-map]
   (let [gsm (gensym 'services-map)]
@@ -85,10 +92,11 @@
      (compojure.api.sweet/routes
       ~@(for [entity (remove :no-api?
                              (vals (eval entities)))]
-          `(context
-            ~(:route-context entity) []
-            :tags ~(:tags entity)
-            (entity->routes (~(:entity entity) entities) ~gsm)))))))
+          `(when (entity-enabled? ~(:entity entity) ~gsm)
+             (context
+                 ~(:route-context entity) []
+               :tags ~(:tags entity)
+               (entity->routes (~(:entity entity) entities) ~gsm))))))))
 
 (def exception-handlers
   {:compojure.api.exception/request-parsing ex/request-parsing-handler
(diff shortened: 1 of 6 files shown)


src/ctia/http/handler.clj
```diff
@@ -78,17 +80,23 @@
...
       ~@(for [entity (remove :no-api?
                              (vals (eval entities)))]
-          `(context
-            ~(:route-context entity) []
-            :tags ~(:tags entity)
-            (entity->routes (~(:entity entity) entities) ~gsm)))))))
+          `(when (entity-enabled? ~(:entity entity) ~gsm)
+             (context
+                 ~(:route-context entity) []
+               :tags ~(:tags entity)
```
ereteog, line 98: I would hide routes in Swagger like this:
```suggestion
               :tags ~(:tags entity)
               :no-doc ~(:no-doc entity)
```

src/ctia/http/handler.clj
```diff
@@ -78,17 +80,23 @@
...
+  [entity
+   {{:keys [enabled?]} :FeaturesService}]
+  (enabled? entity))
+
 (defmacro entity-routes
   [entities services-map]
   (let [gsm (gensym 'services-map)]
   `(let [~gsm ~services-map]
      (compojure.api.sweet/routes
       ~@(for [entity (remove :no-api?
```
ereteog, line 93: this already totally disable routes for some entities.

ereteog (commented):
So the initial issue is to hide the route in Swagger while this proposal disables the publication of the route. In any case it is useful, but you still have to add the ability to only hide routes in Swagger. 
BTW, I just discovered that CTIA already offers to not publish some entities. We have a `:no-api?` field on entities which is leveraged in the ` entity-routes` macro. For now this is only used by the [`identity` entity ](https://github.com/threatgrid/ctia/blob/c5ea98c00714fd805e211e48a48d4ae52c740c12/src/ctia/entity/identity.clj#L101). It should be nice to make it configurable.

* To disable route publication from the conf, the task could simply be add the `:no-api?` field to each entity with the value from the configuration. They are all listed [here](https://github.com/threatgrid/ctia/blob/c5ea98c00714fd805e211e48a48d4ae52c740c12/src/ctia/entity/entities.clj#L35), you could have a function that adds the `:no-api?` field for each one.
* To only hide the routes in Swagger, I would use a similar mechanism to add the field in entities and then simply add `:no-doc` in the macro like it is done for the `tags`.

agzam:
@ereteog 
> you still have to add the ability to only hide routes in Swagger

Ah, I think that's just a slightly incorrect wording. The point is to be able to *completely* disable the routes on selected environments and as the consequence to not to show them in Swagger. You can use `:no-doc`, yes - that hides a given route, you can hide all the routes for selected entities, but that still leaves the context, i.e. it shows the collapsible header in Swagger with no routes - that'd be confusing.

`:no-api` feature works, but like you noted, that is not controlled by config, its purpose to disable routes for an entity declaratively - in the code, not config.

So, basically the proposal is to add a configurable way of completely disabling routes, and removing tags and contexts from Swagger for selected entities. `:no-doc` and `:no-api` are not sufficient to achieve this.

ereteog:
the `:no-api?` field completely disables routes in the macro. You just need to make it configurable from the properties since the value is currently hard coded in entities (this is my point in the first bullet of my previous comment).

src/ctia/http/handler.clj
frenchy64: This probably needs to be computed at runtime, so I'd imagine more like:

```clojure
   :no-doc (not (entity-enabled? ~(:entity entity) ~gsm))
```

frenchy64:
It's probably worth updating https://github.com/threatgrid/response/issues/389 to clarify what is meant by "disabling".

agzam:
> the `:no-api?` field completely disables routes in the macro. You just need to make it configurable from the properties since the value is currently hard coded in entities (this is my point in the first bullet of my previous comment).

@ereteog If I understand it right, you're suggesting to modify entities on the fly, by adding `:no-api?` key in runtime, I'm not sure how to achieve that outside of `ctia.http.handler/entity-routes` macro. That whole thing slightly complicates this task. See, after studying that macro, I realized that it is not using `entities` that being passed into it, it is in fact using `ctia.entity.entities` (ns "global" object). That's why I have to determine for each entity if it should be disabled, the way how I did. If you see a better way, please show me in a code snippet, otherwise I am not sure what you're proposing.

src/ctia/http/handler.clj
agzam: I know that, but I don't see a way of adding the key in runtime. Pls. see the comment: https://github.com/threatgrid/ctia/pull/1017#issuecomment-716940568

src/ctia/http/handler.clj
agzam: AFAIK, `:no-doc` is different, it only hides a route in Swagger (without disabling it)

ereteog:
@agzam I tried to fight with this macro playing with the resolution time, and I have to admit that I lost some teeth :-).
We shall discuss it on CTR Services channel when you will be available.

agzam:
@ereteog My sincere apologies, I should've probably warned you sooner, I as well, spent quite some time fighting it. I even had to ask Ambrose to take a look. If we end up doing something in or around that macro, I will make sure to put some explanatory comments.

Unfortunately PR (as is) does not offer any assurance of that the suggested approach truly works, you either have to try the branch locally, or take my word. I am also hoping to add one or more tests later.

If we to discuss this thing, I feel the discussion would move towards leaving compojure.api for Reitit, but that's going to delay closing this ticket even further. Nevertheless, I will be happy to talk about it some more.

agzam:
https://github.com/threatgrid/ctia/pull/1019 even though looks like a relatively small change, is a game changer in this area. I think it would be easier to cherry-pick some things from here and redo this work in another PR. Closing it for now.

agzam:
superseded with https://github.com/threatgrid/ctia/pull/1020

PR #1020, opened 2020-10-28
Remove http routes for disabled entities

### LINKED ISSUE #1016: Feat: Disable routes and hide Swagger Contexts for selected entities

agzam:
> Close #1016
> Related https://github.com/threatgrid/response/issues/389

Remove http routes and hide Swagger contexts for disabled entities.

for testing locally: 
- add `ctia.features.disable=asset,actor,sighting` to `resources/ctia-default.properties`
- start the server, open Swagger console
- check that contexts for Asset, Actor and Sighting are not shown
 

<a name="qa">[§](#qa)</a> QA
============================

1. Check property configuration for `ctia.features.disabled` record 
2. Run the app for different environments
3. Make sure that the routes and contexts for entities like Asset, AssetMapping, AssetProperties, TargetRecord are enabled everywhere, except the environments where they set to be disabled. e.g.: prod. 

`intern:`
Related configuration changes: https://github.com/threatgrid/tenzin-config/pull/325

0feab531 PR feedback
4b48e512 convolute sexps
849136f2 PR feedback: replace if/else with cond->
8aff4aa4 PR feedback: inline fixtures
7c2fbbc0 PR feedback: remove nils
1ed45c28 PR feedback: leave only one method to reduce ambiguity
4015d267 PR feedback: contrain schema
210a897d PR feedback: rename fn
a976ff13 api tags sorted alphabetically
914fea56 test: featuresService
975d27e5 remove routes for disabled entities
46e43c8e adds tests and docstrings
72892cb7 adds FeaturesService


src/ctia/http/handler.clj
```diff
@@ -179,6 +185,13 @@
                     :tokenUrl token-url
                     :flow flow}))))
 
+(defn- check-entity-enabled
+  [{{:keys [enabled?]} :FeaturesService}
+   {:keys [entity] :as entity-map}]
+  (if (enabled? entity)
```
vemv, line 191: by using `cond->` one reifies the fact that both the 'if' and 'else' branches relate to the same subject

src/ctia/http/handler.clj
```diff
@@ -179,6 +185,13 @@
                     :tokenUrl token-url
                     :flow flow}))))
 
+(defn- check-entity-enabled
```
vemv, line 188: Maybe `mark-disabled-entities` would be a clearer name

src/ctia/properties.clj
```diff
@@ -205,7 +205,8 @@
 
                       "ctia.store.es.event.slicing.granularity"
                       (s/enum :minute :hour :day :week :month :year)})
-   (st/optional-keys {"ctia.versions.config" s/Str})))
+   (st/optional-keys {"ctia.versions.config" s/Str})
+   (st/optional-keys {"ctia.features.disable" s/Str})))
```
vemv, line 209: given the `(string/split x #",")` above I'd suggest:

`(s/constrained s/str (fn [s] (not (re-find #"\s" s))))`

since otherwise whitespace could create odd keywords

test/ctia/features_service_test.clj
```diff
@@ -0,0 +1,40 @@
...
+(use-fixtures :each (join-fixtures [properties-fixture]))
+
+(deftest features-service-test
+  (testing "FeaturesService methods for entity marked as disabled in config"
+   (th/fixture-ctia-with-app
+    (fn [app]
+      (let [{:keys [disabled? enabled?]} (th/get-service-map app :FeaturesService)]
+        (testing "Asset, Actor and Sighting are disabled"
+          (is (every? disabled? [:asset :actor :sighting]))
+          (is (every? (comp not enabled?) [:asset :actor :sighting])))
```
vemv, line 19: `(not-any? enabled? ...`?

src/ctia/features_service.clj
```diff
@@ -0,0 +1,32 @@
...
+
+  If an option like this: `ctia.features.disable=asset,actor,sighting` is added
+  to the properties file (ctia-default.properties) - http routes for Asset,
+  Actor, and Sighting entities would be disabled and their respective contexts
+  would not be visible in Swagger console."
+  (feature-flags [this]
+    "Returns all feature flags defined in the config")
+  (disabled? [this key]
+    "Returns true if the given entity key is marked as Disabled in properties config")
+  (enabled? [this key]
```
vemv, line 16: Maybe I'd offer only `enabled?` as an API since consumers can always do the `not` themselves :)

(and I'd pick enabled? over disabled? for avoiding double negatives)

src/ctia/features_service.clj
```diff
@@ -0,0 +1,32 @@
...
+    "Returns false if the given entity key is marked as Disabled in properties config"))
```
vemv, line 17: the docstring doesn't specify the `else` case

(normally I just write `whether ...` to avoid a verbose description of true/false cases)

src/ctia/http/handler.clj
```diff
@@ -115,36 +115,42 @@
...
+    [:metrics             "Metrics"             "Performance Statistics"]
+    [:properties          "Properties"          "Properties operations"]
+    [:relationship        "Relationship"        "Relationship operations"]
+    [:sighting            "Sighting"            "Sighting operations"]
+    [:status              "Status"              "Status Information"]
+    [:target-record       "Target Record"       "Target Record operations"]
+    [:tool                "Tool"                "Tool operations"]
+    [:verdict             "Verdict"             "Verdict operations"]
+    [:version             "Version"             "Version Information"]]
+   (mapv (fn [[k n desc]]
```
vemv, line 151: you're returning a vector with some nil elements here, due to the `when` below

probably undesirable

test/ctia/features_service_test.clj
```diff
@@ -0,0 +1,40 @@
+(ns ctia.features-service-test
+  (:require  [clojure.test :refer [deftest are is join-fixtures testing use-fixtures]]
+             [ctia.test-helpers.core :as th]))
+
+(defn- properties-fixture [t]
+  (th/with-properties
+    ["ctia.features.disable" "asset,actor,sighting"]
```
vemv, line 7: Because the fixture doesn't merely setup these tests but also affects their outcome, I think that these properties are better inlined in each deftest.

That way it becomes evident that there's a cause-effect relationship between the chosen properties, and the enabled/disabled properties.

And importantly, one can write a more robust test. Sample:

```clj
(testing "When the properties are set to `x,y,z`, x+y+z are disabled"
 ...)
(testing "When the properties are set to `a,b,c`, a+b+c are disabled"
 ...)
```

with a test that shows that changing the properties changes the outcome, we can be sure that the test is passing for the right reason.

vemv (approved):
lgtm!

some suggestions

agzam:
> some suggestions

Very nice suggestions. Addressed them all. Thank you @vemv

src/ctia/http/handler.clj
```diff
@@ -115,36 +115,42 @@
...
+          (when (enabled? k)
+            {:name n :description desc})))
+   (remove nil?)))
```
ereteog, line 153: Minor:
I am not sure that the vector is essential here
```suggestion
   (keep (fn [[k n desc]]
            (when (enabled? k)
               {:name n :description desc})))))
```

src/ctia/http/handler.clj
```diff
@@ -179,6 +185,12 @@
                     :tokenUrl token-url
                     :flow flow}))))
 
+(defn- mark-disabled-entities
+  [{{:keys [enabled?]} :FeaturesService}
+   {:keys [entity] :as entity-map}]
+  (cond-> entity-map
+    (not (enabled? entity)) (assoc :no-api? true)))
```
ereteog, line 192: We must preserve the hard coded value for [identity](https://github.com/threatgrid/ctia/blob/f40375ae1b690c6cf298d6cfac012c51e2134652/src/ctia/entity/identity.clj#L101), it should never be exposed in the API (and thus not be configurable). Then I would rather use update (I cannot argue, just a matter of taste :-) ):
```suggestion
 (update entity-map
             :no-api?
            #(or % (not (enabled? entity))))
```
btw, in the previous PR you had a `disabled?` counterpart that would be handy here :-)
WDYT?

src/ctia/http/handler.clj
```diff
@@ -226,9 +238,11 @@
              ;; must be before the middleware fn
              (version-routes services)
              (middleware [wrap-authenticated]
-               (entities-routes
-                 (vals entities/entities)
-                 services)
+               (->>
+                entities/entities
+                (vals)
```
ereteog, line 243:
```suggestion
                vals
```

src/ctia/properties.clj
```diff
@@ -205,7 +205,8 @@
 
                       "ctia.store.es.event.slicing.granularity"
                       (s/enum :minute :hour :day :week :month :year)})
-   (st/optional-keys {"ctia.versions.config" s/Str})))
+   (st/optional-keys {"ctia.versions.config" s/Str})
+   (st/optional-keys {"ctia.features.disable" (s/constrained s/Str #(not (re-find #"\s" %)))})))
```
ereteog, line 209: I would use a consistent naming with the feature service
```suggestion
   (st/optional-keys {"ctia.features.disabled" (s/constrained s/Str #(not (re-find #"\s" %)))})))
```

ereteog (approved):
LGTM!

test/ctia/features_service_test.clj
```diff
@@ -0,0 +1,33 @@
...
+       (fn [app]
+         (let [{:keys [disabled? enabled?]} (th/get-service-map app :FeaturesService)]
+           (testing "Asset, Actor and Sighting are disabled"
+             (is (not-any? enabled? [:asset :actor :sighting])))
+           (testing "Incident, Indicator entities are enabled"
+             (is (every? enabled? [:incident :indicator])))))))))
+
+(deftest routes-for-disabled-entities-test
+  (testing "No http routes should exist for disabled entities"
+    (th/with-properties ["ctia.features.disable" "asset,actor,sighting"]
```
vemv, line 18: I'd add a second `testing` block with a different set of properties: `["ctia.features.disable" "asset-mappng,indicator,asset-properties"]`

That way we can see that different inputs result in different outputs

vemv (commented):
Nicely applied feedback thanks!

src/ctia/http/handler.clj
ereteog: I juste saw  @vemv opposite feedback on `enabled`/`disabled` :-)
It's a minor remark, you can ignore it but I still suggest the rest of the comment :-)

agzam: But `cond->` keeps the entity map intact, and only adds `:no-api?` if there's a record in the config. If there's no such record, it just passes the map through, either it has `:no-api?` already or it doesn't. So `ctia.entity.identity` should be fine.

src/ctia/properties.clj
agzam: My idea was to indicate the intent, like "I want to disable these entities" vs. "these entities are somewhat deficient already and I need to mark them as such".

src/ctia/features_service.clj
```diff
@@ -0,0 +1,30 @@
...
+(tk/defservice features-service
+  FeaturesService
+  [[:ConfigService get-config get-in-config]]
+  (feature-flags [this]
+    (-> (get-config) :ctia :features))
+  (enabled?
+   [this key]
+   (as-> [:ctia :features :disable] x
+     (get-in-config x)
+     (string/split x #",")
```
ereteog, line 26: you must check that this value is not `nil` (empty by default):

``` clojure
user=> (clojure.string/split nil #",")
Execution error (NullPointerException) at java.util.regex.Matcher/getTextLength (Matcher.java:1283).
null
```

agzam: yeah, I missed something stupid. Thank you for spotting it

hshahman:
Verified in rel 1.64

PR #1031, opened 2020-11-10
Expose POST /:id/expire routes for Judgement, AssetMapping, AssetProperties

frenchy64:
Related https://github.com/threatgrid/iroh/issues/4293

All routes support `wait_for` and Judgement requires an extra `reason` query parameter.

Previously AssetMapping and AssetProperties used `/expire/:id` instead of `/:id/expire`. The implementation of those routes has been generalized so it can be activated by any entity via `entity-crud-routes` (assuming it doesn't have fancy requirements like extra query parameters).

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: Expose POST /:id/expire routes for Judgement, AssetMapping, AssetProperties
```


src/ctia/lib/utils.clj
```diff
@@ -173,3 +173,21 @@
                           service-fns)))))
           {}
           (partition-all 2 selectors)))
+
+(defn assoc-new-keys
+  "Like assoc, except only associates keys contained in m.
+  Note: for duplicate keys in kv, last wins.
+  
+  (assoc-new-keys m k v ...) is like (into (hash-map k v ...) m)
```
agzam, line 181: Hmmm... this statement is a bit confusing:

```
(into (hash-map :foo 2) {:foo 1}) ; => {:foo 1}
(into (hash-map :foo 1) {:foo 2}) ; => {:foo 2}
(assoc-new-keys {:foo 1} :foo 2) ; => {:foo 1}
```

For data transform functions like that, maybe it's useful to attach some examples. Perhaps even something like this: https://github.com/threatgrid/ctia/blob/master/test/ctia/test_helpers/pagination.clj#L128

frenchy64: Thanks, I'm going to ditch this for `into`.

frenchy64: It's always the helpers that get feedback isn't it... :)

frenchy64: FWIW your second example had the kv and m backwards from the description. The first and third example matched in both pattern and result.

agzam: > It's always the helpers that get feedback isn't it... :)

Probably because they are small and easily REPLable.

frenchy64: Easy bikesheddable! :D

ereteog (approved):
Nice!

src/ctia/entity/asset_mapping.clj
```diff
@@ -164,7 +125,8 @@
      :external-id-capabilities :read-asset-mapping
      :can-aggregate?           true
      :histogram-fields         asset-mapping-histogram-fields
-     :enumerable-fields        asset-mapping-enumerable-fields})))
+     :enumerable-fields        asset-mapping-enumerable-fields
+     :can-revoke?              asset-mapping-can-revoke?}))
```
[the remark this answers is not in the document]
frenchy64: I use these extra variables in the tests to activate the `/:id/expire` tests.

frenchy64:
I'll merge this then test it on INT. Then I'll see what documentation we need and if all the stakeholders are happy.

frenchy64:
Oh I forgot QA. I'll fill that out now.

kaursundeep84:
Verified in v1.64

PR #1038, opened 2020-11-16
List capabilities in swagger routes & REPL dev improvements

frenchy64:
**Epic** https://github.com/threatgrid/iroh/issues/3061
Related https://github.com/threatgrid/iroh/issues/3985
Related https://github.com/threatgrid/iroh/issues/4443

Implements first task of https://github.com/threatgrid/iroh/issues/3061: `Expose the capabilities on the route specifications so they are displayed on Swagger UI`.

Screenshot 2020-11-16 at 12 45 32 PM
Screenshot 2020-11-16 at 12 46 56 PM

Reviewers: please hide whitespace changes.

With the REPL improvements I was able to develop this feature without restarting my REPL.

<a name="qa">[§](#qa)</a> QA
============================

1. All routes that require a capability should have a description in the swagger.json correctly listing their required capabilities. See above screenshots for examples. This should include all routes except:
```
"/swagger.json"
"/doc/*.*"
"/ctia/feed/:id/view.txt"
"/ctia/feed/:id/view"
"/ctia/version"
"/ctia/status"
```

2. Please confirm that `GET /sighting/external_id/:external_id` correctly requires the `read-sighting` capability. https://github.com/threatgrid/iroh/issues/4443

```
intern: List capabilities in swagger routes & REPL dev improvements
```

### DIFF
--- a/src/ctia/auth/capabilities.clj
+++ b/src/ctia/auth/capabilities.clj
@@ -1,12 +1,12 @@
 (ns ctia.auth.capabilities
   (:require
-   [ctia.entity.entities
-    :refer [entities]]
+   [ctia.entity.entities :as entities :refer [AllEntities]]
    [clojure.set :as set]
-   [clojure.string :as string]))
+   [clojure.string :as string]
+   [schema.core :as s]))
 
-(def all-entities
-  (assoc entities
+(s/defn all-entities :- AllEntities []
+  (assoc (entities/all-entities)
          :verdict
          {:plural :verdicts
           :entity :verdict}))
@@ -27,11 +27,11 @@
                          (name plural)
                          (name entity)))))))
 
-(def all-entity-capabilities
+(defn all-entity-capabilities []
   (apply set/union
          (map #(gen-capabilities-for-entity-and-accesses
                 % (keys prefixes))
-              (vals all-entities))))
+              (vals (all-entities)))))
 
 (def misc-capabilities
   #{:read-verdict
@@ -40,10 +40,10 @@
     :specify-id
     :import-bundle})
 
-(def all-capabilities
+(defn all-capabilities []
   (set/union
    misc-capabilities
-   all-entity-capabilities))
+   (all-entity-capabilities)))
 
 (comment
 
@@ -69,8 +69,9 @@
         (str "private-intel/" loc))))
 
   (sort (set/union
-         (set (map cap-to-scope all-capabilities)))))
+         (set (map cap-to-scope (all-capabilities))))))
 
+;; TODO def => defn
 (def default-capabilities
   {:user
    #{:read-actor
@@ -104,4 +105,4 @@
      :list-weaknesses
      :import-bundle}
    :admin
-   all-capabilities})
+   (all-capabilities)})
--- a/src/ctia/auth/capabilities.clj
+++ b/src/ctia/auth/capabilities.clj
@@ -5,7 +5,10 @@
    [clojure.string :as string]
    [schema.core :as s]))
 
-(s/defn all-entities :- AllEntities []
+(s/defn all-entities
+  ;; FIXME :verdict is not an Entity
+  #_#_:- AllEntities
+  []
   (assoc (entities/all-entities)
          :verdict
          {:plural :verdicts
--- a/src/ctia/auth/capabilities.clj
+++ b/src/ctia/auth/capabilities.clj
@@ -1,13 +1,21 @@
 (ns ctia.auth.capabilities
   (:require
-   [ctia.entity.entities :as entities :refer [AllEntities]]
    [clojure.set :as set]
    [clojure.string :as string]
-   [schema.core :as s]))
+   [ctia.entity.entities :as entities]
+   [ctia.schemas.core :refer [Entity]]
+   [ctia.schemas.utils :as csu]
+   [schema.core :as s]
+   [schema-tools.core :as st]))
+
+(s/defschema EntityCapabilitySuffixes
+  (-> Entity
+      (csu/select-all-keys [:entity :plural])
+      st/required-keys
+      (st/assoc s/Keyword s/Any)))
 
 (s/defn all-entities
-  ;; FIXME :verdict is not an Entity
-  #_#_:- AllEntities
+  :- {(s/pred simple-keyword?) EntityCapabilitySuffixes}
   []
   (assoc (entities/all-entities)
          :verdict
@@ -18,10 +26,10 @@
   {:read #{:read :search :list}
    :write #{:create :delete}})
 
-(defn gen-capabilities-for-entity-and-accesses
+(s/defn gen-capabilities-for-entity-and-accesses :- #{(s/pred simple-keyword?)}
   "Given an entity and a set of access (:read or :write) generate a set of
   capabilities"
-  [{:keys [entity plural]}
+  [{:keys [entity plural]} :- EntityCapabilitySuffixes
    accesses]
   (set (for [access accesses
              prefix (get prefixes access)]
@@ -30,7 +38,7 @@
                          (name plural)
                          (name entity)))))))
 
-(defn all-entity-capabilities []
+(s/defn all-entity-capabilities [] :- #{(s/pred simple-keyword?)}
   (apply set/union
          (map #(gen-capabilities-for-entity-and-accesses
                 % (keys prefixes))
--- a/src/ctia/bundle/routes.clj
+++ b/src/ctia/bundle/routes.clj
@@ -77,7 +77,6 @@
                   :return NewBundleExport
                   :query [q BundleExportQuery]
                   :summary "Export records with their local relationships. Ids are URIs (with port if specified)."
-                  :description (common/capabilities->description capabilities)
                   :capabilities capabilities
                   :auth-identity identity
                   :identity-map identity-map
@@ -94,7 +93,6 @@
                   :query [q BundleExportOptions]
                   :body [b BundleExportIds]
                   :summary "Export records with their local relationships. Ids are URIs (with port if specified)."
-                  :description (common/capabilities->description capabilities)
                   :capabilities capabilities
                   :auth-identity identity
                   :identity-map identity-map
@@ -134,7 +132,6 @@
                      nil}]
                    :summary "POST many new entities using a single HTTP call"
                    :auth-identity auth-identity
-                   :description (common/capabilities->description capabilities)
                    :capabilities capabilities
                    (let [max-size (bundle-max-size get-in-config)]
                      (if (< max-size (bundle-size bundle))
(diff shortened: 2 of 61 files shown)


src/ctia/init.clj
```diff
@@ -88,7 +87,6 @@
   "Lower-level Trapperkeeper booting function for
   custom services and config."
   [{:keys [services config]}]
-  (validate-entities)
```
frenchy64, line 91: Moved to schema check on the return value of `ctia.entity.entities/all-entities`.

src/ctia/entity/actor.clj
```diff
@@ -153,5 +154,5 @@
    :realize-fn realize-actor
    :es-store ->ActorStore
    :es-mapping actor-mapping
-   :services->routes actor-routes
+   :services->routes #'actor-routes
```
agzam, line 157: Curious, why var quoting?

frenchy64: Since `actor-entity` is a top-level immutable data structure, it'll deref `actor-routes` only at compile-time. Reloads of `actor-routes` will not be seen.

By passing a var instead, calling `services->routes` will always see the latest value.

Ideally, `actor-entity` should eventually be a function. This is a halfway measure until that time.

agzam: Interesting. Thank you for explaining it.

ereteog: thanks for the explanation. I assume that it's for helping in the repl?
I know that it would be repeated in each entity, but a comment on the code too could help and prevent future "regressions" on this.

src/ctia/entity/entities.clj
```diff
@@ -32,7 +31,10 @@
     [investigation :refer [investigation-entity]]]
    [schema.core :as s]))
 
-(def entities
+(s/defschema AllEntities 
+  {(s/pred simple-keyword?) Entity})
+
+(s/defn all-entities :- AllEntities []
```
ereteog, line 37: 👍

src/ctia/entity/relationship.clj
```diff
@@ -90,98 +91,100 @@
...
+        capabilities #{:read-incident
+                       :read-relationship
+                       :create-relationship}]
+    (POST "/:id/link" []
+          :return rs/Relationship
+          :body [link-req IncidentLinkRequest
+                 {:description "an Incident Link request"}]
+          :summary "Link an Incident to a Casebook or Investigation"
+          :path-params [id :- s/Str]
+          :description (common/capabilities->description capabilities)
```
ereteog, line 104: You could modify this description to precise the additional required capabilities.
``` clojure
(str (common/capabilities->description capabilities)
       "read-casebook or read-investigation are also respectively required for linked casebook or investigation.")
```
(or anything else less franglish :-) )

src/ctia/graphql/routes.clj
```diff
@@ -25,70 +26,71 @@
...
+                         :read-data-table
+                         :list-weaknesses
+                         :read-weakness
+                         :list-vulnerabilities
+                         :read-vulnerability}]
+      (POST "/graphql" []
+            :tags ["GraphQL"]
+            :return gql/RelayGraphQLResponse
+            :body [body gql/RelayGraphQLQuery {:description "a Relay compatible GraphQL body"}]
+            :summary "EXPERIMENTAL: Executes a Relay compatible GraphQL query"
```
ereteog, line 77: that's a long experiment :-)

src/ctia/http/routes/common.clj
```diff
@@ -134,3 +134,19 @@
...
+                     :else #{(s/pred simple-keyword?)})]
+  (cond
+    (keyword? capabilities) (str "Requires capability " (name capabilities) ".")
+    ((every-pred set? seq) capabilities) (-> (apply str "Requires capabilities "
+                                                    (->> capabilities
+                                                         sort
+                                                         (map name)
+                                                         (str/join ", ")))
+                                             (str "."))
+    :else "No capabilities needed."))
```
ereteog, line 152: We should ask QA to check this last case is never reached :-)

src/ctia/http/routes/crud.clj
```diff
@@ -270,161 +244,207 @@
...
                            un-store)]
                 (ok updated-rec)
-                (not-found))))
+                (not-found)))))
+     (when can-patch?
+       (let [capabilities patch-capabilities]
+         (PATCH "/:id" []
+                :return entity-schema
+                :body [partial-update patch-schema {:description (format "%s partial update" capitalized)}]
+                :summary (format "Partially Update %s" capitalized)
```
ereteog, line 257:
```suggestion
                :summary (format "Partially Updates an existing %s" capitalized)
```

ereteog (approved):
LGTM!

src/ctia/http/routes/common.clj
frenchy64: I think it's reachable. Should we make it an error?

ereteog: @gbuisson wdyt?

src/ctia/http/routes/crud.clj
frenchy64: Updated these docs with more imperative language.

src/ctia/auth/capabilities.clj
```diff
@@ -1,12 +1,15 @@
...
    [clojure.set :as set]
-   [clojure.string :as string]))
+   [clojure.string :as string]
+   [schema.core :as s]))
 
-(def all-entities
-  (assoc entities
+(s/defn all-entities
+  ;; FIXME :verdict is not an Entity
+  #_#_:- AllEntities
```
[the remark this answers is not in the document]
frenchy64: I rolled my own `select-all-keys` to implement this. See `all-entities feedback + some generative testing practice` commit.

src/ctia/entity/actor.clj
frenchy64: Great idea. I made a `reloadable-function` macro instead of the opaque `#'` syntax and added a docstring there instead.

src/ctia/http/routes/common.clj
frenchy64: It looks like the `"/external_id/:external_id"` routes might not always have a capability. I'll do more digging.

frenchy64: https://github.com/threatgrid/iroh/issues/4443

src/ctia/entity/relationship.clj
frenchy64: Nice catch! Generated this docstring by moving some data around.

Screenshot 2020-11-17 at 8 02 20 PM

frenchy64:
I added a test for the presence of `:description` in our swagger routes.

yogsototh:
Before reading the code, I must comment that, we should probably duplicate this feature a bit to handle `scopes`.

So when we use CTIA with JWT filtering (as our 2 main instances) we transform the `scopes` declared in the JWT in a set of capabilities.
Users have a UI where they could see their scopes, but there is no way a user can see his/her list of  capabilities. 
So I think if the node uses JWT we should also display a list of scopes with probably a short sentence explaining the relation with capabilities in Swagger UI.

Also we should probably expose a route so clients can get their exhaustive list of capabilities (I don't think we have one).

src/ctia/http/routes/common.clj
```diff
@@ -134,3 +134,47 @@
...
+  this means updates to the route function will
+  not be observed until *both* the entity namespace and
+  the server's routes have been reloaded.
+
+  Using (var v) instead of v in top-level
+  maps enables routes to dynamically
+  update."
+  [v]
+  {:pre [(symbol? v)]}
+  `(var ~v))
```
yogsototh, line 180: That's neat!

src/ctia/bundle/routes.clj
```diff
@@ -71,65 +72,71 @@
...
-                     identity-map
-                     identity
-                     q
-                     services)))
+           (let [capabilities export-capabilities]
+             (GET "/export" []
+                  :return NewBundleExport
+                  :query [q BundleExportQuery]
+                  :summary "Export records with their local relationships. Ids are URIs (with port if specified)."
+                  :description (common/capabilities->description capabilities)
```
yogsototh, line 80: So I think this could be problematic, we should probably use something like this: 

https://github.com/threatgrid/iroh/blob/master/lib/iroh-web/src/iroh_web/compojure_api.clj#L25

Because we probably want to be able to prepend the capabilities to existing description.

yogsototh (approved):
LGTM and proposed to show scopes when this is relevant. Also you should probably use `compojure-api` restructure params to be able to automatically prepend the `capabilities` to existing description.

src/ctia/bundle/routes.clj
frenchy64: > So I think this could be problematic, we should probably use something like this:

I don't think that's possible because `:capabilities` is runtime information in CTIA. `-meta/restructure-param :scopes` takes compile-time information.

frenchy64:
@yogsototh thanks for the review! I added your suggestions to the epic https://github.com/threatgrid/iroh/issues/3061

src/ctia/http/routes/common.clj
frenchy64: Made an error.

src/ctia/bundle/routes.clj
frenchy64: Another piece of the puzzle is that the swagger metas can't see the `:lets` in scope in the same `restructure-param` call https://github.com/metosin/compojure-api/blob/1.1.13/src/compojure/api/meta.clj#L321

src/ctia/http/routes/common.clj
vemv, line 180: I don't get how writing `(reloadable-function f)` is different from writing `#'f`?

frenchy64:
It prevents having to respond to future questions about why var is used here.

src/ctia/http/routes/common.clj
frenchy64: It prevents having to respond to future questions about why var is used here.

vemv: 👍it's not perfect (since one normally doesn't alias clojure.core fns to denote intent) but it works.

Maybe the `reloadable-function` docstring could be clearer in reflecting that it doesn't do something special.

Or as an alternative:

`^:do-not-unwrap-var #'foo, ^:do-not-unwrap-var #'bar` etc. git log/blame would do the rest

frenchy64: I concluded that since 3 team members asked me about it, that using var here is "special enough" to warrant this. Obviously I didn't start this PR with `reloadable-function`.

You can add meta to vars with that syntax? I thought that only worked with IObjs.

PR #1041, opened 2020-11-18
DELETE search route

ereteog:
> Close #https://github.com/threatgrid/iroh/issues/4385

This PR introduces a `DELETE /ctia/{entity}/search` route that enables users to delete entities that match given filters.
The search filters are the same than those of the `search` / `count`and metrics routes.
In order to prevent (as far as possible) deletion errors, the route triggers a dry run by default that only returns the number of matched entities and empty queries (that would match all entities) are forbidden. One can still delete all it's visible entities by using the `query` filter with `*`.
I also reduced some boilerplate in access control tests and improved es store crud tests.

Reviewers: you should hide whitespace changes.

<a name="qa">[§](#qa)</a> QA
============================

let 3 users
user1 in group1 with search and delete capabilities
user2 is in group1 with search and delete capabilities
user3 is in group1 with only search capability
user4 is in group2 with search and delete capabilities

1) user 1 post a bundle with 3 sightings green, 2 sighting amber, 1 sighting red

2) user4 makes a `DELETE /ctia/sighting/search` query that matches all created sightings with `REALLY_DELETE_ALL_THOSE_ENTITIES` parameter set `true`.
==> Check that the response is 0 and that no entities were deleted.

3) user2 makes a `DELETE /ctia/sighting/search` query that matches all created sightings with `REALLY_DELETE_ALL_THOSE_ENTITIES` parameter set `false`.
==> Check that the response is 5 and that no entities were deleted.

4) user2 makes a `DELETE /ctia/sighting/search` query that matches all created sightings without `REALLY_DELETE_ALL_THOSE_ENTITIES` parameter.
==> Check that the response is 5 and that no entities were deleted.

5) user2 makes a `DELETE /ctia/sighting/search` without any filters and `REALLY_DELETE_ALL_THOSE_ENTITIES` parameter et to true.
==> Check that the response has status 403 and that no entities were deleted.

6) user3 makes a `DELETE /ctia/sighting/search` query that matches all created sightings with `REALLY_DELETE_ALL_THOSE_ENTITIES` parameter set to `true`.
==> Check that the response has status 401 and that no entities were deleted.

7) user2 makes a `DELETE /ctia/sighting/search` query that matches all created sightings with `REALLY_DELETE_ALL_THOSE_ENTITIES` parameter set to `true`.
==> Check that the response is 5 and that only the 3 green and 2 amber entities were deleted.

8) user1 makes a `DELETE /ctia/sighting/search` query that matches all created sightings with `REALLY_DELETE_ALL_THOSE_ENTITIES` parameter set to `true`.
==> Check that the response is 1 and that only the remaining 2 red entities were deleted.

```
intern: introduced DELETE search route.
```


src/ctia/task/purge_es_stores.clj
```diff
@@ -1,40 +0,0 @@
-(ns ctia.task.purge-es-stores
```
ereteog, line 1: this is not maintained nor used anymore.
One can simply launch 
* `DELETE http://localhost:9200/ctia_*` to delete CTIA indices
* `DELETE http://localhost:9200/_template/ctia_*` to delete CTIA indices

src/ctia/http/routes/crud.clj
```diff
@@ -310,34 +311,74 @@
...
+                                              " the inevitable disaster that will occur after"
+                                              " you perform that operation."
+                                              " DO NOT FORGET TO SET THAT TO FALSE AFTER EACH DELETION"
+                                              " IF YOU INTEND TO USE THAT ROUTE MULTIPLE TIMES."))})]
+             (let [query (->> (dissoc params :wait_for :REALLY_DELETE_ALL_THOSE_ENTITIES)
+                              (search-query date-field))]
+               (if (empty? query)
+                 (forbidden {:error "you must provide at least one of from, to, query or any field filter."})
+                 (ok
+                  (if (:REALLY_DELETE_ALL_THOSE_ENTITIES params)
```
frenchy64, line 370: `THOSE` => `THESE`

test/ctia/test_helpers/es.clj
```diff
@@ -256,10 +256,14 @@
...
-           (testing (format "%s (ES version: %s)." msg#  ~'version)
+           (testing (format "%s (ES version: %s).\n" msg#  ~'version)
              (when clean-fn#
                (clean-fn# ~'conn))
-             ~@body)
+             (h/with-properties*
+               ["ctia.store.es.default.port" ~'es-port
+                "ctia.store.es.default.version" ~'version]
+               (fn []
+                 ~@body)))
```
frenchy64, line 266: I'd recommend using `(h/with-properties [...] ~@body)` (the macro) or wrapping the body in a `do` `(h/with-properties* [...] (fn [] (do ~@body))` to distinguish from the :pre/post syntax.

frenchy64 (approved):
LGTM!

src/ctia/entity/event/store.clj
```diff
@@ -14,6 +14,7 @@
     (crud/handle-create state new-events))
   (list-events [this filter-map ident params]
     (crud/handle-list state filter-map ident params))
+  (close [_] (close-cm! state))
```
agzam, line 17: Perhaps, at some point we could discuss renaming `ctia.stores.es.store/close-cm!`. Arguably, `cm` maybe not very obvious abbreviation,  without jumping to where the fn is implemented, it's not clear. But, anyone who reads the code has to do that only once, so, really not that big of a deal.

agzam (approved):
Looks good.

src/ctia/entity/event/store.clj
ereteog: I agree. I always prefer easier code to read. I replaced it by `close-connections` (indeed it closes a connection manager).

NeelSaini:
verified in rel v1.63

PR #1042, opened 2020-11-18
make ES indices unique during tests

frenchy64:
**Epic** https://github.com/threatgrid/iroh/issues/4447

Constructs all configurations with unique ES indices by default. Adds cleanup logic after every (test) app shutdown to delete all indices and templates created during the app.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: make ES indicies unique during tests
```

### DIFF
--- a/test/ctia/test_helpers/core.clj
+++ b/test/ctia/test_helpers/core.clj
@@ -253,7 +253,7 @@
        :get-in-config)))
 
 ;; workaround cycle with this namespace
-(def ^:private purge-indices (delay (requiring-resolve 'ctia.test-helpers.es/purge-indices)))
+(def ^:private purge-indices-and-templates (delay (requiring-resolve 'ctia.test-helpers.es/purge-indices-and-templates)))
 
 (defn ^:private stop-and-cleanup
   "Stop and garbage collect a running app."
@@ -265,7 +265,7 @@
         get-in-config (partial get-in (get-config))
         all-stores (constantly (all-stores))]
     (app/stop app)
-    (@purge-indices
+    (@purge-indices-and-templates
       all-stores
       get-in-config)))
 
--- a/test/ctia/test_helpers/core.clj
+++ b/test/ctia/test_helpers/core.clj
@@ -356,7 +356,6 @@
                                                   :capabilities caps})))
 
 (defmacro deftest-for-each-fixture-with-app [test-name fixture-map app & body]
-  (assert (seq fixture-map))
   (assert (simple-symbol? app) (pr-str app))
   `(do
      ~@(for [[name-key fixture-fn] fixture-map]
--- a/test/ctia/test_helpers/core.clj
+++ b/test/ctia/test_helpers/core.clj
@@ -64,7 +64,7 @@
 (assert (even? (count *properties-overrides*)))
 
 (defn- isolate-config-indices
-  "Updates all ES indices to be unique."
+  "Updates all ES indices in config to be unique."
   [config]
   (let [suffix (UUID/randomUUID)]
     (-> config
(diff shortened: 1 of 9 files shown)


frenchy64:
@ereteog what do you think of this approach? Where should we delete these indices?

ereteog (commented):
I like the approach but I would prefer to turn it off by default and rather have an explicit option to turn it on with `fixture-ctia-with-app`. My concern is the readability of the tests and ease their understanding without having to look at implementation details.
wdyt?

frenchy64:
Can you expand more on the readability? Do you mean the readability of the test output, or the code itself? This should only affect the former. If a UUID is too ugly, we can probably also use gensym with similar results.

frenchy64:
@ereteog more directly, could you give an example where non-unique indicies improves the test suite in some way?

ereteog:
I mean the readability of the test itself.
When you see a test containing:
```clojure
(with-ctia-fixture-app
    (fn [app]
        ...
    ))
```
It's not explicit that the configuration is overrided and to discover it you have to:
- go to the code of `with-ctia-fixture-app` definition
- detect `(build-transformed-init-config)` that have some effect on the configuration
- go to `build-transformed-init-config`
- detect the use of `*config-transformers*`
- go to `*config-transformers*` definition
- go to `isolate-config-indices`

As a summary, we will both know it, but it will be harder for other developers to deal with it.
A solution could be then to have an explicit options like `enable-http?` to turn it on.

Then, In any case, this will go well with an helper that would enable to easily retrieve the state of the store to emphasize a best practice. Perhaps something like the one that I wrote in my current delete-by-query PR:
``` 
(defn get-conn-state
  [store-kw app]
  (let [{:keys [all-stores]} (helpers/get-service-map app :StoreService)]
    (-> (all-stores)
         :store-kw
        first
        :state)
```

Finally, there is the default-properties and `ctia.test-helpers.es/fixture-properties:es-store`. We should review the default behavior for our test.

frenchy64:
Ah yes good point. Giving `with-ctia-fixture-app` a config argument would also fix this problem. I think we should be able to do something like this:

```clojure
(with-ctia-fixture-app
   (build-transformed-init-config)
    (fn [app]
        ...
    ))
```

If that's still not clear enough, we might expose a different function `(build-transformed-init-isolated-config)` that uniquifies the indices.

ereteog:
since we already have an option `enable-http?`, perhaps passing a map of options with `enable-http?` and `isolated-indices?`. 
``` clojure
(with-ctia-fixture-app
   {:isolated-indices? true
    :enable-http? false}
    (fn [app]
        ...
    ))
```
Calling `(build-transformed-init-config)` would be an implementation detail resulting from setting `isolated-indices?` to `true` that could be easily found since it would be explicit.

frenchy64:
> Calling (build-transformed-init-config) would be an implementation detail ...

You probably meant `isolate-config-indices`?

I'm a little puzzled at your choice of defaults here, are you really sure generating a config that is appropriate only for sequential tests is a good default? In the spirit of making test isolation easy and making sequential tests explicit, I think `:isolated-indices?` should default to true.

We choose a good default for isolated tests for the server port, for example.

I'll let you have the final word on that.

ereteog:
Ok I agree with you to default it to true.
At some point, when we will have stabilized our test practices, we should document it.

frenchy64:
Ok sounds good. What are your thoughts about garbage collection of these indices? Perhaps `fixture-ctia-with-app` could delete all indices mentioned in the config after the app is stopped? I don't think this needs to touch `src` either, just an extra line in `fixture-ctia-with-app`.

ereteog:
yes! Indeed we will need to clean the templates too that will be created with the indexname.

frenchy64:
Ok good to know, I'll try it out.

test/ctia/test_helpers/core.clj
```diff
@@ -230,14 +252,31 @@
...
+      all-stores
+      get-in-config)))
+
 (s/defn fixture-ctia-with-app
+  "Note: ES indicies are unique, use `with-config-transformer`
+  to make them explicit."
   ([t-with-app :- (s/=> s/Any
-                        (s/=> s/Any
-                              (s/named s/Any 'app)))]
+                        (s/named s/Any 'app))]
```
frenchy64, line 276: These types were wrong.

frenchy64:
@ereteog ok this should be ready to go.

test/ctia/test_helpers/core.clj
```diff
@@ -266,17 +305,24 @@
...
-             (app/stop app))))))))
-
-(defn fixture-ctia
-  ([t] (fixture-ctia t true))
-  ([t enable-http?]
+             (stop-and-cleanup app))))))))
+
+(s/defn fixture-ctia
+  "Note: ES indices are unique, use `with-config-transformer`
+  to make them explicit."
```
ereteog, line 312: 👍

ereteog (approved):
LGTM!

PR #1047, opened 2020-12-07
re-asssign entity owner on the event update

agzam:
Re-assign owner on the event update flow

> **Epic** #
> Close https://github.com/threatgrid/iroh/issues/3596
> Related #

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
```

### DIFF
--- a/test/ctia/entity/event_test.clj
+++ b/test/ctia/entity/event_test.clj
@@ -1,27 +1,19 @@
 (ns ctia.entity.event-test
   (:require
-   [ctim.schemas.common :refer [ctim-schema-version]]
-   [ctim.domain.id :refer [str->short-id]]
-   [clojure
-    [string :as str]
-    [test :refer [is testing]]]
-   [clj-momo.lib.time :as time]
-   [clj-momo.lib.clj-time.core :as t]
-   [clojure.test :refer [deftest use-fixtures]]
-   [ctim.domain.id :as id]
-   [ctia.entity.event :as ev]
-   [ctia.test-helpers
-    [auth :refer [all-capabilities]]
-    [core :as helpers
-     :refer [DELETE POST PUT GET fixture-with-fixed-time
-             with-sequential-uuid]]
-    [fake-whoami-service :as whoami-helpers]
-    [store :refer [test-for-each-store-with-app]]]
-   [ctim.examples.incidents :refer [new-incident-minimal]]
-   [ctim.examples.casebooks :refer [new-casebook-minimal]]
-   [ctim.domain.id :as id]
    [cemerick.uri :as uri]
-   [ctia.test-helpers.es :as es-helpers]
+   [clj-momo.lib.clj-time.core :as t]
+   [clj-momo.lib.time :as time]
+   [clojure.test :refer [is testing deftest use-fixtures]]
+   [ctia.auth.capabilities :refer [all-capabilities]]
+   [ctia.entity.event :as ev]
+   [ctia.test-helpers.core :as helpers
+    :refer [DELETE POST PUT GET fixture-with-fixed-time with-sequential-uuid]]
+   [ctia.test-helpers.fake-whoami-service :as whoami-helpers]
+   [ctia.test-helpers.store :refer [test-for-each-store-with-app]]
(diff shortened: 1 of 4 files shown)


test/ctia/flows/crud_test.clj
```diff
@@ -1,10 +1,9 @@
...
-  (:require [ctia.flows.crud :as sut]
-            [ctia.lib.collection :as coll]
-            [clj-momo.test-helpers
-             [core :as mth]]
-            [clj-momo.lib.map :refer [deep-merge-with]]
-            [clojure.test :refer [deftest testing is]]))
+  (:require [clj-momo.lib.map :refer [deep-merge-with]]
+            [clojure.test :refer [deftest testing is]]
+            [ctia.auth.threatgrid :refer [map->Identity]]
+            [ctia.flows.crud :as crud]
```
ereteog, line 5: I like the convention of using `sut` (system under test).
In any test ns that you look, you immediately now that it refers to the tested ns.

agzam: I like it over (`sut` naming convention), because:

- It doesn't make it harder to review the code without seeing the entire ns or jumping to the header
- It simplifies grepping
- Overtime, it naturally, slowly forms a convention for naming aliases (familiar names take over things that are named incidentally or without much thinking)
- `sut/preserve-errors` is not very obvious; you still have to look at the filename and/or ns declaration, `crud/preserve-errors` is more pronounced, especially if you're familiar with the namespace where it's defined
- Sometimes, you are not really testing one system in the same namespace; we shouldn't, but sometimes it's inevitable. - I think SUT, CUT (class under test), AUT (application under test), et al., are the recurrence of traits from OOP practices, and it doesn't make sense to use that in Clojure.

But all these points are just my personal opinion; this isn't the kind of hill I'd chose to die on. The point of that change was exactly that: the invitation for a discussion.

ereteog: I like your arguments :-).
Clojure definitively has a problem, we always need to format code for `grep` :-), but with or without `sut` you still can grep the whole namespace to know where it is used. 
In any case it's not blocking, but most of the code mostly uses `sut`.
Since you prefer a more explicit alias that will be consistent across the project, I recommend to pick more than the suffix `crud` that is used in many namespaces (routes, es, some entities and probably others). `flows.crud` will not be ambiguous.

test/ctia/entity/event_test.clj
```diff
@@ -185,18 +177,16 @@
                 (testing "event timeline should contain all actions by user, with respect to their visibility"
 
                   (is (= '(1 3) (map :count timeline1-body)))
-                  (is (every? #(= "user1" (:owner %))
-                              timeline1-body))
-                  (is (every? #(= "user1" (:owner %))
-                              timeline2-body))
+                  (is (= (->> timeline1-body (map :owner) count)
+                         (->> timeline1-body (map :owner) set count))
+                      "owners should differ")
```
ereteog, line 182: I would be explicit on the expected list to also check that TLP is repected (as far as I checked only events from `group1` should be in this list).
```suggestion
                  (is (= (= #{"user1" "user2}
                               (set (map :owner timeline1-body))) 
                           "owners should differ")
```

test/ctia/flows/crud_test.clj
```diff
@@ -79,8 +78,21 @@
...
         "when entities are empty, apply-create-store-fn should not apply store-fn")
     (is (every? #(:applied-store-create %)
-                (:entities (sut/apply-create-store-fn flow-with-entities)))
+                (:entities (crud/apply-create-store-fn flow-with-entities)))
         "store-fn shall be applied to every entities")))
+
+(deftest create-events-test
+  (testing "flow-type :update should update the owner"
+    (let [updated-entities (atom [])
+          flow-map         {:services        {:ConfigService {:get-in-config (fn [_] true)}}
```
ereteog, line 90:
```suggestion
          flow-map         {:services        {:ConfigService {:get-in-config (constantly true)}}
```

ereteog (approved):
LGTM

ereteog:
@agzam you need to add the QA section.

PR #1053, opened 2020-12-27
Minor tooling improvements

vemv:
Related https://github.com/threatgrid/iroh/issues/3985

* Make a `PrintWriter` cross-JDK
  * I use `8`
* Introduce :ci profile, discard :dev-test profile
  * :dev and :test are standard Lein profiles that can be expected to be in essentially any Lein project, while :dev-test is not.
Requiring developers to opt-in to this profile precludes the possibility of having a unified Lein setup that will work in any project.
* Setup optional tools.namespace setup
  * Doesn't impact non-users (e.g. no code will be loaded if you won't tell it to `(refresh)`)
* Increase :repl-timeout
  * Arbitrary ~/.lein code can require such a high value.
Otherwise setting this in project.clj would override a user-provided value.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: Minor tooling improvements
```

### DIFF
--- a/build/run-tests.sh
+++ b/build/run-tests.sh
@@ -3,4 +3,4 @@
 set -ex
 
 # shellcheck disable=SC1010
-lein with-profile -dev do clean, javac, split-test :no-gen
+lein with-profile -dev,+ci do clean, javac, split-test :no-gen
--- a/dev/ctia/dev/split_tests.clj
+++ b/dev/ctia/dev/split_tests.clj
@@ -167,7 +167,11 @@
         example-nses (-> example-timings keys sort vec)]
     (doseq [nsplits (range 1 15)]
       (let [all-splits (for [id (range nsplits)]
-                         (binding [*out* (java.io.PrintWriter. "/dev/null")]
+                         (binding [*out* (java.io.PrintWriter. (proxy [java.io.OutputStream] []
+                                                                 (write
+                                                                   ([a])
+                                                                   ([a b c])
+                                                                   ([a b c d e]))))]
                            (this-split-using-scheduling-with-full-knowledge
                              example-timings
                              [id nsplits]
--- a/dev/ctia/dev/split_tests.clj
+++ b/dev/ctia/dev/split_tests.clj
@@ -167,9 +167,7 @@
         example-nses (-> example-timings keys sort vec)]
     (doseq [nsplits (range 1 15)]
       (let [all-splits (for [id (range nsplits)]
-                         (binding [*out* (java.io.PrintWriter.
-                                           ;; JDK11+
-                                           (java.io.OutputStream/nullOutputStream))]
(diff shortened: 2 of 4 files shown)


project.clj
```diff
@@ -141,7 +141,7 @@
                                   (:out (clojure.java.shell/sh
                                          "git" "symbolic-ref" "--short" "HEAD")))})}]
 
-  :global-vars {*warn-on-reflection* true}
```
vemv, line 144: Moved to the `:ci` profile because it also affects 3rd party libs, which possibly results in a wall of warns (depending on what you have in your ~/.lein) on repl startup:

Screenshot 2020-12-27 at 11 33 42

IROH's approach is better, since it affects only the project itself and also actively fails the build.

We could implement it here now (wasn't possible at all before the summer refactorings I think; code now is more reloadable), as a subsequent PR

dev/ctia/dev/split_tests.clj
```diff
@@ -167,9 +167,7 @@
         example-nses (-> example-timings keys sort vec)]
     (doseq [nsplits (range 1 15)]
       (let [all-splits (for [id (range nsplits)]
-                         (binding [*out* (java.io.PrintWriter.
-                                           ;; JDK11+
-                                           (java.io.OutputStream/nullOutputStream))]
+                         (binding [*out* (java.io.PrintWriter. "/dev/null")]
```
ereteog, line 170: this does not work on windows (shall it be?).
BTW, we need to properly identify which JDK we want to support, 11 is the last LTS, 8 is not maintained anymore.

vemv: I'd say that current consensus implcitily goes along these lines: https://github.com/threatgrid/iroh/issues/4211#issuecomment-697433583

I'll fix the /dev/null to a proper `reify` - coincidentally I have it at hand (and saved it as a handy emacs `yasnippet` ;p)

vemv:
The build failed for unrelated reasons:

image

...I'll leave it like that in case it helps, not sure if we are already chasing test flakiness

build/run-tests.sh
```diff
@@ -3,4 +3,4 @@
 set -ex
 
 # shellcheck disable=SC1010
-lein with-profile -dev do clean, javac, split-test :no-gen
+lein with-profile -dev,+ci do clean, javac, split-test :no-gen
```
vemv, line 6: Had forgotten this in my original PR, added now.

Perhaps this file could be discarded in favor of a Lein alias? That way we have a single point of change - project.clj

ereteog (approved):
LGTM

frenchy64:
Nice, thanks @vemv!

build/run-tests.sh
frenchy64: I'd be in favor of moving the logic to project.clj but keeping the file.

PR #1054, opened 2021-01-04
fix metric test time bomb

ereteog:
> **Epic** #
> Close #https://github.com/threatgrid/iroh/issues/4575

this PR removes a time bomb in tests
Former code uses a fixed time, but the creation date is controlled by the launched server.
This PR generates relative timedata instead of using a fixed date.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: fix a bug in metric test helpers.
```

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -40,6 +40,7 @@
 
   :dependencies [[org.clojure/clojure ~clj-version]
                  [clj-time "0.15.2"]
+                 [clojure.java-time "0.3.2"]
                  [org.clojure/core.async "1.0.567"]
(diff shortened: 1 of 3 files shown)


project.clj
```diff
@@ -40,6 +40,7 @@
 
   :dependencies [[org.clojure/clojure ~clj-version]
                  [clj-time "0.15.2"]
+                 [clojure.java-time "0.3.2"]
```
ereteog, line 43: I introduce java-time here, we need to get rid of clj-momo time library.

vemv (approved):
LGTM!

PR #1061, opened 2021-01-12
Remove disabled entities from Bulk routes

agzam:
Prevent disabled entities to be used in Bulk Routes

> **Epic** #
> Close https://github.com/threatgrid/iroh/issues/4549
> Related: https://github.com/threatgrid/iroh/issues/4630, https://github.com/threatgrid/iroh/issues/4471

We've added ability to disable entities, using `:FeaturesService`, see: https://github.com/threatgrid/ctia/pull/1020
Now we don't want them to be used in Bulk routes

<a name="qa">[§](#qa)</a> QA
============================

Testing path is similar to one described in https://github.com/threatgrid/ctia/pull/1020

1. Check property configuration for `ctia.features.disabled` record
3. Try to use Bulk routes with entities like  Asset, AssetMapping, AssetProperties, TargetRecord (they typically should be disabled on Prod)

For testing locally:

1. add `ctia.features.disable=asset,actor,sighting` to `resources/ctia-default.properties`
2. start the server, open Swagger console
3. try issuing request through Bulk routes using disabled entities (it should not be permitted)

9b6ccc2b rename as suggested
c091e123 change the signature, so fn denotes returning schema
335d06b8 use more permissive schema
c6ffb811 clean up ns declaration
bbe650ae remove unused schema
632f3ac2 propagate changes to BulkRefs schema
ed16e67e propagate changes to Bulk schema
bb9ed74b remove disabled entities from the schema

### DIFF
--- a/src/ctia/bulk/schemas.clj
+++ b/src/ctia/bulk/schemas.clj
@@ -1,7 +1,7 @@
 (ns ctia.bulk.schemas
   (:require [clojure.string :as str]
             [ctia.entity.entities :as entities]
-            [ctia.schemas.core :refer [TempIDs Reference]]
+            [ctia.schemas.core :refer [TempIDs Reference APIHandlerServices]]
             [schema-tools.core :as st]
             [schema.core :as s]))
 
@@ -42,6 +42,8 @@
    (entities-bulk-schema (entities/all-entities) [(s/maybe Reference)])
    (s/optional-key :tempids) TempIDs))
 
-; TODO def => defn
-(s/defschema NewBulk
-  (entities-bulk-schema (entities/all-entities) :new-schema))
+(s/defn NewBulk :- s/Any
+  "Returns NewBulk schema without disabled entities"
+  [{{:keys [enabled?]} :FeaturesService} :- APIHandlerServices]
+  (let [ents (->> (entities/all-entities) (filter (fn [[k _]] (enabled? k))))]
+    (entities-bulk-schema ents :new-schema)))
--- a/src/ctia/bulk/schemas.clj
+++ b/src/ctia/bulk/schemas.clj
@@ -35,18 +35,18 @@
   [{{:keys [enabled?]} :FeaturesService} :- ConfigurationServices]
   (->> (entities/all-entities) (filter (fn [[k _]] (enabled? k)))))
 
-(s/defn Bulk :- s/Any
+(s/defn Bulk :- (s/protocol s/Schema)
   "Returns Bulk schema without disabled entities"
   [services :- ConfigurationServices]
   (entities-bulk-schema (get-entities services) :schema))
 
-(s/defn BulkRefs :- s/Any
+(s/defn BulkRefs :- (s/protocol s/Schema)
   [services :- ConfigurationServices]
   (st/assoc
    (entities-bulk-schema (get-entities services) [(s/maybe Reference)])
    (s/optional-key :tempids) TempIDs))
 
-(s/defn NewBulk :- s/Any
+(s/defn NewBulk :- (s/protocol s/Schema)
   "Returns NewBulk schema without disabled entities"
   [services :- ConfigurationServices]
   (entities-bulk-schema (get-entities services) :new-schema))
--- a/src/ctia/bulk/schemas.clj
+++ b/src/ctia/bulk/schemas.clj
@@ -42,10 +42,10 @@
 (s/defschema StoredBulk
   (entities-bulk-schema (entities/all-entities) :stored-schema))
 
-; TODO def => defn
-(s/defschema BulkRefs
+(s/defn BulkRefs :- s/Any
+  [services :- APIHandlerServices]
   (st/assoc
-   (entities-bulk-schema (entities/all-entities) [(s/maybe Reference)])
+   (entities-bulk-schema (get-entities services) [(s/maybe Reference)])
    (s/optional-key :tempids) TempIDs))
 
 (s/defn NewBulk :- s/Any
--- a/src/ctia/bulk/schemas.clj
+++ b/src/ctia/bulk/schemas.clj
@@ -1,9 +1,7 @@
 (ns ctia.bulk.schemas
   (:require [clojure.string :as str]
             [ctia.entity.entities :as entities]
-            [ctia.schemas.core :refer [TempIDs Reference APIHandlerServices
-                                       ConfigurationServices
-                                       ]]
+            [ctia.schemas.core :refer [TempIDs Reference GetEntitiesServices]]
             [schema-tools.core :as st]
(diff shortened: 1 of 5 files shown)


src/ctia/bulk/schemas.clj
```diff
@@ -28,20 +28,23 @@
   "Error related to one entity of the bulk"
   {:error s/Any})
 
-; TODO def => defn
-(s/defschema Bulk
-  (entities-bulk-schema (entities/all-entities) :schema))
+(s/defn get-entities :- [s/Any]
+  "Returns list of enabled entities"
+  [{{:keys [enabled?]} :FeaturesService} :- APIHandlerServices]
```
frenchy64, line 33: It might help to use a more permissive schema here to get the tests to pass with schema enabled.

eg., `HTTPShowServices` exists for similar reasons https://github.com/threatgrid/ctia/blob/e6b45249024c73093ae7692e9b7ff1cbef207f7d/src/ctia/schemas/core.clj#L46

agzam: I'm not exactly sure what you're saying. Why does it have to be more open? So we don't have to do something like this? https://github.com/threatgrid/ctia/blob/6382bd7753dc4e3aceceb451ad94e05a52ab1985/test/ctia/bulk/core_test.clj#L36

ereteog: Yes, it's handy for testing with a subset of services in tests.

ereteog (approved):
LGTM

src/ctia/bulk/schemas.clj
frenchy64: Yes, we should do everything we can to avoid `(s/set-fn-validation! false)`, especially if this PR introduces the schemas that fail in the first place.

src/ctia/bulk/schemas.clj
```diff
@@ -28,20 +28,23 @@
...
+  (->> (entities/all-entities) (filter (fn [[k _]] (enabled? k)))))
 
-; TODO def => defn
-(s/defschema StoredBulk
-  (entities-bulk-schema (entities/all-entities) :stored-schema))
+(s/defn Bulk :- s/Any
```
frenchy64, line 36: This might be a more informative return schema https://github.com/threatgrid/ctia/blob/e6b45249024c73093ae7692e9b7ff1cbef207f7d/src/ctia/schemas/core.clj#L101

Ditto the other `defn`s in this ns.

test/ctia/bulk/core_test.clj
```diff
@@ -16,3 +23,24 @@
...
+  (let [reqs (->> schema st/required-keys keys)
+        opts (->> schema st/optional-keys keys (map :k))]
+    (set (concat reqs opts))))
+
+(deftest bulk-schema-excludes-disabled-test
+  (testing "ensure NewBulk schema includes only enabled entities"
+    (s/set-fn-validation! false) ;; otherwise it fails for incomplete :APIHandlerServices (we only need :FeaturesService for the test)
+    (with-app-with-config app
+      [features-svc/features-service] {:ctia {:features {}}}
+      (let [bulk-schema (NewBulk (app/service-graph app))]
```
frenchy64, line 39: This could be cleaner if you write a utility function that selects part of the service graph like https://github.com/threatgrid/ctia/blob/e6b45249024c73093ae7692e9b7ff1cbef207f7d/test/ctia/test_helpers/http.clj#L45-L50

Once you decide a name for your schema in the [other conversation](https://github.com/threatgrid/ctia/pull/1061/files#r556857544), just write a function `app->MyName` as above and use it to avoid "passing the world" here.

Using this style results in a stronger test, as we can be certain the test passes with minimal access to the app state.

frenchy64 (approved):
Nice work, but there are some details that should be addressed.

test/ctia/bulk/core_test.clj
agzam: @frenchy64 Can you give it another gander: https://github.com/threatgrid/ctia/pull/1061/commits/ede787938887ce2bafe8bf909aa649a5c78f413f
thanks!

src/ctia/schemas/core.clj
```diff
@@ -316,3 +316,12 @@
 (defschema ObservableTypeIdentifier
   vocs/ObservableTypeIdentifier
   "vocab.observable-type-id")
+
+(s/defschema ConfigurationServices
```
frenchy64, line 320: There might be a better name for this schema. I might call it `GetEntitiesServices`.

src/ctia/schemas/core.clj
```diff
@@ -316,3 +316,12 @@
...
+  {:ConfigService   {:get-in-config (s/=>* s/Any
+                                           [[s/Any]]
+                                           [[s/Any] s/Any])
+                     s/Keyword      s/Any}
+   :FeaturesService {:enabled?      (s/=> s/Bool s/Keyword)
```
frenchy64, line 325: Just `:enabled?` is used right? We can probably remove the ones we don't use.

src/ctia/schemas/core.clj
```diff
@@ -316,3 +316,12 @@
 (defschema ObservableTypeIdentifier
   vocs/ObservableTypeIdentifier
   "vocab.observable-type-id")
+
+(s/defschema ConfigurationServices
+  {:ConfigService   {:get-in-config (s/=>* s/Any
```
frenchy64, line 321: Is `:ConfigService` needed?

test/ctia/bulk/core_test.clj
```diff
@@ -16,3 +23,23 @@
...
+  (let [reqs (->> schema st/required-keys keys)
+        opts (->> schema st/optional-keys keys (map :k))]
+    (set (concat reqs opts))))
+
+(deftest bulk-schema-excludes-disabled-test
+  (testing "ensure NewBulk schema includes only enabled entities"
+    (with-app-with-config app
+      [features-svc/features-service] {:ctia {:features {}}}
+      (let [bulk-schema (NewBulk (helpers/app->ConfigurationServices app))]
+        (schema->keys bulk-schema)
```
frenchy64, line 39: This line looks like a typo.

test/ctia/bulk/core_test.clj
```diff
@@ -16,3 +23,23 @@
...
+        (is (true? (set/subset? #{:assets :actors} (schema->keys bulk-schema))))))
```
frenchy64, line 40: Might as well just write `(is (set/subset? ...))`.

frenchy64 (commented):
Nice! I added more nitpicks but this is mergable.

Could you merge the latest master also? I'm seeing old code in the diff.

hisans:
QA Verified

PR #1062, opened 2021-01-13
ES Auth

ereteog:
> Close #threatgrid/iroh#4381

Enable ES authentication.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

============================

Authentication is ready in CTIA. 

```
intern: CTIA can now use an authenticated connection to Elasticsearch.
```

### DIFF
--- a/test/ctia/task/rollover_test.clj
+++ b/test/ctia/task/rollover_test.clj
@@ -1,12 +1,11 @@
 (ns ctia.task.rollover-test
   (:require [ductile.index :as es-index]
             [clojure.string :as string]
-            [clojure.test :refer [deftest testing is]]
+            [clojure.test :refer [deftest is]]
             [ctia.stores.es.init :as init]
             [ctia.task.rollover :as sut]
             [ctia.test-helpers.core :as helpers]
             [ctia.test-helpers.es :as es-helpers]
-            [ctia.test-helpers.fake-whoami-service :as whoami-helpers]
             [ctia.test-helpers.fixtures :as fixt]))
 
 (deftest rollover-aliased-test
@@ -17,6 +16,7 @@
    (helpers/with-properties
      ["ctia.store.es.default.port" es-port
       "ctia.store.es.default.version" version
+      "ctia.store.es.default.auth" es-helpers/basic-auth
       "ctia.auth.type" "allow-all"]
      (helpers/fixture-ctia-with-app
        (fn [app]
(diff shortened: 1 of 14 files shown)


agzam (approved):
Not exactly sure what's going on here, but the code looks alright.

test/ctia/task/rollover_test.clj
```diff
@@ -64,7 +66,7 @@
            (is (= 3 (count-indices)))))))))
 
 (deftest rollover-stores-error-test
-  (with-redefs [es-index/rollover! (fn [_ alias _]
+  (with-redefs [es-index/rollover! (fn [_ alias _ _]
```
frenchy64, line 69: Minor: perhaps also provide a 3 arg arity?

test/ctia/task/rollover_test.clj
```diff
@@ -73,7 +75,8 @@
     (helpers/with-properties*
       ["ctia.auth.type" "allow-all"
        "ctia.store.es.default.port" 9207
-       "ctia.store.es.default.version" 7]
+       "ctia.store.es.default.version" 7
+      "ctia.store.es.default.auth" es-helpers/basic-auth]
```
frenchy64, line 79: Indentation.

frenchy64 (approved):
Nice!

vemv (approved):
LGTM!

Just in case, looks like https://github.com/threatgrid/ductile/blob/46099a4abf3f431a021394b95b7a93c2ad7508f5/test/ductile/auth_test.clj#L72 could also exercise the `:oauth-token, :bearer, :headers` methods?

test/ctia/task/rollover_test.clj
ereteog: I finally switched back to the 3 args arity because the 4th parameter was empty (back and forth with ductile changes, sorry). In any case this refined function is a ductile function which is used with only one arity (finally 3 args ;-)) in the rollover task.

ereteog:
> Just in case, looks like https://github.com/threatgrid/ductile/blob/46099a4abf3f431a021394b95b7a93c2ad7508f5/test/ductile/auth_test.clj#L72 could also exercise the `:oauth-token, :bearer, :headers` methods?

@vemv I confess that I was lazy here, I only tested the simplest to configure in docker: `basic-auth`, and the one that OPS chose for future ES7 cluster: API keys. 
I refactored this test function to add failing auth and ease adding any other authentication parameters in the future.
However, IMO, the responsability to perform integration testing of the different types of authentication belongs to the ES library, ductile. I created an issue for missing integration tests: https://github.com/threatgrid/ductile/issues/12

PR #1065, opened 2021-01-21
`c/described-entity-entries` in Malware, Tool, Attack Pattern

ereteog:
> Related #https://github.com/threatgrid/ctim/pull/320
> Close #https://github.com/threatgrid/iroh/issues/4058

Makes Malware, Tool and Attack Pattern described entities and provide corresponding migration.
Campaing and Actor are also changed from describable to described entities making `title`, `description` and `short_description` required.

Most of the edits in `ctia.task.migration.migrations` are moving code to a separate `ns`.

<a name="qa">[§](#qa)</a> QA
============================

After the migration, Malware, Tool and Attack Pattern  entities do not have anymore the field name and must now have `title`, `description` and `short_description` fields. For Campaing and Actor entities these fields are no more optional but required.

This schema change require a migration of the three corresponding entities. See [migration issue in tenzin-config](https://github.com/threatgrid/tenzin-config/issues/369).

```
intern: Makes Malware, Tool and Attack Pattern are now described entities with `title`, `description` and `short_description` fields.
```

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -59,7 +59,7 @@
                  [metosin/schema-tools "0.12.2"]
                  [threatgrid/flanders "0.1.23"]
 
-                 [threatgrid/ctim "1.0.23"]
+                 [threatgrid/ctim "1.0.24-SNAPSHOT"]
                  [threatgrid/clj-momo "0.3.5"]
                  [threatgrid/ductile "0.3.0"]
 
--- a/project.clj
+++ b/project.clj
@@ -48,6 +48,7 @@
                  [org.clojure/tools.logging "1.1.0"]
                  [org.clojure/tools.cli "1.0.194"]
                  [pandect "0.6.1"]
+                 [org.clojure/math.combinatorics "0.1.6"]
 
                  ;; Trapperkeeper
                  [puppetlabs/trapperkeeper ~trapperkeeper-version]
--- a/project.clj
+++ b/project.clj
@@ -59,7 +59,7 @@
                  [metosin/schema-tools "0.12.2"]
                  [threatgrid/flanders "0.1.23"]
 
-                 [threatgrid/ctim "1.0.24-SNAPSHOT"]
+                 [threatgrid/ctim "1.1.0-SNAPSHOT"]
                  [threatgrid/clj-momo "0.3.5"]
                  [threatgrid/ductile "0.3.0"]
 
--- a/project.clj
+++ b/project.clj
@@ -60,7 +60,7 @@
                  [metosin/schema-tools "0.12.2"]
                  [threatgrid/flanders "0.1.23"]
 
-                 [threatgrid/ctim "1.1.0"]
+                 [threatgrid/ctim "1.1.2-SNAPSHOT"]
                  [threatgrid/clj-momo "0.3.5"]
                  [threatgrid/ductile "0.3.0"]
 
--- a/project.clj
+++ b/project.clj
@@ -60,7 +60,7 @@
                  [metosin/schema-tools "0.12.2"]
                  [threatgrid/flanders "0.1.23"]
 
-                 [threatgrid/ctim "1.1.0-SNAPSHOT"]
+                 [threatgrid/ctim "1.1.0"]
                  [threatgrid/clj-momo "0.3.5"]
                  [threatgrid/ductile "0.3.0"]
 
--- a/project.clj
+++ b/project.clj
@@ -60,7 +60,7 @@
                  [metosin/schema-tools "0.12.2"]
                  [threatgrid/flanders "0.1.23"]
 
-                 [threatgrid/ctim "1.1.2-SNAPSHOT"]
+                 [threatgrid/ctim "1.1.1"]
                  [threatgrid/clj-momo "0.3.5"]
                  [threatgrid/ductile "0.3.0"]
 
--- a/resources/ctia-default.properties
+++ b/resources/ctia-default.properties
@@ -1,3 +1,5 @@
+# properties documentation: https://github.com/threatgrid/ctia/blob/master/doc/properties.org
+
 ctia.auth.type=allow-all
 ctia.auth.threatgrid.cache=true
 ctia.auth.entities.scope=private-intel
@@ -9,13 +11,6 @@ ctia.access-control.max-record-visibility everyone
 
 ctia.encryption.type=default
 ctia.encryption.key.filepath=resources/cert/ctia-encryption.key
-#ctia.encryption.secret=...
-
-# You can optionaly require a specific API key
-#ctia.auth.type=static
-#ctia.auth.static.name=example
-#ctia.auth.static.secret=password
-#ctia.auth.static.readonly-for-anonymous=true
 
 ctia.http.enabled=true
 ctia.http.port=3000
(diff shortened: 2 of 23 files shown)


resources/ctia-default.properties
```diff
@@ -166,17 +166,17 @@ ctia.migration.store-keys=malware,tool
...
 #ctia.migration.store.es.default.replicas=1
 #ctia.migration.store.es.default.refresh_interval=1s
 #ctia.migration.store.es.default.default_operator=AND
-ctia.migration.store.es.default.shards=1
+#ctia.migration.store.es.default.shards=1
 #ctia.migration.store.es.default.refresh=false
 #ctia.migration.store.es.default.rollover.max_docs=10000000
 #ctia.migration.store.es.default.aliased=true
-ctia.migration.store.es.default.version=5
+#ctia.migration.store.es.default.version=5
```
ereteog, line 179: all these properties are redefined in migration tests.
I let them commented to ease local test but avoid interference with production migration properties (for instance in this migration we want to keep source index shard configuration and letting it in default properties would set a new value for the created index).

project.clj
```diff
@@ -59,7 +59,7 @@
                  [metosin/schema-tools "0.12.2"]
                  [threatgrid/flanders "0.1.23"]
 
-                 [threatgrid/ctim "1.0.23"]
+                 [threatgrid/ctim "1.0.24-SNAPSHOT"]
```
ereteog, line 62: need to deploy CTIM release after merging this PR: https://github.com/threatgrid/ctim/pull/320

src/ctia/entity/attack_pattern.clj
```diff
@@ -48,10 +48,7 @@
 (def realize-attack-pattern
   (default-realize-fn "attack-pattern" NewAttackPattern StoredAttackPattern))
 
-(def attack-pattern-fields
-  (concat sorting/base-entity-sort-fields
-          sorting/sourcable-entity-sort-fields
-          [:name]))
+(def attack-pattern-fields sorting/default-entity-sort-fields)
```
ereteog, line 51: remove `:name` and add describable-fields correspond to [`default-entity-sort-fields`](https://github.com/threatgrid/ctia/blob/ac300dfa9a86cb61f49e8a490da0982e7164f234/src/ctia/schemas/sorting.clj#L15):
``` clojure
(def default-entity-sort-fields
  (concat base-entity-sort-fields
          sourcable-entity-sort-fields
          describable-entity-sort-fields))
```

src/ctia/entity/attack_pattern.clj
```diff
@@ -63,10 +60,9 @@
     (merge
      em/base-entity-mapping
      em/sourcable-entity-mapping
+     em/describable-entity-mapping
      em/stored-entity-mapping
      {:abstraction_level em/token
-      :name em/token
```
ereteog, line 68: merging this PR will make CTIA fail to restart due to a forbidden mapping change.

src/ctia/task/migration/migrations.clj
```diff
@@ -150,96 +151,6 @@
                (update :incident_time simplify-incident-time))
            doc))))
 
-;; Add new investigation fields
```
ereteog, line 153: moved to a dedicated ns.

src/ctia/task/migration/migrations/describe.clj
```diff
@@ -0,0 +1,13 @@
+(ns ctia.task.migration.migrations.describe)
```
ereteog, line 1: moved from `ctia.task.migration.migrations`

src/ctia/task/migration/migrations/investigation_actions.clj
```diff
@@ -0,0 +1,91 @@
+(ns ctia.task.migration.migrations.investigation-actions
```
ereteog, line 1: moved from `ctia.task.migration.migrations`

test/ctia/entity/casebook_test.clj
```diff
@@ -42,13 +42,11 @@
```
ereteog, line 42: In this ns I simply fixed a fixture with new malware schema, factorized this fixture, and removed dead code.

test/ctia/task/migration/migrations/investigation_actions_test.clj
```diff
@@ -0,0 +1,82 @@
+(ns ctia.task.migration.migrations.investigation-actions-test
```
ereteog, line 1: moved from `ctia.task.migration.migrations-test`

test/ctia/task/migration/migrations_test.clj
```diff
@@ -310,80 +309,3 @@
                        verdict
                        casebook])))))
 
-(deftest derive-action-data-test
```
ereteog, line 313: moved to `ctia.task.migration.migrations.investigation-actions-test`

resources/ctia-default.properties
vemv: By any chance would it make sense to have separate .properties for production, dev and test?

That way, maybe we wouldn't have to toggle comments in order to try out things.

It's somewhat common to have `dev-resources/`, `test-resources/` folders in a given project.

(If trying this approach, note you need to add said folders to Lein `:resource-paths`)

test/ctia/task/migration/migrations/describe_test.clj
```diff
@@ -0,0 +1,203 @@
...
+          (is (seq migrated-short-desc))
+
+          (if src-name
+            (is (= src-name migrated-title)
+                "the name key must be renamed into title")
+            (is (= "No title provided" migrated-title)
+                "missing name on corrupted data should be fixed."))
+
+          (when (and (seq src-name) (nil? src-desc))
+            (is (= migrated-title migrated-short-desc)
```
vemv, line 160: `is` inside `when` tends to need extra care as there's the possibility of the assertion not being ever run

For addressing that, you can extract `(and (seq src-name) (nil? src-desc))` as a reusable predicate.

Then, before the doseq over `(zipmap ...)`, you `(assert (some the-predicate (zipmap ...)))`

test/ctia/task/migration/migrations/describe_test.clj
```diff
@@ -0,0 +1,203 @@
...
+                "when provided, the title should be used as short_description on non describable entities"))
+
+          (when (and (nil? src-name) (nil? src-desc))
```
vemv, line 163: (same)

test/ctia/task/migration/migrations/describe_test.clj
```diff
@@ -0,0 +1,203 @@
...
+
+              {migrated-title :title
+               migrated-desc :description
+               migrated-short-desc :short_description} migrated]
+
+          (is (seq migrated-title))
+          (is (seq migrated-desc))
+          (is (seq migrated-short-desc))
+
+          (when (seq src-title)
```
vemv, line 186: (same)

resources/ctia-default.properties
ereteog: So here is some context. `default.properties` is not a dev file but the default properties that are used in CTIA if not defined in `ctia.properties` that is injected at built time for each envrionement and that could also be overridden by JVM properties and environment variables. [Here](https://github.com/threatgrid/ctia/blob/ac300dfa9a86cb61f49e8a490da0982e7164f234/src/ctia/properties.clj#L218) is the `build-init-config` fn. 

> By any chance would it make sense to have separate .properties for production, dev and test?

We do have separate properties for each CTIA instance (private and public intel) for environment (production, dev and test) and also for each zone for private intel (EU, US, APJC).
Here is the corresponding [tenzin-config PR](https://github.com/threatgrid/tenzin-config/pull/368).

> That way, maybe we wouldn't have to toggle comments in order to try out things.

Indeed, there is no real default properties for a migration (most of them should be specific to each migration), I should simply remove them from `default.properties`. The [migration doc](https://github.com/threatgrid/ctia/blob/master/doc/migration.md#migration-properties) is sufficiently detailed about it.

> It's somewhat common to have `dev-resources/`, `test-resources/` folders in a given project.

@frenchy64 recently rationalized our [properties management](https://github.com/threatgrid/ctia/blob/ac300dfa9a86cb61f49e8a490da0982e7164f234/test/ctia/test_helpers/core.clj#L199) in tests. We also have some [ES fixtures properties](https://github.com/threatgrid/ctia/blob/ac300dfa9a86cb61f49e8a490da0982e7164f234/test/ctia/test_helpers/es.clj#L90) that are also used to override some properties. However we could consider this suggestion for future refactoring about the test properties.

vemv: Thanks for the explanation! SGTM :)

Maybe I interpreted `I let them commented to ease local tests` incorrectly.

>  I should simply remove them from default.properties. The migration doc is sufficiently detailed about it.

Yes this sounds like a win, since commented-out code can sometimes bring more questions than anyting else. Maybe as a comment inside the properties file, I'd link to https://github.com/threatgrid/ctia/blob/master/doc/migration.md#migration-properties

test/ctia/task/migration/migrations/describe_test.clj
ereteog: [what about that?](https://github.com/threatgrid/ctia/pull/1065/commits/0862d62bc9b9d6c58a4066a3d30e4ca5317b848f)

I replaced the random generation with a deterministic generation of all mutations in order to avoid ending in a flaky test with the probability (even low) to have a missing mutant and thus a failing `assert`.

vemv: Bravo 🙌

TIL about `subsets`!

Checking out a sample run:

```clj
> (clojure.math.combinatorics/subsets [1 2 3])
(() (1) (2) (3) (1 2) (1 3) (2 3) (1 2 3))
```

It makes me wonder if ordering matters? e.g one might want to exercise both `x y z` and `z y x`

ereteog: here it's subsets of fields that will be dissoced so the order does not matter, but yes it should be kept in mind for some fonctions.

gbuisson (approved):
LGTM

hisans:
“malware” entity did not treat the “title”, “description” and “short_description” fields as mandatory fields. Dev to Fix.

hisans:
QA Verified in 1.66 Release

PR #1064, opened 2021-01-20
Add asset entities to graphql route #4521

agzam:
> **Epic** #
> Close 
> Related https://github.com/threatgrid/iroh/issues/4521

*Note: this is a multi-PR work. The second part is [here #1068 ](https://github.com/threatgrid/ctia/pull/1068)*

<a name="qa">[§](#qa)</a> QA
============================

1. Create one or multiple Assets by issuing requests to `POST ​/ctia​/asset` route
2. Try running simple graphql queries, e.g.: 

```
POST  /ctia/graphql
{
"query": "query Assets { assets { nodes {id} }}"
}
```

3. Make sure the results include all (previously created) assets
4. Similarly, test it for `AssetMapping`, `AssetProperties` and `TargetRecord`

<a name="release-notes">[§](#release-notes)</a> Release Notes
=============================================================

```
intern: graphql queries for Asset, AssetMapping, AssetProperties and TargetRecord entities
public: one line description for public release notes.
```

<a name="squashed-commits">[§](#squashed-commits)</a> Squashed Commits
======================================================================

2d651cc4 adds an asset-mapping specific test case
95357042 remove unused code
43710f3f adds target-record graphql tests
b97bc4f7 adds target-record graphql schema
6581d90c remove singular form of asset-properties
2b084878 adds AssetProperties tests
0ef440bf normalize ns headers
3785a4e9 adds asset-properties graphql schema
609fe710 adds basic asset-mapping graphql test
16bc4277 Adds graphql schemas for asset-mapping
214f34c1 extend asset graphql tests
883b3e3f move asset graphql schemas to its own namespace
657f7363 adds additional graphql schema things
7254460c adds AssetType to graphql/schemas
dd0f4d03 normalize and sort ns headers

PR #1073, opened 2021-02-10
Remove disabled entities from GraphQl route

agzam:
Entities marked as disabled in the config no longer can be used in graphql queries

> **Epic** #
> Close https://github.com/threatgrid/iroh/issues/4522
> Related #1020, #1064

<a name="qa">[§](#qa)</a> QA
============================

1. Test on "public intel", where Asset, AssetMapping, AssetProperties, TargetRecord are [disabled in the config](https://github.com/threatgrid/tenzin-config/blob/master/ctia/TEST/us-east-1/ctia.properties#L101)
2. Run the app, open up the Swagger console. 
3. Make sure that no contexts shown for disabled entities (routes for Asset, AssetMapping, AssetProperties, TargetRecord should not be available)
4. Try to post graphql queries including the disabled entities, e.g.:
```
POST  /ctia/graphql
{
"query": "query Assets { assets { nodes {id} }}"
}
```
5.  Expected: The system should fail with the validation error (since the schema does not include Assets)

25390b27 test: add plural forms check
9626566a PR feedback
93976be1 proper plurals
014db209 removes disabled entities from GraphQL
d0de05c8 adds second arity to FeaturesService/enabled?
57b2c4a3 FeaturesService/enable? should ignore non-existing entities

### DIFF
--- a/src/ctia/features_service.clj
+++ b/src/ctia/features_service.clj
@@ -13,8 +13,11 @@
   would not be visible in Swagger console."
   (feature-flags [this]
     "Returns all feature flags defined in the config")
-  (enabled? [this key]
-    "Returns `true` unless entity key is marked as Disabled in properties config"))
+  (enabled?
+    [this]
+    [this key]
+    "When called with no provided key - returns all entities except those that are explicitly disabled in properties config.
+     If the entity key argument used, it returns `true` unless the entity marked as disabled in properties config."))
 
 (tk/defservice features-service
   FeaturesService
@@ -33,4 +36,9 @@
        (map keyword x)
        (set x)
        (contains? x key)
-       (not x)))))
+       (not x))))
+  (enabled?
+   [this]
+   (->> (entities/all-entities)
+        keys
+        (filter (partial enabled? this)))))
--- a/src/ctia/features_service.clj
+++ b/src/ctia/features_service.clj
@@ -1,6 +1,8 @@
 (ns ctia.features-service
-  (:require [puppetlabs.trapperkeeper.core :as tk]
-            [clojure.string :as string]))
+  (:require
+   [clojure.string :as string]
+   [ctia.entity.entities :as entities]
+   [puppetlabs.trapperkeeper.core :as tk]))
 
 (defprotocol FeaturesService
   "Service to read configuration properties under [:ctia :features] key in config.
@@ -21,10 +23,14 @@
     (get-in-config [:ctia :features]))
   (enabled?
    [this key]
-   (as-> (get-in-config [:ctia :features :disable]) x
-     (str x)
-     (string/split x #",")
-     (map keyword x)
-     (set x)
-     (contains? x key)
-     (not x))))
+   (when (-> (entities/all-entities)
+             keys
+             set
+             (contains? key))
+     (as-> (get-in-config [:ctia :features :disable]) x
+       (str x)
+       (string/split x #",")
+       (map keyword x)
+       (set x)
+       (contains? x key)
+       (not x)))))
--- a/src/ctia/features_service.clj
+++ b/src/ctia/features_service.clj
@@ -25,6 +25,11 @@
   (feature-flags [this]
     (get-in-config [:ctia :features]))
   (enabled?
+   [this]
+   (->> (entities/all-entities)
+        keys
+        (filter (partial enabled? this))))
+  (enabled?
    [this key]
    (when (-> (entities/all-entities)
              keys
@@ -36,9 +41,4 @@
        (map keyword x)
        (set x)
        (contains? x key)
-       (not x))))
-  (enabled?
-   [this]
-   (->> (entities/all-entities)
-        keys
-        (filter (partial enabled? this)))))
+       (not x)))))
--- a/src/ctia/features_service.clj
+++ b/src/ctia/features_service.clj
@@ -13,22 +13,14 @@
   would not be visible in Swagger console."
   (feature-flags [this]
     "Returns all feature flags defined in the config")
-  (enabled?
-    [this]
-    [this key]
-    "When called with no provided key - returns all entities except those that are explicitly disabled in properties config.
-     If the entity key argument used, it returns `true` unless the entity marked as disabled in properties config."))
+  (enabled? [this key]
+    "Returns `true` unless entity key is marked as Disabled in properties config"))
 
 (tk/defservice features-service
   FeaturesService
   [[:ConfigService get-in-config]]
   (feature-flags [this]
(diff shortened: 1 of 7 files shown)


ereteog:
@agzam FYI
For the QA section, they generally use the configured public and private instances. I sync with QA for previous "disable entities" and it's not handy to launch a dedicated server. Thus, in TEST, I [disabled asset entities on the private intel](https://github.com/threatgrid/tenzin-config/blob/master/private-ctia/TEST/us-east-1/ctia.properties) and [enabled it on the public intel](https://github.com/threatgrid/tenzin-config/blob/ac85f62eb33b58382ddae8c0f6e6afae7e9b539f/ctia/TEST/us-east-1/ctia.properties#L101) (indeed, for now, it does not make sense to have assets on public intel). So basically to test they will try that the entities are properly hidden on public intel in TEST env.

src/ctia/graphql/schemas.clj
```diff
@@ -42,137 +44,167 @@
...
+(defn- remove-disabled
+  "Removes GraphQl fields of keys of disabled entities."
+  [services graphql-fields]
+  (let [+plurals (reduce
+                  (fn [a n] (let [k      (dash->underscore n)
+                                  plural (-> k name (str "s") keyword)]
+                              (into a [k plural])))
+                  []
+                  (disabled-entities services))]
+    (apply dissoc graphql-fields +plurals)))
```
ereteog, line 191: considering my previous comment, I would write this function like this:
``` clojure
(->> (disabled-entities services)
     (mapcat (fn [{:keys [entity plural]}] [entity plural]))]
     (apply dissoc graphql-fields))
```

test/ctia/features_service_test.clj
```diff
@@ -11,11 +13,18 @@
...
        (fn [app]
-         (let [{:keys [disabled? enabled?]} (th/get-service-map app :FeaturesService)]
+         (let [{:keys [enabled?]} (th/get-service-map app :FeaturesService)]
            (testing "Asset, Actor and Sighting are disabled"
              (is (not-any? enabled? [:asset :actor :sighting])))
            (testing "Incident, Indicator entities are enabled"
-             (is (every? enabled? [:incident :indicator])))))))))
+             (is (every? enabled? [:incident :indicator])))
+           (testing "It should not return `true` for non-existing entity keys"
+             (is (not (enabled? :lorem-ipsum))))
```
ereteog, line 22: I would ensure that's it is a boolean here
```suggestion
             (is (false? (enabled? :lorem-ipsum))))
```

src/ctia/graphql/schemas.clj
```diff
@@ -42,137 +44,167 @@
...
+  (-> k
+      name
+      (string/replace "-" "_")
+      keyword))
+
+(defn- disabled-entities
+  [{{:keys [enabled?]} :FeaturesService}]
+  (->> (entities/all-entities)
+       keys
+       (filter #(not (enabled? %)))))
```
ereteog, line 180: I would transfer this function to the FeatureService.

src/ctia/graphql/schemas.clj
ereteog: A variant with `enbaled-entities` could be more elegant abd easier to reason about:
``` clojure
(->> (enabled-entities services)
     (mapcat (fn [{:keys [entity plural]}] [entity plural]))]
     (select-keys graphql-fields))
```

ereteog (commented):
Nice work. There is a bug on the plural management.
My comments could be summarized by:
In FeatureService, preserve the original `enabled?`, add `enabled-entities` and use the plural field of entities.
Could you also consider my first comment to improve the QA section?

src/ctia/schemas/core.clj
```diff
@@ -22,26 +22,26 @@
...
+                                                     :apply-hooks}))
+   :StoreService                    {:get-store GetStoreFn}
+   :IAuth                           {:identity-for-token (s/=> s/Any s/Any)}
+   :GraphQLNamedTypeRegistryService {:get-or-update-named-type-registry
+                                     (s/=> graphql.schema.GraphQLType
+                                           s/Str
+                                           (s/=> graphql.schema.GraphQLType))}
+   :IEncryption                     {:encrypt (s/=> s/Any s/Any)
+                                     :decrypt (s/=> s/Any s/Any)}
+   :FeaturesService                 {:enabled?      (s/=> s/Keyword s/Bool)
```
ereteog, line 43: former syntax was right :
http://plumatic.github.io/schema/schema.core.html#var-.3D.3E

src/ctia/schemas/core.clj
```diff
@@ -325,6 +325,6 @@
   "vocab.observable-type-id")
 
 (s/defschema GetEntitiesServices
-  {:FeaturesService {:enabled? (s/=> s/Bool s/Keyword)
+  {:FeaturesService {:enabled? (s/=> s/Keyword s/Bool)
```
ereteog, line 328: same remark

test/ctia/features_service_test.clj
```diff
@@ -11,11 +13,18 @@
...
+           (testing "It should return all enabled entities when no key argument is used"
+             (is (= #{:asset :actor :sighting}
+                    (set/difference
+                     (-> (entities/all-entities) keys set)
+                     (set (enabled?))))))))))))
```
ereteog, line 27: you need to remove this test

ereteog (commented):
quick feedback.

ereteog (approved):
LGTM

src/ctia/graphql/schemas.clj
agzam: I thought about that, but then decided not to do it at this point (requires changing schemas, etc.), maybe some other time if we see something like that needed. I'll keep in mind that we have it here and move it then.

src/ctia/features_service.clj
```diff
@@ -11,20 +13,32 @@
   would not be visible in Swagger console."
   (feature-flags [this]
     "Returns all feature flags defined in the config")
-  (enabled? [this key]
-    "Returns `true` unless entity key is marked as Disabled in properties config"))
+  (enabled?
+    [this]
+    [this key]
+    "When called with no provided key - returns all entities except those that are explicitly disabled in properties config.
+     If the entity key argument used, it returns `true` unless the entity marked as disabled in properties config."))
```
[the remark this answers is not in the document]
agzam: Yeah, my doodling around this ns turned out to be just that. Sorry for the noise.

hisans:
QA Verified in 1.67 Release

PR #1074, opened 2021-02-12
Display schema error for the text/plain content type

msprunck:
> Close threatgrid/iroh#3151

- Defines a response format encoder for the `text/plain` content type.
- Set the `accept` header to `text/plain` if the request `uri` ends with `.txt`

If a request is made with the `text/plain` content type and a wrong request parameter instead of returning a 500 with an undefined exception `{"type":"unknown-exception","class":"clojure.lang.ExceptionInfo"}` it will return a 400 with the stringified schema error `{:errors {:s missing-required-key}}` as string.

Note for reviewers: Enable the `Hide whitespace changes` option

<a name="qa">[§](#qa)</a> QA
============================

1. Send a GET to /view.txt, but instead of using s as the query parameter use q, like:
`curl -X GET "https://private.intel.test.iroh.site/ctia/feed/https%3A%2F%2Fprivate.intel.test.iroh.site%3A443%2Fctia%2Ffeed%2Ffeed-2d3b363a-5740-43b9-8f05-91c062e68bb0/view.txt?q=9d1601de-9b64-4a17-aeca-5aa56b4b9792" -H "authorization: JWT"`

2. It should return a 400 with the following text `{:errors {:s missing-required-key}}`

```
intern: Display schema error for the text/plain content type
```

### DIFF
--- a/src/ctia/http/handler.clj
+++ b/src/ctia/http/handler.clj
@@ -197,7 +197,11 @@
    See the `ring.middleware.format-response/wrap-restful-response`
    middleware used by compojure API for more details"
   []
-  (make-encoder (fn [body] (str body)) "text/plain"))
+  (make-encoder (fn text-plain-encoder [body]
+                  (if (string? body)
+                    body
+                    (pr-str body)))
+                "text/plain"))
 
 (s/defn api-handler [{{:keys [get-in-config]} :ConfigService
                       :as services} :- APIHandlerServices]
--- a/src/ctia/http/handler.clj
+++ b/src/ctia/http/handler.clj
@@ -31,7 +31,10 @@
             [ctia.status.routes :refer [status-routes]]
             [ring.middleware.not-modified :refer [wrap-not-modified]]
             [ring.util.http-response :refer [ok]]
-            [schema.core :as s]))
+            [schema.core :as s]
+            [compojure.api.middleware :refer [api-middleware-defaults]]
+            [ring.middleware.format-response :refer [make-encoder]]
+            [compojure.api.middleware :refer [->mime-types]]))
 
 (def api-description
   "A Threat Intelligence API service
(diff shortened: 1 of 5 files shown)


src/ctia/http/server.clj
```diff
@@ -122,6 +122,17 @@
...
                            (into headers response-headers)))))))
 
+(defn wrap-txt-accept-header
+  "Enforces the `accept` request header to `text/plain` when the
+   uri ends with `.txt`. Mainly used by the `/feed/{id}/view.txt` endpoint"
+  [handler]
+  (fn [{:keys [uri] :as request}]
+    (let [new-request
+          (cond-> request
+            (string/ends-with? uri ".txt")
```
vemv, line 132: Probably not all endpoints support .txt, but this middleware would suggest so.

i.e. a client shouldn't be able to dictate what a server accepts? (or at least not unilaterally)

(perhaps I'd try to code an integration test showing why this is needed)

msprunck: The `.txt` suffix is only supported if the route is defined, otherwise a 404 is returned.

This middleware is here to set default mime type accepted by the client if the `.txt` extension is used at the end of the request URI. If not set, the response encoder for json is used by default. It comes from the `wrap-restful-response` middleware that takes the first format defined in its options (https://github.com/metosin/compojure-api/blob/8b5122ad308f3ed17c452bbfc3cb4e3ea2cae84c/src/compojure/api/middleware.clj#L154).

The integration test is here https://github.com/threatgrid/ctia/pull/1074/files?diff=split&w=1#diff-c0821316e41b318d4f743b7391b511c7ab557d55f3b61dc643b1ca1767cd1949R159. I forgot to push the tests :)

src/ctia/http/handler.clj
```diff
@@ -188,11 +191,23 @@
   (cond-> entity-map
     (not (enabled? entity)) (assoc :no-api? true)))
 
+(defn make-text-plain-format-encoder
+  "Make a simple format encoder for the text/plain
+   mime-type. It stringifies the response body.
+   See the `ring.middleware.format-response/wrap-restful-response`
+   middleware used by compojure API for more details"
+  []
+  (make-encoder (fn [body] (str body)) "text/plain"))
```
ereteog, line 200: I never remember why, but I think that @yogsototh would propose this:
```suggestion
  (make-encoder (fn [body] (pr-str body)) "text/plain"))
```

ereteog (approved):
LGTM!

Perhaps the QA should exercise it with multiple Accept headers to reflect that we ignore it whatever is the value for `view.txt`

src/ctia/http/server.clj
vemv: > The .txt suffix is only supported if the route is defined, otherwise a 404 is returned.

Good point. Probably I was thinking of Rails where clients can add .json/.txt/.xml/... to _any_ route and will affect content negotiation.

agzam (approved):

gbuisson (approved):
LGTM!

hisans:
QA Verified in 1.67 Release

PR #1075, opened 2021-02-12
don't throw an arity exception calling bundle export with an empty seq

gbuisson:
> **Epic** #
> Close https://github.com/threatgrid/iroh/issues/4768
> Related #

Calling the Bundle export route with an empty sequence currently returns an ArityException
This PR fixes that behavior providing a val argument to a reduce.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: hardened the Bundle Export route
```

### DIFF
--- a/src/ctia/bundle/core.clj
+++ b/src/ctia/bundle/core.clj
@@ -451,6 +451,8 @@
    ident
    params
    services :- APIHandlerServices]
-  (->> (map #(export-entities % identity-map ident params services) ids)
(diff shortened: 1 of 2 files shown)


vemv (approved):
lgtm!

ereteog (approved):
LGTM
However, shall we not simply reject empty ids?

gbuisson:
@ereteog I don't have a preference, I would rather not change the contract even though it never worked with empty Ids.

PR #1077, opened 2021-02-16
Issue-4744: Search vulnerabilities by CPEs

halcyon:
Progress toward #4744

> **Epic** https://github.com/threatgrid/iroh/issues/4744

> Related https://github.com/threatgrid/iroh/issues/4789

Add route for finding vulnerabilities associated with specific combinations of CPEs (Common Platform Enumerations).

Proposed route:
GET /ctia/vulnerability/search/cpes?cpes-string=space delimited list of cpes

<a name="qa">[§](#qa)</a> QA
============================

```
```

### DIFF
--- a/src/ctia/entity/vulnerability.clj
+++ b/src/ctia/entity/vulnerability.clj
@@ -1,10 +1,11 @@
 (ns ctia.entity.vulnerability
   (:require
+   [ctia.lib.compojure.api.core :refer [context GET routes]]
    [flanders.utils :as fu]
    [ctia.entity.feedback.graphql-schemas :as feedback]
    [ctia.entity.relationship.graphql-schemas :as relationship]
    [ctia.entity.vulnerability.mapping :refer [vulnerability-mapping]]
-   [ctia.domain.entities :refer [default-realize-fn]]
+   [ctia.domain.entities :as ent]
    [ctia.schemas.graphql
     [sorting :as graphql-sorting]
     [flanders :as flanders]
@@ -23,7 +24,11 @@
    [schema-tools.core :as st]
    [schema.core :as s]
    [ctia.schemas.sorting :as sorting]
-   [ctia.schemas.graphql.ownership :as go]))
+   [ctia.schemas.graphql.ownership :as go]
+   [ring.swagger.schema :as swagger-schema]
+   [ctia.entity.vulnerability.cpe :as cpe]
+   [ctia.store :as ctia-store]
+   [clojure.string :as str]))
 
 (def-acl-schema Vulnerability
   ws/Vulnerability
@@ -46,13 +51,14 @@
   (csu/optional-keys-schema StoredVulnerability))
 
 (def realize-vulnerability
-  (default-realize-fn "vulnerability" NewVulnerability StoredVulnerability))
+  (ent/default-realize-fn "vulnerability" NewVulnerability StoredVulnerability))
 
 (def-es-store VulnerabilityStore :vulnerability StoredVulnerability PartialStoredVulnerability)
 
 (def vulnerability-fields
   (concat sorting/base-entity-sort-fields
           sorting/sourcable-entity-sort-fields
+          sorting/describable-entity-sort-fields
           []))
 
 (def vulnerability-sort-fields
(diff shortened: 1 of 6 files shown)


src/ctia/entity/vulnerability/cpe.clj
```diff
@@ -0,0 +1,53 @@
...
+(defn query-vulnerability
+  [cpes {id :id {:keys [nodes]} :configurations}]
+  (when (seq (keep (fn [node]
+                     (query-node node cpes))
+                   nodes))
+    id))
+
+(defn ids
+  [cpes vulnerabilities]
+  (reduce (fn ids-rf
```
ereteog, line 47: If I am not mistaken this reduce is a [`keep`](https://clojuredocs.org/clojure.core/keep), isn't it?

ereteog (commented):
I know that's a wip but I was curious :-)
About `list-all-pages` this put the server at risk, you need to find a way to paginate.
In addition, we have a ongoing [task about improving the CTIA documentation](https://github.com/threatgrid/iroh/issues/4574), could you add a developer documentation explaining the process under this helper

src/ctia/entity/vulnerability/cpe.clj
halcyon: Yes it is - thank you.

src/ctia/entity/vulnerability.clj
```diff
@@ -109,29 +115,69 @@
...
+        :query [{cpe-ids-string :cpe_ids :as params}
+                (st/merge VulnerabilitySearchParams {:cpe_ids (swagger-schema/describe s/Str "CPE ids to search by")})]
+        :summary "List Vulnerabilities by CPE ids"
+        :description (routes.common/capabilities->description capabilities)
+        :capabilities capabilities
+        :auth-identity identity
+        :identity-map identity-map
+        (let [cpe-ids (str/split cpe-ids-string #"\s+")
+              vulnerability-ids
+              (loop [query-params params
```
ereteog, line 133: This loop is exactly what does `list-all-pages` and the reason of OOM. If we do this we do not control anymore the volume of data loaded in memory in the server for this request. 
You need to expose pagination parameters and let the client paginate.

halcyon: Thank you for looking at this. Please consider the following and let me know your thoughts:

- There are two queries being used within this endpoint 
- The first query which is within the loop, is using server side paging, and processing each page individually
- The purpose of the loop, is to construct a vector of IDs which are used to construct the second query.
- This is different from `list-all-pages` in that each page is being processed individually by `cpe/vulnerabilities->ids` to extract only the ids of vulnerabilities that have matching configurations.
- Unlike `list-all-pages` which accumulates all of the query responses into memory, this loop only accumulating the `ids` of vulnerabilities with matching configurations.
- It is frequently the case that the first query in the loop will not find any vulnerability ids with matching configurations in the first several pages. Were we to use client side paging in the first query, this could cause the second query to incorrectly not return anything to the user.
- The second query employs client side paging

ereteog: I understand your point of view and made my previous remark with this understanding of the algorithm. The issue is to ensure the protection of the server over potentially unexpected cases (e.g. unexpected input, unexpected state of the vulnerability store, high number of concurrent queries).
I propose that you use a [lazyseq](https://clojuredocs.org/clojure.core/lazy-seq) like [here](https://github.com/threatgrid/ctia/blob/master/src/ctia/task/migration/migrate_es_stores.clj#L126). 
It would be a simple replacement of your `loop`:
```clojure

(defn list-vulnerability-ids
    [query-params]
    (lazy-seq
	 (let [{:keys [data paging]}
                 (ctia-store/list-records (get-store :vulnerability)
                                               {:one-of {:configurations.nodes.cpe_match.cpe23Uri cpe-ids
                                                         :configurations.nodes.children.cpe_match.cpe23Uri cpe-ids}}
                                               identity-map
                                               (merge query-params {:fields [:configurations.nodes :id]}))
                      ids (cpe/vulnerabilities->ids cpe-ids data)]
                  (if-let [next-params (:next paging)]
                     (concat ids (list-vulnerability-ids next-params))
                     ids))))
```
It should be tested, but the pagination should be done with the parameters of this function, right?
Since it's a little bit tricky, you should carefully check this pagination.

halcyon: I do like the direction you are going in with a lazy sequence of ids. Would the proposed implementation using `concat` be vulnerable to [Clojure Don’ts: Concat](https://stuartsierra.com/2015/04/26/clojure-donts-concat)?

ereteog: > Would the proposed implementation using `concat` be vulnerable to [Clojure Don’ts: Concat](https://stuartsierra.com/2015/04/26/clojure-donts-concat)?

That was a quick proposition, but I think that `into` would be more appropriate. There are many traps that force the realization of a lazy collection, that's why you should properly test this lazyness.
I propose to invoke some clojure god and ask if @frenchy64 could help us to properly do it :-).

frenchy64: FWIW I think lazy-seq + concat is safe. I believe the blog is talking about loop (or any strict op) + concat. IIUC the stackoverflow happens when you call `first` on a large stack of `concat`s. I don't think this is possible with `lazy-seq`, but is easy to reproduce with reduce or loop.

halcyon:
The feedback from this PR has been incorporated into PR #1099 and will be closed.

PR #1078, opened 2021-02-20
Asset scope capabilities #4677

agzam:
Update the JWT authentication layer to support Assets.

> **Epic** #
> Close https://github.com/threatgrid/iroh/issues/4677
> Related #

<a name="qa">[§](#qa)</a> QA
============================
TODO: Add QA instructions

```
```

### DIFF
--- a/src/ctia/auth/jwt.clj
+++ b/src/ctia/auth/jwt.clj
@@ -63,9 +63,7 @@
                               "casebook"))
 
 (defn assets-root-scope [get-in-config]
-  (if (= "private-intel" (entity-root-scope get-in-config))
-    "no-asset-intel"
-    (get-in-config [:ctia :auth :assets :scope] "asset-intel")))
+  (get-in-config [:ctia :auth :assets :scope] "asset-intel"))
 
 (defn claim-prefix [get-in-config]
   (get-in-config [:ctia :http :jwt :claim-prefix]
--- a/src/ctia/auth/jwt.clj
+++ b/src/ctia/auth/jwt.clj
@@ -113,14 +113,12 @@
 (defn gen-assets-capabilities
   "Generate capabilities for the root-scope 'asset-intel'."
   [scope-repr]
-  (if (= "asset-intel" (first (:path scope-repr)))
-    (->> [:asset :asset-mapping :asset-properties :target-record]
-         (select-keys (all-entities))
-         vals
-         (map #(gen-capabilities-for-entity-and-accesses
-                % (:access scope-repr)))
-         unionize)
-    #{}))
+  (->> [:asset :asset-mapping :asset-properties :target-record]
+       (select-keys (all-entities))
+       vals
+       (map #(gen-capabilities-for-entity-and-accesses
+              % (:access scope-repr)))
+       unionize))
 
 (defn scope-to-capabilities
   "given a scope generate capabilities"
--- a/src/ctia/auth/jwt.clj
+++ b/src/ctia/auth/jwt.clj
@@ -1,18 +1,17 @@
 (ns ctia.auth.jwt
   (:refer-clojure :exclude [identity])
-  (:require [cheshire.core :as json]
-            [clj-jwt.key :refer [public-key]]
-            [clj-momo.lib.set :refer [as-set]]
-            [clojure.set :as set]
-            [clojure.string :as string]
-            [clojure.tools.logging :as log]
-            [ctia.auth :as auth :refer [IIdentity]]
-            [ctia.auth.capabilities
-             :refer
-             [all-entities gen-capabilities-for-entity-and-accesses]]
-            [ctia.properties :as p]
-            [ring.util.http-response :as resp]
-            [scopula.core :as scopula]))
+  (:require
+   [cheshire.core :as json]
+   [clj-jwt.key :refer [public-key]]
+   [clj-momo.lib.set :refer [as-set]]
+   [clojure.set :as set]
+   [clojure.string :as string]
+   [clojure.tools.logging :as log]
+   [ctia.auth :as auth :refer [IIdentity]]
+   [ctia.auth.capabilities :refer
+    [all-entities gen-capabilities-for-entity-and-accesses]]
+   [ring.util.http-response :as resp]
+   [scopula.core :as scopula]))
 
 (defn load-public-key
   [path]
@@ -64,7 +63,9 @@
                               "casebook"))
 
 (defn assets-root-scope [get-in-config]
-  (get-in-config [:ctia :auth :assets :scope] "asset-intel"))
+  (if (= "private-intel" (entity-root-scope get-in-config))
+    "no-asset-intel"
+    (get-in-config [:ctia :auth :assets :scope] "asset-intel")))
 
 (defn claim-prefix [get-in-config]
   (get-in-config [:ctia :http :jwt :claim-prefix]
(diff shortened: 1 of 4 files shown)


src/ctia/auth/jwt.clj
```diff
@@ -72,6 +71,11 @@
   (get-in-config [:ctia :auth :casebook :scope]
                               "casebook"))
 
+(defn assets-root-scope [get-in-config]
+  (if (= "private-intel" (entity-root-scope get-in-config))
```
gbuisson, line 75: I'm not sure I understand why you require this conditional statement, IMO it should be the same code as `casebook-root-scope` but for `asset`

src/ctia/auth/jwt.clj
```diff
@@ -117,13 +121,26 @@
    (:casebook (all-entities))
    (:access scope-repr)))
 
+(defn gen-assets-capabilities
+  "Generate capabilities for the root-scope 'asset-intel'."
+  [scope-repr]
+  (if (= "asset-intel" (first (:path scope-repr)))
```
gbuisson, line 127: I *think* you could loose the `if` statement

src/ctia/auth/jwt.clj
gbuisson, line 75: I think you may probably loose this `if` statement, make it exactly like `casebook-root-scope` but for `asset`

gbuisson (commented):
I think you may also add an entry to ctia-default.properties,
I think besides my limited remarks it looks ok so far.

src/ctia/auth/jwt.clj
agzam: Hmm. As you stated yourself in the https://github.com/threatgrid/iroh/issues/4677

> In a second step, exclude all Assets capabilities from the private-intel scope.

So I thought: to have asset capabilities one has to have both "non-private-intel" scope and "asset-intel" scope. No?

src/ctia/auth/jwt.clj
agzam: It is there for the same reason as in the above comment https://github.com/threatgrid/ctia/pull/1078#discussion_r581281725. Without it, it will make assets available when only one of the required scopes present.

src/ctia/auth/jwt.clj
gbuisson: haha no, for example one could have both the `private-intel` scope and the `asset` scope and be allowed to access everything at once, it's likely what's going to happen anyway.

agzam: Okay, I'm revisiting the PR and I think I'm still confused. 

My current understanding is:
- if user has `asset-intel` scope  - they should have all asset enabled capabilities
- if user has `private-intel` scope, but doesn't have `asset-intel` scope - they shouldn't have access to asset enable capabilities

Which simply seems to be boiling down to: 
- if user doesn't have `asset-intel` scope - they shouldn't have access to asset capabilities
- `asset-intel` scope enabled by default
- `asset-intel` scope is **not**  enabled by default if `private-intel` scope is present, for a user with `private-intel` scope to have access to asset capabilities, they must include `asset-intel` scope explicitly.

Does that look accurate? Or am I still confused?

@ereteog

ereteog: >     * if user has `asset-intel` scope  - they should have all asset enabled capabilities
> 
>     * if user has `private-intel` scope, but doesn't have `asset-intel` scope - they shouldn't have access to asset enable capabilities

Exactly


> Which simply seems to be boiling down to:
> 
>     * if user doesn't have `asset-intel` scope - they shouldn't have access to asset capabilities
> 
>     * `asset-intel` scope enabled by default
> 
>     * `asset-intel` scope is **not**  enabled by default if `private-intel` scope is present, for a user with `private-intel` scope to have access to asset capabilities, they must include `asset-intel` scope explicitly.
> 

I am not sure that 3 bullet points is a summary of 2 bullet points :-).

Let's keep it simple: asset related routes require `asset-intel` scope.
==>
```clojure
(defn assets-root-scope [get-in-config]
    (get-in-config [:ctia :auth :assets :scope] "asset-intel"))

File: src/ctia/auth/jwt.clj (hunk 2 above)
Comment (ereteog) [reply src/ctia/auth/jwt.clj:127]: If our answers to comment https://github.com/threatgrid/ctia/pull/1078#discussion_r581281725 are clear enough, you can now loose the `if` statement :-)

**Review by ereteog — COMMENTED** [review]
I answered to ongoing questions. 
Let me know if it's not clear, and if you have any others reach out to me in DM to clarify any other doubts in order to finish this PRs..

File: src/ctia/auth/jwt.clj (hunk 1 above)
Comment (agzam) [reply src/ctia/auth/jwt.clj]: Addressed in a410e79162a6a3cf5d7836261d8d883485269977

Beware though. We're going to have to explicitly disable asset root scope, if we don't want to expose Asset capabilities.

File: test/ctia/auth/jwt_test.clj [hunk 3]
```diff
@@ -38,66 +39,99 @@
...
+                    :create-target-record :delete-asset :delete-asset-mapping
+                    :delete-asset-properties :delete-target-record :list-asset-mappings
+                    :list-asset-properties :list-assets :list-target-records :read-asset
+                    :read-asset-mapping :read-asset-properties :read-target-record :search-asset
+                    :search-asset-mapping :search-asset-properties :search-target-record}
+                  (sut/scope-to-capabilities (sut/assets-root-scope get-in-config) get-in-config))
+               "By default asset capabilities enabled on global-intel scope")))))
+    (testing "Asset capabilities explicitly disabled"
+      (helpers/with-properties ["ctia.auth.entities.scope" "global-intel"
+                                "ctia.auth.assets.scope" "no-asset-intel"]
```
ereteog, line 60: This configuration is only here to specify the value of the scope that provides asset capabilities, "asset-intel" being the default value for such a scope. So such a conf would specify that the scope accepted to have asset capabilities would be `no-asset-intel`

ereteog (commented):
I think that you still need to properly understand how scopes works and how they are turned into capabilities in CTIA. 
While the scopes are decided and provided by IROH, CTIA is mostly ignorant from IROH and only sees JWTs. This is the conf that enables CTIA to interpret the scopes of the JWT and to turn it into CTIA capabilities. So if tomorrow we want to enable the access to asset capabilities with the scope "Ag-loves-emacs", we would only have to update the conf with: 
`"ctia.auth.assets.scope" "Ag-loves-emacs"`
If you still confused, do not hesitate to reach out to me when you are up, we can do a quick webex to definitively clarify it.

test/ctia/auth/jwt_test.clj
agzam: I tried addressing your comments in https://github.com/threatgrid/ctia/pull/1078/commits/fc313a228753acdf01f680f85bdcb7318fc9d065

And Guillaume's in https://github.com/threatgrid/ctia/pull/1078/commits/a76efff22ee5110ed1ac73fa60dbffc68b9468f4

Would that do? Let me know if I missed anything else.

ereteog (approved):
LGTM

ereteog:
@agzam did you synced with QA to test it? You must add the QA section for the release, that's a highly important step.

hisans:
As discussed with Ag Ibragimov, This can be verified once the https://github.com/threatgrid/iroh/issues/4680 is fixed.

ereteog:
> As discussed with Ag Ibragimov, This can be verified once the [threatgrid/iroh#4680](https://github.com/threatgrid/iroh/issues/4680) is fixed.

@hisans this is ready for next release: https://github.com/threatgrid/tenzin-config/pull/407

PR #1083, opened 2021-03-05
JWT middleware should respect expected content type

frenchy64:
Close https://github.com/threatgrid/iroh/issues/4541

Reviewers: please hide whitespace changes, a nested let is merged.

The ring-jwt-middleware error handler was hardcoded to return JSON. A call to `wrap-restful-response` after jwt parsing lets us do the right thing here.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: JWT middleware should respect expected content type
```

### DIFF
--- a/src/ctia/auth/jwt.clj
+++ b/src/ctia/auth/jwt.clj
@@ -55,14 +55,15 @@
              (throw (ex-info err-msg {:jwt-pubkey-map txt})))))))
 
 
-(defn jwt-error-handler
+(defn ->jwt-error-handler
   "Return an `unauthorized` HTTP response
   and log the error along debug infos"
-  [error-msg infos]
-  (let [err {:error :invalid_jwt
-             :error_description error-msg}]
-    (log/info error-msg (pr-str (into infos err)))
-    (resp/unauthorized (json/generate-string err))))
+  [request]
+  (fn [error-msg infos]
+    (let [err {:error :invalid_jwt
+               :error_description error-msg}]
+      (log/info error-msg (pr-str (into infos err)))
+      (resp/unauthorized (json/generate-string err)))))
 
 (defn entity-root-scope [get-in-config]
   (get-in-config [:ctia :auth :entities :scope]
--- a/src/ctia/auth/jwt.clj
+++ b/src/ctia/auth/jwt.clj
@@ -11,6 +11,7 @@
              :refer
              [all-entities gen-capabilities-for-entity-and-accesses]]
             [ctia.properties :as p]
+            [ring-jwt-middleware.core :as rjwt]
             [ring.util.http-response :as resp]
             [scopula.core :as scopula]))
 
@@ -54,16 +55,6 @@
              (log/error err-msg)
              (throw (ex-info err-msg {:jwt-pubkey-map txt})))))))
 
-
-(defn ->jwt-error-handler
-  "Return an `unauthorized` HTTP response
-  and log the error along debug infos"
-  [request]
-  (fn [error-msg infos]
-    (let [err {:error :invalid_jwt
-               :error_description error-msg}]
-      (log/info error-msg (pr-str (into infos err)))
-      (resp/unauthorized (json/generate-string err)))))
 
 (defn entity-root-scope [get-in-config]
   (get-in-config [:ctia :auth :entities :scope]
--- a/src/ctia/auth/jwt.clj
+++ b/src/ctia/auth/jwt.clj
@@ -11,7 +11,6 @@
              :refer
              [all-entities gen-capabilities-for-entity-and-accesses]]
             [ctia.properties :as p]
-            [ring-jwt-middleware.core :as rjwt]
             [ring.util.http-response :as resp]
             [scopula.core :as scopula]))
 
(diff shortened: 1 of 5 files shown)


test/ctia/entity/web_test.clj
```diff
@@ -403,15 +401,14 @@
...
              (testing "Check URL server returns 401"
                (is (= 401 (:status (get-judgement jwt-1))))
-               (is (= {:error "invalid_jwt"
+               (is (= {:error :invalid_jwt
                        :error_description
                        (str "(56bb5f8c-cc4e-4ed3-a91a-c6604287fe32) SERVER ERROR DESCRIPTION: Bearer "
                             jwt-1)}
-                      (json/read-str (:body (get-judgement jwt-1))
-                                     :key-fn keyword))
+                      (:parsed-body (get-judgement jwt-1)))
```
frenchy64, line 411: This seems to be the only place the jwt invalidation code is exercised. The json format is no longer hard-coded, and is actually EDN thanks to the defaults of clj-momo's GET.

vemv: Might be good to also exercise json.

Also I'm not sure that leveraging clj-momo's GET default yields a clear test.

A test would ideally exemplify a typical usage. If typical usage is json, then reflecting EDN as the default can result in  a test that can break later (and importantly, leave people unsure of what the right fix would be)

src/ctia/auth/jwt.clj
```diff
@@ -55,15 +55,6 @@
              (throw (ex-info err-msg {:jwt-pubkey-map txt})))))))
 
 
-(defn jwt-error-handler
-  "Return an `unauthorized` HTTP response
-  and log the error along debug infos"
-  [error-msg infos]
-  (let [err {:error :invalid_jwt
-             :error_description error-msg}]
-    (log/info error-msg (pr-str (into infos err)))
```
vemv, line 64: Is the logging still performed after this removal?

src/ctia/http/server.clj
```diff
@@ -208,6 +209,9 @@
             (when-let [lifetime (:lifetime-in-sec jwt)]
               {:jwt-max-lifetime-in-sec lifetime}))))
 
+         ;; encode error responses from ring-jwt-middleware
+         (:enabled jwt) (wrap-restful-response (handler/->format-options))
```
msprunck, line 213: The `wrap-restful-response` middleware is already applied by [compojure-api](https://github.com/metosin/compojure-api/blob/1.1.13/src/compojure/api/middleware.clj#L244). This middleware is not explicitly defined as idempotent. We don't know how it behaves with all content types and we don't want to rely on its implementation.

We don't have this issue in IROH, the [`wrap-jwt-auth-fn`](https://github.com/threatgrid/iroh/blob/b89cfdd47b3ea5ab0c5664ae2a3182ae346be3a9/lib/iroh-web/src/iroh_web/core.clj#L291) is passed to the `compojure.api.api/api` fn. Could we have the same approach with CTIA? It was probably not possible before the integration of TK in CTIA.

src/ctia/auth/jwt.clj
frenchy64: Yes. The default handler is exactly this, except it doesn't hardcode json response string (seems like a workaround of this very issue).

src/ctia/http/server.clj
frenchy64: > This middleware is not explicitly defined as idempotent.

If the primary concern is idempotency, here's an alternative implementation that immediately wraps the `:error-handler` in `wrap-restful-response`, and so it does not get called repeatedly: https://github.com/threatgrid/ctia/compare/06bdb88373a4dbe1b362caac3a290fa7fb27acac...7c7b1db66db7d0f5b067964a3a69e583a4ffcf7b?w=1#diff-fd6005488e3bd6e64d13d9d694ad18bfea2ea0ed5be6f2a0d2d81794275c3e85L187-R233

FWIW `wrap-restful-response` is [documented](https://github.com/ngrunwald/ring-middleware-format) to be idempotent on string/binary `:body` (`If a route returns a String, File, InputStream or nil, nothing will be done.`). With that assumption, I settled on the current implementation.

> Could we have the same approach with CTIA?

It would be cleaner and I've tried it unsuccessfully several times. I think it would be safer for someone more familiar with the security side of things to try, since it may involve changing the order of some middleware.

msprunck: > FWIW wrap-restful-response is documented to be idempotent on string/binary :body (If a route returns a String, File, InputStream or nil, nothing will be done.). With that assumption, I settled on the current implementation.

Then the current implementation is correct. Please add a comment explaining why we can use `wrap-restful-response` here even it it's already included in the compojure API middlewares.

msprunck (approved):
LGTM!

ereteog (approved):
LGTM

test/ctia/entity/web_test.clj
frenchy64: Thanks, it took me a week to think of the best way to achieve that but this explicits tests both json and edn https://github.com/threatgrid/ctia/pull/1083/commits/5aef81dca784523aab8830f40e62c47a1d414680

frenchy64: @vemv PR ready for another look.

vemv (approved):
Great work!

frenchy64:
Appreciate the reviews!

PR #1085, opened 2021-03-08
fix flaky rate limit test

ereteog:
> Close #https://github.com/threatgrid/iroh/issues/4899

The value returned by `turnstile` and `turnstile-middleware` is not always the exact expected duration while properly doing its job for `Retry-After` value.
We could probably make it more deterministic on turnstile [`here`](https://github.com/threatgrid/turnstile/blob/1390ffa520ca7b31f9101ac2970ce84772b626f2/src/turnstile/core.clj#L25), [`here`](https://github.com/threatgrid/ring-turnstile-middleware/blob/6bef68919a0038d55721c44a84c03e7827a26f42/src/ring/middleware/turnstile.clj#L85), or [`here`](https://github.com/threatgrid/ring-turnstile-middleware/blob/6bef68919a0038d55721c44a84c03e7827a26f42/src/ring/middleware/turnstile.clj#L144), I don't think that it worths the effort (the rate limiting doing its job).

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

### DIFF
--- a/test/ctia/http/middleware/ratelimit_test.clj
+++ b/test/ctia/http/middleware/ratelimit_test.clj
@@ -122,7 +122,8 @@
                (dotimes [_ 4] (call app))
                (let [response (call app)]
                  (is (= 429 (:status response)))
-                 (is (= "3600" (get-in response [:headers "Retry-After"])))
+                 (is (#{"3599" "3600"} ;; avoid flakyness on slow test server.
+                      (get-in response [:headers "Retry-After"])))
                  (is (= "{\"error\": \"Too Many Requests\"}"
                         (:body response)))))))))
       (testing "Custom group limits"
--- a/test/ctia/http/middleware/ratelimit_test.clj
+++ b/test/ctia/http/middleware/ratelimit_test.clj
@@ -6,7 +6,6 @@
             [schema.test :refer [validate-schemas]]
             [ctia.auth :as auth :refer [IIdentity]]
             [ctia.test-helpers.core :as helpers]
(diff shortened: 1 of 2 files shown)


test/ctia/http/middleware/ratelimit_test.clj
```diff
@@ -122,7 +122,8 @@
                (dotimes [_ 4] (call app))
                (let [response (call app)]
                  (is (= 429 (:status response)))
-                 (is (= "3600" (get-in response [:headers "Retry-After"])))
+                 (is (#{"3599" "3600"} ;; avoid flakyness on slow test server.
```
vemv, line 125: With a range from 3593 to 3600 you'd guard against a variety of "acceptable slowness" (e.g. slow Travis, GC pauses, test fixtures, etc) without compromising much the accuracy of the test.

vemv (approved):

test/ctia/http/middleware/ratelimit_test.clj
ereteog: Yes I agree. I just observed  3595 on my local env.
Due to some kind of OCD I was about to push 3590, any special explanation for 3593 :-)?

frenchy64 (approved):
Appreciate the fix!

test/ctia/http/middleware/ratelimit_test.clj
ereteog: Ok nice, I still in the range with 10 and my OCD that pushes me to use a "round" value :-)

PR #1087, opened 2021-03-08
warm .m2 cache in prebuild

frenchy64:
> **Epic** #
> Close #
> Related #

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
```

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -88,42 +88,32 @@ jobs:
       - name: Run shellcheck
         run: ./scripts/shellcheck-build.sh
 
-      # note: this slows the setup build considerably. the cache-hit
-      #       logic does not seem to work, so steps are run even when
-      #       not needed.
-      ### - name: Maven Cache
-      ###   id: maven-cache
-      ###   uses: actions/cache@v2
-      ###   with:
-      ###     path: ~/.m2/repository
-      ###     key: ctia-m2-cache-${{ hashFiles('project.clj') }}
-      ###     restore-keys: |
-      ###       ctia-m2-cache-
-      ### # warm maven dependency cache if needed
-      ### - name: Setup Java
-      ###   if: steps.maven-cache.cache-hit != 'true'
-      ###   uses: actions/setup-java@v1
-      ###   with:
-      ###     java-version: ${{ env.JAVA_VERSION }}
-      ### - run: java -version
-      ### # install lein using cache to avoid extra download and startup time
-      ### - name: Cache Leiningen
-      ###   if: steps.maven-cache.cache-hit != 'true'
-      ###   uses: actions/cache@v2
-      ###   id: lein-cache
-      ###   with:
-      ###     path: ${{ env.BIN_PATH }}/lein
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yml
```diff
@@ -88,42 +101,47 @@ jobs:
...
+          path: |
+            ~/bin/lein
+            ~/.lein
+          key: ctia-lein-cache-${{ env.ACTIONS_CACHE_VERSION }}-${{ env.LEIN_VERSION }}
+      - name: Install Leiningen
+        if: steps.maven-cache.outputs.cache-hit != 'true' && steps.lein-cache.outputs.cache-hit != 'true'
+        run: |
+          cd "${BIN_PATH}"
+          # poor man's travis_retry
+          for i in {1..${ACTIONS_RETRY_TIMES}}; do curl -sLO "https://raw.githubusercontent.com/technomancy/leiningen/${LEIN_VERSION}/bin/lein" && break; done
```
vemv, line 137: Merely curious, why are these Lein commands flaky?

.github/workflows/pr.yml
```diff
@@ -192,22 +210,21 @@ jobs:
...
-            chmod a+x lein
-            lein version
-          fi
-
+          cd "${BIN_PATH}"
+          # poor man's travis_retry
+          for i in {1..${ACTIONS_RETRY_TIMES}}; do curl -sLO "https://raw.githubusercontent.com/technomancy/leiningen/${LEIN_VERSION}/bin/lein" && break; done
+          chmod a+x lein
+          # poor man's travis_retry
+          for i in {1..${ACTIONS_RETRY_TIMES}}; do lein version && break; done
```
vemv, line 227: Probably you've heard of https://github.com/DeLaGuardo/setup-clojure ?

Might save some LOCs. Using it sha-versioned would constrain risks.

vemv (commented):
Looks reasonable!

Is it possible to test out these changes in some sample branches, exercising the described edge cases?

.github/workflows/pr.yml
frenchy64: I've never observed lein downloads as flaky, these are looped mainly because it's in the category of "downloading stuff".

.github/workflows/pr.yml
frenchy64: Haven't had a great experience with that action in terms of caching. I can't see an indication that's improved, but IIRC caching just "didn't work".

frenchy64:
I was thinking the same thing. It turned out to be somewhat tricky to remove the caches when basing a PR on this branch, so I directly "disabled" the caches in the commit `workaround cache eviction`, which exercised most of those edge cases except the strange ones like a fanned-out job failing to pick up a just-warmed cache.

vemv (approved):
LGTM!

ereteog (commented):
@frenchy64 Before merging it, I would like to talk to you about this PR and the course of action to tackle the recent build problems that started about 3 weeks ago. I think that we need a global plan because it's like the test / build just reached a breaking point.
IMO we need first to identify first the real cause of the slow tests.

PR #1089, opened 2021-03-09
Bundle Import Entities with Asset Refs

agzam:
> **Epic** #
> Close https://github.com/threatgrid/iroh/issues/4751, https://github.com/threatgrid/iroh/issues/4917
  > Related https://github.com/threatgrid/ctim/issues/333, https://github.com/threatgrid/ctim/issues/334

For Assets, (unlike other entities) it was decided not to use Relationship objects to define association with other entities. Specifically, AssetMap and AssetProperties are linked to Assets via their :asset-ref fields. 
This PR ensures `AssetMapping` and `AssetProperties` created via Bundle Import, end up having proper `:asset-ref` values, non-transient IDs of newly created Assets.

<a name="qa">[§](#qa)</a> QA
============================
1. Post a `Bundle` with `AssetMapping` with a long, non-transient `asset_ref`
  Expected: New `AssetMapping` object should be created
  
  ```json
{
  "description" : "test bundle",
  "schema_version" : "1.0",
  "revision" : 0,
  "type" : "bundle",
  "source" : "test",
  "asset_mappings" : [ {
    "asset_type" : "device",
    "valid_time" : {
      "start_time" : "2020-01-11T00:40:48Z",
      "end_time" : "2525-01-01T00:00:00Z"
    },
    "stability" : "Managed",
    "schema_version" : "1.0",
    "revision" : 1,
    "observable" : {
      "type" : "ip",
      "value" : "100.213.110.122"
    },
    "asset_ref" : "https://private.intel.int.iroh.site:443/ctia/asset/asset-275399fe-fbca-4165-bd48-f326bffaa0bc",
    "type" : "asset-mapping",
    "source" : "cisco:unified_connect",
    "external_ids" : [ "d2dcbd00e9bb49719d3fa0a59b1ddfdf" ],
    "external_references" : [ {
      "source_name" : "source",
      "external_id" : "T1061",
      "url" : "https://ex.tld/wiki/T1061",
      "hashes" : [ "#section1" ],
      "description" : "Description text"
    }],
    "source_uri" : "http://example.com/asset-mapping-1",
    "language" : "language",
    "tlp" : "green",
    "specificity" : "Unique",
    "confidence" : "High"
  }]
}
  ```

2. Post a `Bundle` with `AssetMapping` and `Asset`. `AssetMapping` should have a transient `asset_ref` and `Asset` should have the same ID field
  Expected: New `AssetMapping` and new `Asset` objects should be created
  
 ```json
{
    "description" : "test bundle",
    "schema_version" : "1.0",
    "revision" : 0,
    "type" : "bundle",
    "source" : "test",
    "assets": [{
            "id": "transient:asset-1",
            "description" : "asus router",
            "asset_type" : "device",
            "valid_time" : {
                "start_time" : "2020-01-11T00:40:48Z",
                "end_time" : "2525-01-01T00:00:00Z"
            },
            "schema_version" : "1.0",
            "revision" : 1,
            "type" : "asset",
            "source" : "source",
            "external_ids" : [ "61884b14e2734930a5ffdcce69207724" ],
            "short_description" : "awesome router",
            "title" : "ASUS-ROUTER",
            "external_references" : [ {
                "source_name" : "source",
                "external_id" : "T1061",
                "url" : "https://ex.tld/wiki/T1061",
                "hashes" : [ "#section1" ],
                "description" : "doesn't matter"
            }],
            "source_uri" : "http://example.com/asset/asus-router-1",
            "language" : "EN",
            "tlp" : "green",
            "timestamp" : "2020-02-11T00:40:48Z"
        }],
    "asset_mappings" : [{
        "asset_ref" : "transient:asset-1",
        "asset_type" : "device",
        "valid_time" : {
            "start_time" : "2020-01-11T00:40:48Z",
            "end_time" : "2525-01-01T00:00:00Z"
        },
        "stability" : "Managed",
        "schema_version" : "1.0",
        "revision" : 1,
        "observable" : {
            "type" : "ip",
            "value" : "100.213.110.122"
        },
        "type" : "asset-mapping",
        "source" : "cisco:unified_connect",
        "external_ids" : [ "d2dcbd00e9bb49719d3fa0a59b1ddfdf" ],
        "external_references" : [ {
            "source_name" : "source",
            "external_id" : "T1061",
            "url" : "https://ex.tld/wiki/T1061",
            "hashes" : [ "#section1" ],
            "description" : "Description text"
        } ],
        "source_uri" : "http://example.com/asset-mapping-1",
        "language" : "language",
        "tlp" : "green",
        "specificity" : "Unique",
        "confidence" : "High"
    }]
}
 ``` 
  
3. Post a `Bundle` with a single `AssetMapping` with a transient `asset_ref`
  Expected: It should not allow it, since the Bundle doesn't contain `Asset` with the same transient ID
  
  ```json
{
    "description" : "test bundle",
    "schema_version" : "1.0",
    "revision" : 0,
    "type" : "bundle",
    "source" : "test",
    "asset_mappings" : [{
        "asset_ref" : "transient:asset-1",
        "asset_type" : "device",
        "valid_time" : {
            "start_time" : "2020-01-11T00:40:48Z",
            "end_time" : "2525-01-01T00:00:00Z"
        },
        "stability" : "Managed",
        "schema_version" : "1.0",
        "revision" : 1,
        "observable" : {
            "type" : "ip",
            "value" : "100.213.110.122"
        },
        "type" : "asset-mapping",
        "source" : "cisco:unified_connect",
        "external_ids" : [ "d2dcbd00e9bb49719d3fa0a59b1ddfdf" ],
        "external_references" : [ {
            "source_name" : "source",
            "external_id" : "T1061",
            "url" : "https://ex.tld/wiki/T1061",
            "hashes" : [ "#section1" ],
            "description" : "Description text"
        } ],
        "source_uri" : "http://example.com/asset-mapping-1",
        "language" : "language",
        "tlp" : "green",
        "specificity" : "Unique",
        "confidence" : "High"
    }]
}
  ```
4. Repeat the same checks using `AssetProperties` instead of `AssetMapping`
5. Optional: check that Bundle Import works for other entities (as before)

```
intern: Bundle imports now properly work for AssetMapping, AssetProperties
```

910bfe4c assert for the value to be non-empty
a8049085 replace local version of fn
995677ab re-enable disabled tests
87fa0c68 remove unused fn
4c813fb5 test: validate asset_refs
261f2621 move the logic for linking asset_refs into realize functions
f2b22cad bump ctim version
4bc68728 updates test, address PR comments
6a08bef7 dissoc IDs for testing
2a5ae284 simplify the test body
e6a13d12 use examples instead of hand-crafter fixture
7d69ac99 don't dissoc transient IDs
5370dec4 leave long IDs alone and adjust only the transients
15abca2b move transient-id? fn to schemas.core
016387a9 disable another failing test
67e4f8ac temporarily disable failing test
85fa2518 adjust tests
9cb3afc5 test: asset-ref entities
479c01f1 adds create-asset-ref-entities fn
e2c5e56b sort ns header
140b3c61 adds fake test


test/ctia/bundle/core_asset_ref_entities_test.clj
```diff
@@ -0,0 +1,197 @@
...
+                               {:refresh true}
+                               services)
+           get-records        (fn [store]
+                                (-> (get-store store)
+                                    (store/list-records
+                                     {:query "*"}
+                                     {:login  "foouser"
+                                      :groups ["foogroup"]} {})
+                                    :data))
+           expected-asset-ref (get tempids (-> bundle-ents :assets first :id))]
```
vemv, line 190: probably worth `assert`ing `expected-asset-ref`, since `(= nil ...)` is probably not what you would mean

test/ctia/bundle/core_asset_ref_entities_test.clj
```diff
@@ -0,0 +1,197 @@
...
+       (testing "every AssetMapping and AssetProperties have proper
+                 non-transient :asset-ref"
+        (are [store] (every?
+                      (partial = expected-asset-ref)
+                      (->> store get-records (map :asset_ref)))
```
vemv, line 195: You can `assert seq` the result of `get-records`, otherwise `every?` can return `true` for an empty coll

test/ctia/bundle/core_asset_ref_entities_test.clj
```diff
@@ -0,0 +1,197 @@
...
+                                    (store/list-records
+                                     {:query "*"}
+                                     {:login  "foouser"
+                                      :groups ["foogroup"]} {})
+                                    :data))
+           expected-asset-ref (get tempids (-> bundle-ents :assets first :id))]
+       (testing "every AssetMapping and AssetProperties have proper
+                 non-transient :asset-ref"
+        (are [store] (every?
+                      (partial = expected-asset-ref)
```
vemv, line 194: `every?` and `partial =` are two things that tend to not report particularly well under clojure.test.

Maybe I'd use simple `=` and `doseq` here

vemv (commented):
Some superficial remarks

src/ctia/bulk/core.clj
```diff
@@ -157,6 +157,28 @@
...
+   services :- APIHandlerServices]
+  (let [bulk' (select-keys bulk [:asset_mappings :asset_properties])
+        ;; replace :asset_ref fields with non-transient IDs
+        ents (zipmap
+              (keys bulk')
+              (->> bulk'
+                   vals
+                   (map
+                    (partial
+                     map #(assoc % :asset_ref (get temp-ids (:asset_ref %)))))))]
```
ereteog, line 172: perhaps we should add some check here in case the entities was not created for any reason.

agzam: Absolutely. Will be addressed within the scope of https://github.com/threatgrid/iroh/issues/4917

ereteog (commented):
We must preserve the ability to push `asset-mappings` and `asset-properties` with long ids and in bundles that do not contain the referenced assets.
`asset-mappings` and `asset-properties` expire and new ones are added during the life cycle of an asset.
For instance a device while be assigned new ips, and we while have new sightings observed and thus mapped to it.

test/ctia/bundle/core_asset_ref_entities_test.clj
```diff
@@ -0,0 +1,197 @@
...
+              (bulk/create-bulk
+               bulk
+               tempids
+               login
+               {}
+               services))))))))))
+
+(use-fixtures :once
+  (join-fixtures
+   [whoami-helpers/fixture-server]))
```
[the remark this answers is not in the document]
ereteog: If you need to set whoami response and capabilities, you should do it in the test directly to ensure a better test isolation (we made an intense use of `use-fixtures` but developed some kind of allergy when it comes to maintain and reason about the test states over time).
https://github.com/threatgrid/ctia/pull/1094/files#diff-1688b4250f73b5659728c664f6ea0412f24b88435eed3f1688f87a8336704cfc

src/ctia/bulk/core.clj
agzam: > long-id ==> keep it (we do not check that the corresponding asset exists, it's the responsability of the client)

Can you elaborate on that? So you think we'd have situations where AssetMappings/AssetProperties would be created ahead-of-time (before Assets)? And then the client would have to create Assets later?
If we allow that to happen, wouldn't that increase entropy and allow inconsistent states? I guess my assumptions in threatgrid/iroh#4917 are slightly not correct?

test/ctia/entity/asset_mapping_test.clj
```diff
@@ -73,12 +75,11 @@
...
-                app
-                new-asset-mapping-maximal
-                :asset_mappings
-                30
-                {"Authorization" "45c1f5e3f05d0"})]
+     (let [ids (-> app
+                (helpers/POST-bulk
+                 (-> (fixtures/bundle 10 true)
+                     (select-keys [:assets :asset_mappings])))
+                :asset_mappings)]
```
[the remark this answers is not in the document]
agzam: Ah, missed that one. BTW, the other day I discussed with Victor the possibility of moving IROH's `user.format/clean-staged-namespaces!` to CTIA. Turns out it's not as simple. According to him: 

> (This) isn't so trivial, because formatting-stack and user.* are densely/delicately configured

Maybe if I annoy him with inconsistent formatting and indentation he'd finally extract and bring it over here :)

test/ctia/entity/asset_mapping_test.clj
```diff
@@ -93,7 +94,8 @@
         {"Authorization" "45c1f5e3f05d0"}
         sut/asset-mapping-fields)))))
 
-(deftest asset-mapping-metric-routes-test
+;; TODO: Fix it
```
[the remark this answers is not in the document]
agzam: I don't think there's anything has to be done here, addressing this https://github.com/threatgrid/ctim/issues/334 should fix it.

src/ctia/bulk/core.clj
ereteog: No I mean that asset-mappings and asset-properties could be pushed later.
Like I mentioned:
> For instance a device while be assigned new ips, and we while have new sightings observed and thus mapped to it.

So in this example here could be the life cycle:
- at t1, A bundle is pushed with an `asset` to represent a machine and an `asset-mapping` to map it to an ip.
- at t2 the machine ip changes. 
- at t3  a bundle is pushed with an `asset-mapping` to map it to an ip.

It could also happen to any other mapped observable and any other mapped properties.

test/ctia/entity/asset_mapping_test.clj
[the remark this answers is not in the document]
agzam: Umm, it looks like the problem here is due to how the `:refresh` parameter gets set in the tests. Unless it's set to `true`, some entities won't be seen immediately (in this case `asset_mappings/asset_properties`). And posting bundle using the http route sets it to `wait_for`.

test/ctia/entity/asset_mapping_test.clj
```diff
@@ -73,12 +75,11 @@
...
                                          "user")
 
-     (let [ids (helpers/POST-entity-bulk
-                app
-                new-asset-mapping-maximal
-                :asset_mappings
-                30
-                {"Authorization" "45c1f5e3f05d0"})]
+     (let [ids (-> app
+                (helpers/POST-bulk
```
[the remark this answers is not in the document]
agzam: I still would like to address this in threatgrid/iroh#4917. I will update the ticket - the description no longer accurate, I will copy your comments about it over there. Would that be okay?

test/ctia/bundle/core_asset_ref_entities_test.clj
agzam: addressed in cd60d032cca74dcc311d62a3c3a908ef34059243

test/ctia/bundle/core_asset_ref_entities_test.clj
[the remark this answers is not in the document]
agzam: addressed in https://github.com/threatgrid/ctia/pull/1089/commits/cd60d032cca74dcc311d62a3c3a908ef34059243

src/ctia/bulk/core.clj
```diff
@@ -157,6 +157,31 @@
...
+                   vals
+                   (map (partial map set-ref))))]
+    ;; create asset-mapping and asset-properties entities
+    (gen-bulk-from-fn
+     create-entities
+     ents
+     temp-ids
+     login
+     {:refresh refresh}
+     services)))
```
ereteog, line 183: Indeed you should simply do this:
``` clojure
(gen-bulk-from-fn
     create-entities
     (select-keys bulk [:asset_mappings :asset_properties])
     temp-ids
     login
     {:refresh refresh}
     services)))
```
and move the transient id resolution  to a realize-fn function that would be defined as a field of asset-mapping and asset-properties

agzam: done in https://github.com/threatgrid/ctia/pull/1089/commits/261f262171cbfef23ad467ac63eeb5cc1b69b6e2

src/ctia/bulk/core.clj
```diff
@@ -167,26 +167,34 @@
...
-                            (merge-tempids new-relationships))
-         all-entities (into new-entities new-relationships)
+         entities-tempids (into tempids (merge-tempids new-entities))
+         new-linked-ents (gen-bulk-from-fn
+                          create-entities
+                          (select-keys
+                           bulk
+                           [:relationships
+                            :asset_mappings
+                            :asset_properties])
```
ereteog, line 191: perfect!

src/ctia/task/migration/migrations/investigation_actions.clj
```diff
@@ -1,6 +1,8 @@
 (ns ctia.task.migration.migrations.investigation-actions
   (:require [cheshire.core :as json]))
 
+
+;; TODO: replace with ctia.schemas.core/transient-id?
```
ereteog, line 5: could you handle it in this PR?

test/ctia/bundle/core_asset_ref_entities_test.clj
```diff
@@ -0,0 +1,149 @@
...
+                                       :data)))
+             asset-refs        (->> results
+                                  (filter #(-> % :result (= "created")))
+                                  (filter #(-> % :type (= :asset)))
+                                  (map :id)
+                                  set)]
+         (doseq [store [:asset-mapping :asset-properties]]
+           (is (every?
+                (partial contains? asset-refs)
+                (->> store get-records (map :asset_ref))))))))))
```
ereteog, line 106: Learning from @vemv :-)
```clojure
(every? check? nil)
;; => true
```

I would assert that it's not empty

```suggestion
(let [refs (->> store get-records (map :asset_ref))]
     (assert (seq refs))
     (is (every? (partial contains? asset-refs)
                        refs)))))))))
```

ereteog (approved):
Great work!

NeelSaini:
While testing the third step mentioned in the QA section. Following behaviour is noticed

- When hitting the bundle import endpoint with the script mentioned in third step after deleting the already existing asset-mapping the response is as mentioned below: @agzam Do you mind taking a look please if that's expected

`{
"type": "unknown-exception",
"class": "clojure.lang.ExceptionInfo"
}`

hisans:
QA Verified in 1.72 Release

PR #1092, opened 2021-03-09
Replace most of Travis cron with scalable GitHub Actions job and record individual deftest timings

frenchy64:
Related https://github.com/threatgrid/iroh/issues/4920
Related https://github.com/threatgrid/iroh/issues/4926

Adds a cron job in GitHub Actions that is like the Travis one, except it splits each job into 2. This might help with memory issues, and, if not, there's a handy dial we can turn increase cron parallelism (`cron-ctia-nsplits`).

I think this is a reasonable way to tackle https://github.com/threatgrid/iroh/issues/4920 (moving to a more flexible CI provider).

See `test cron` commit below for it in action--it completes in 45m: https://github.com/threatgrid/ctia/runs/2070971550

Actions has higher per-job and per-workflow limits than Travis, but has slightly less memory.

To see deftest timing information, see the `Print test timings` section of the `all-pr-checks` job, and scroll down to where the deftest's are listed. It'll look like [this](https://github.com/threatgrid/iroh/issues/4926#issuecomment-794601057).

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: Replace Travis cron with GitHub Actions
```


dev/ctia/dev/test_reporter.clj
```diff
@@ -49,10 +49,14 @@
 
   (begin-test-var [this m])
 
-  (end-test-var [this {:keys [elapsed] var-ref :var}]
+  (end-test-var [this {elapsed-s :elapsed var-ref :var}]
+    (-> out-dir File. .mkdirs)
+    (let [elapsed-ns (* elapsed-s 1e+9)]
+      (spit (str out-dir "/var-timing-" (gensym (munge (symbol var-ref))) ".edn")
```
vemv, line 55: You can use `File.` instead of string concatenation

.github/workflows/pr.yml
```diff
@@ -8,6 +8,10 @@ on:
   push:
     branches:
       - master
+  # automatically uses master https://stackoverflow.com/a/58800550
+  schedule:
+    # daily, 3am UTC
```
vemv, line 13: Worth documenting - what is being gained by running this setup as a cron?

Right now I can't spot the difference between a `master` and `cron` build.

vemv (approved):
LGTM!

Worst-case scenario appears to be simply a failed cron build, so this could always be improved iteratively.

ereteog (approved):
LGTM

.github/workflows/pr.yml
frenchy64: Yes, the difference is hidden away in the `cron-matrix` function (jdk+clj matrix build). I'll document that here.

frenchy64:
@ereteog yes, there's a risk that the custom test runner is wrong. That's one reason why I left one Travis cron job. It's probably about time to pull the custom testing logic into it's own unit-tested library to minimize that risk.

PR #1094, opened 2021-03-10
Only test wait_for with incidents

frenchy64:
Related https://github.com/threatgrid/iroh/issues/4926

We just need to test wait_for on one entity. It takes about 7 seconds on my machine to test wait_for for an entity, and there are about 26 entities in total, so this might save a few minutes on our test suite.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: Only test wait_for with incidents
```


test/ctia/stores/es/wait_for_test.clj
```diff
@@ -0,0 +1,31 @@
...
+    #{:es-store}
+    (fn [app]
+      (helpers/set-capabilities! app "foouser" ["foogroup"] "user" (all-capabilities))
+      (whoami-helpers/set-whoami-response app "45c1f5e3f05d0" "foouser" "foogroup" "user")
+      (let [{{:keys [get-in-config]} :ConfigService} (app/service-graph app)
+            {:keys [entity] :as parameters} (into incident-entity
+                                                  {:app app
+                                                   :example new-incident-maximal
+                                                   :headers {:Authorization "45c1f5e3f05d0"}})]
+        (is (= "es" (get-in-config [:ctia :store entity])))
```
ereteog, line 30: Perhaps `assert` since this is more a prerequisite than a real test.

ereteog (approved):
LGTM!

Thanks!

test/ctia/stores/es/wait_for_test.clj
```diff
@@ -0,0 +1,31 @@
+(ns ctia.stores.es.wait-for-test
```
ereteog, line 1: Just one detail to discuss before you merge it.
Even if the feature is closely tied to ES, this test a feature at the route level. 
Perhaps it belongs more to `ctia.http.routes.common-test`.
wdyt?

frenchy64: Yes, makes sense.

frenchy64:
I had to make a few changes to existing tests because schema checking wasn't enabled in `ctia.http.routes.common-test`. I also rearranged and added some schemas in `src`.

PR #1097, opened 2021-03-16
Fix Actions timing cache, fall back to cron timing if needed

frenchy64:
Also Close https://github.com/threatgrid/iroh/issues/4950

The original goal here was to ensure pull requests don't use cron timing information, because https://github.com/threatgrid/ctia/pull/1095 made the cron build different from normal builds.

Unfortunately, the approach in https://github.com/threatgrid/ctia/pull/1095 was flawed by using `${{ github.event_name }}` in the cache key because that makes caches for `push` and `pr` distinct, which means pull requests will never see caches generated by master (and those are the only ones reachable for the first build of a PR).

Instead, this groups pr/push timing information in the same cache via `${{ env.CTIA_TEST_SUITE }}` (set to `cron` for cron, and `ci` for pr/push builds) so pr's can see push caches again, and allows pr/pushes to use cron timing as a last resort.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: Fix Actions timing cache, fall back to cron timing if needed
```


scripts/actions/setup-env.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+set -e
```
vemv, line 3: more flags?

vemv (approved):
LGTM, thanks for the explanation!

scripts/actions/setup-env.sh
frenchy64: Thanks, yes. First I'll see if I can rewrite it in bb :)

frenchy64:
I think I found the cause of some of the symptoms of https://github.com/threatgrid/iroh/issues/4950.

It turns out that `-dev` doesn't work yet to run the tests. We were warming the Maven cache with `+test`, but some of the ci commands use `-dev,+test`. This resulted in different versions of libraries being used.

I'm surprised `:pedantic? :abort` didn't pick up the differences in versions, but I guess that would only happen if we use `:managed-dependencies` to fix our transitive dependencies.

scripts/actions/setup-env.sh
frenchy64: Well, there was a some bootstrapping issues between this script setting up the local binary path that `bb` is installed to, but I managed to rewrite it in `bb`.

PR #1100, opened 2021-03-18
Cache future Clojure jars in Actions builds

frenchy64:
Close https://github.com/threatgrid/iroh/issues/4997

Previously, `LEIN_OFFLINE=true` was added to CI to catch uncached dependencies in the Actions builds. Well, it caught a FIXME that future Clojure versions were never cached in the cron build: https://github.com/threatgrid/ctia/runs/2136456845

The basic approach to fix this is download all permutations of build dependencies (pr + cron matrix) into a single cache, indexed by the hash of `project.clj`. The main reason this is the easiest approach is because the cron job uses a single setup job to download dependencies, but then it fans out into multiple configurations. By using a single cache for all permutations, we can keep this setup and minimize the number of simultaneous jobs downloading deps. FWIW this approach means one cache per project.clj hash instead of multiple for cron/ci (total caches are limited to 2GB).

I briefly considered using Leiningen composite profiles for this, but I found this approach more extensible and robust. We now have a single place to add `-dev` that is compatible with nested `with-profile` calls (seemingly the cause of https://github.com/threatgrid/iroh/issues/4950, `with-profile` seems to automatically reenable the base profiles).

```
$ lein with-profile -dev,+test deps :why circleci/circleci.test
$ lein with-profile -dev,+test with-profile +test deps :why circleci/circleci.test
 [circleci/circleci.test 0.4.3]
```

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: Cache future Clojure jars in Actions builds
```


project.clj
```diff
@@ -179,20 +193,7 @@
...
-                                 (System/getProperty "TRAVIS") (into ["-Xms3g"
-                                                                      "-Xmx3g"])
-                                 ; we have 7GB RAM on Actions
-                                 ; - https://docs.github.com/en/free-pro-team@latest/actions/reference/specifications-for-github-hosted-runners#supported-runners-and-hardware-resources
-                                 ; docker reserves an unknown amount of RAM.
-                                 ; reserving 3GB for jvm -- this might need tweaking as we learn
-                                 ; more about docker on actions.
-                                 (System/getProperty "GITHUB_ACTIONS") (into ["-Xms3g"
-                                                                              "-Xmx3g"]))
+             :test {:jvm-opts ["-Dlog.console.threshold=WARN"]
```
frenchy64, line 196: This is dead code (should have used `System/getenv`).

scripts/actions/print-matrix.clj
```diff
@@ -1,41 +1,47 @@
...
+(defn splits-for [base nsplits]
+  {:post [(= (range nsplits)
+             (map :this_split %))
+          (= #{nsplits}
+             (into #{} (map :total_splits %)))]}
+  (for [this-split (range nsplits)]
+    (assoc base
+           :this_split this-split
+           :total_splits nsplits)))
+
```
frenchy64, line 21: This code has been bugging me because mistakes here means tests being skipped. I added sufficient post conditions so I can sleep at night.

scripts/actions/print-matrix.clj
```diff
@@ -1,41 +1,56 @@
...
-     :clojure_version default-clojure-version
-     :java_version default-java-version}))
+  (splits-for
+    {:ci_profiles "default"
+     :java_version default-java-version}
+    non-cron-ctia-nsplits))
 
 (defn cron-matrix
   "Actions matrix for cron builds"
   []
```
vemv, line 41: any postcondition worth adding?

vemv (approved):
LGTM!

scripts/actions/print-matrix.clj
```diff
@@ -1,41 +1,56 @@
...
+                  {:java_version java-15-version}]))))
 
 (defn edn-matrix []
   {:post [(seq %)]}
   (if (= "schedule" (System/getenv "GITHUB_EVENT_NAME"))
     (cron-matrix)
     (non-cron-matrix)))
 
-(println "::set-output name=matrix::" (json/generate-string (edn-matrix) {:pretty false}))
+(println (str "::set-output name=matrix::" (json/generate-string (edn-matrix) {:pretty false})))
```
vemv, line 56: looks like there used to be a single whitespace between `name=matrix::` and the json string; I assume the removal is intentional?

frenchy64: Yes, the syntax is `::set-output name={name}::{value}`.

scripts/actions/print-matrix.clj
frenchy64: Thanks for asking, a bunch came to me https://github.com/threatgrid/ctia/pull/1100/commits/f47e36d3d9384b9760180395de7d9b03e4e739f4

frenchy64:
Thanks for the review!

PR #1102, opened 2021-03-23
Check Bundle with asset_refs

agzam:
Bundle Import. Don't create `AssetMapping`,`AssetProperties` with transient asset_refs, if the `Bundle` doesn't have corresponding `Asset`(s)

> **Epic** #
> Close #https://github.com/threatgrid/iroh/issues/4917
> Related #

<a name="qa">[§](#qa)</a> QA
============================

1) Post a `Bundle` with `AssetMapping` with a long, non-transient `asset_ref`
    Expected: New `AssetMapping` object should be created
2) Post a `Bundle` with `AssetMapping` and `Asset`. `AssetMapping` should have a transient `asset_ref` and `Asset` should have the same `ID` field
    Expected: New `AssetMapping` and new `Asset` objects should be created
3) Post a `Bundle` with `AssetMapping` with a transient `asset_ref`
    Expected: It should not allow that, since the Bundle doesn't contain `Asset` with the same `ID`
4) Repeat the checks using `AssetProperties` instead of `AssetMapping`


agzam:
The feature was implemented in https://github.com/threatgrid/ctia/pull/1089.

PR #1107, opened 2021-03-30
Issue-1106: promotion_method in Incident entity

mcosta85:
Adds support for the new `promotion_method` field being added in
CTIM.

> Related # https://github.com/threatgrid/ctim/issues/338
> Related # https://github.com/threatgrid/response/issues/625

Describe your PR for reviewers.

This PR adds the promotion_method field to the incident entity. It is optional and
can have a value of either `manual` or `automatic`.

<a name="qa">[§](#qa)</a> QA
============================

Positive case - no errors expected
1. POST an incident with a promotion_method of `manual`
2. Note the id
3. GET the incident by id

Negative case - errors expected
1. POST an incident with a promotion_method of `securex`

```
Adds promotion_method to incident model
```

### DIFF
--- a/src/ctia/entity/incident.clj
+++ b/src/ctia/entity/incident.clj
@@ -149,7 +149,8 @@
       :categories em/token
       :discovery_method em/token
       :intended_effect em/token
-      :assignees em/token})}})
+      :assignees em/token
+      :promotion_method em/token})}})
 
 (def-es-store IncidentStore :incident StoredIncident PartialStoredIncident)
 
@@ -167,7 +168,8 @@
            :incident_time.rejected
            :discovery_method
            :intended_effect
-           :assignees]))
+           :assignees
+           :promotion_method]))
 
 (def incident-sort-fields
   (apply s/enum incident-fields))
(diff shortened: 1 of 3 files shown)


src/ctia/entity/incident.clj
```diff
@@ -149,7 +149,8 @@
       :categories em/token
       :discovery_method em/token
       :intended_effect em/token
-      :assignees em/token})}})
+      :assignees em/token
+      :promotion_method em/token})}})
```
ereteog, line 153: This require an update of the mapping from OPS, unless it will be ignored and not indexed (and thus not searchable).

ereteog (commented):
lgtm.
However it cannot be merged yet.
I need to think about how we will update the mapping of existing instances.
We previously added automatic mapping update at startup and an update mapping task.
But all this is deactivated, waiting for ES7 migration that will introduce new mapping management.

ereteog:
I will probably trigger a partial mapping update manually on each env: 
https://www.elastic.co/guide/en/elasticsearch/reference/current/indices-put-mapping.html
I will investigate it next week.

ereteog:
@mcosta85 @saintx I made the mapping change on the INT private incident index.
I merged the [CTIM PR](https://github.com/threatgrid/ctim), you need to do an additional one for the release and then bump the version in this PR that uses the SNAPSHOT
Once the release is done, check if it works on it and let me know.

ereteog (approved):
LGTM!

PR #1108, opened 2021-03-31
renames DateRange query to correspond to ES DSL

agzam:
The name should reflect the features of Elasticsearch Range query
https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-range-query.html#range-query-ex-request

> **Epic** #
> Close https://github.com/threatgrid/iroh/issues/5075
> Related #

No QA is needed.

157b9167 rename :date-range key
c40463db trying to leave the key as is
cd1b4948 renames DateRange query to correspond to ES DSL


ereteog (approved):
NIce!

PR #1109, opened 2021-04-06
Improve full-text search

agzam:
This PR adds a capacity to use different types of Elasticsearch [Full text queries](https://www.elastic.co/guide/en/elasticsearch/reference/current/full-text-queries.html)

> **Epic** #
> Close https://github.com/threatgrid/iroh/issues/4959
> Related #

<a name="qa">[§](#qa)</a> QA
============================

We now support these types of ES queries: [query_string](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html), [simple_query_string](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-simple-query-string-query.html), and [multi_match](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-multi-match-query.html) queries.

To comprehensively test this newly added feature, *(although not required)* it helps to understand subtle differences between these query modes. Of course, the number of combinations and permutations of possible query variations is virtually unlimited. We encourage you to try to think of different scenarios and edge cases; of course, there's always a chance that we might have missed something.

When testing, please ensure that the existing data in the cluster satisfies the expectations, i.e., if the query doesn't return expected results - check if the data is there; if the query returns more items than expected - perhaps multiple items match the criteria and some cleaning required. Generally, it's probably a good idea to delete all the records 

image

and create new ones before testing each test case.

Alternatively, you may import a Bundle with different types of entities (e.g., Assets, Incidents, Indicators) and test each case on a different Entity type. Since these tests are to ensure the search per entity type, not across-the-board - i.e., surplus of Indicators data won't interfere with the search performed on Incidents.

### Return all the records
- Create a few Incidents (either one by one or by using the Bundle Import path)
- Then use `/ctia/incident/search` path
- Set `query` to `*`, execute
- Expected: The response should contain every created Incident entity

### Test for specific field value
  - Create a few Incidents with random values
  - A couple (or a few) of them should have the same value, for example:
<details>
    <summary> The following Bundle contains 3 incidents, 2 of them have the Title: "Lorem Ipsum Test Incident" </summary>

      {
        "description" : "incidents to test",
        "valid_time" : {
          "end_time" : "2016-07-11T00:40:48Z",
          "start_time" : "2016-05-11T00:40:48Z"
        },
        "schema_version" : "1.1.3",
        "revision" : 1,
        "source" : "source",
        "short_description" : "bundle",
        "title" : "title",
        "source_uri" : "http://example.com",
        "language" : "language",
        "tlp" : "green",
        "incidents" : [{
          "description" : "description of first incident",
          "assignees" : [ ],
          "schema_version" : "1.1.3",
          "revision" : 1,
          "type" : "incident",
          "source" : "source",
          "external_ids" : [ ],
          "short_description" : "first incident",
          "title" : "Lorem Ipsum Test Incident",
          "incident_time" : {
            "closed" : "2016-02-11T00:40:48Z",
            "discovered" : "2016-02-11T00:40:48Z",
            "opened" : "2016-02-11T00:40:48Z",
            "rejected" : "2016-02-11T00:40:48Z",
            "remediated" : "2016-02-11T00:40:48Z",
            "reported" : "2016-02-11T00:40:48Z"
          },
          "external_references" : [ ],
          "discovery_method" : "Log Review",
          "source_uri" : "http://example.com/incident-one",
          "intended_effect" : "Extortion",
          "categories" : [ "Denial of Service", "Improper Usage" ],
          "status" : "Open",
          "language" : "language",
          "tlp" : "green",
          "confidence" : "High"
        }, {
          "description" : "second incident",
          "assignees" : [ ],
          "schema_version" : "1.1.3",
          "revision" : 1,
          "type" : "incident",
          "source" : "source",
          "external_ids" : [ ],
          "short_description" : "incident numero does",
          "title" : "Lorem Ipsum Test Incident",
          "incident_time" : {
            "closed" : "2016-02-11T00:40:48Z",
            "discovered" : "2016-02-11T00:40:48Z",
            "opened" : "2016-02-11T00:40:48Z",
            "rejected" : "2016-02-11T00:40:48Z",
            "remediated" : "2016-02-11T00:40:48Z",
            "reported" : "2016-02-11T00:40:48Z"
          },
          "external_references" : [ ],
          "discovery_method" : "Log Review",
          "source_uri" : "http://example.com/incident-two",
          "intended_effect" : "Extortion",
          "categories" : [ "Denial of Service", "Improper Usage" ],
          "status" : "Open",
          "language" : "language",
          "tlp" : "green",
          "confidence" : "High"
        }, {
          "description" : "third incident",
          "assignees" : [ ],
          "schema_version" : "1.1.3",
          "revision" : 1,
          "type" : "incident",
          "source" : "source",
          "external_ids" : [ ],
          "short_description" : "incident numero tres",
          "title" : "incident three",
          "incident_time" : {
            "closed" : "2016-02-11T00:40:48Z",
            "discovered" : "2016-02-11T00:40:48Z",
            "opened" : "2016-02-11T00:40:48Z",
            "rejected" : "2016-02-11T00:40:48Z",
            "remediated" : "2016-02-11T00:40:48Z",
            "reported" : "2016-02-11T00:40:48Z"
          },
          "external_references" : [ ],
          "discovery_method" : "Log Review",
          "source_uri" : "http://example.com/incident-three",
          "intended_effect" : "Extortion",
          "categories" : [ "Denial of Service", "Improper Usage" ],
          "status" : "Open",
          "language" : "language",
          "tlp" : "green",
          "confidence" : "High"
        } ],
        "timestamp" : "2016-02-11T00:40:48Z"
      }

</details>
      
- Import the Bundle
- Issue search request `/ctia/incident/search`, where `query` set to: `title:Lorem Ipsum Test Incident`
- Expected: Results should contain two Incidents

- Issue search request, where `query:` set to: `Lorem Ipsum Test Incident` and `search_fields` set to: `title`
- Expected: Results should contain two Incidents (the same as the previous query)

### Query for non-existing values
- Using Bundle data from the previous test case
- Execute a search request `/ctia/incident/search`, where `query` is set to: `title:7bdfb3fe-d806-4e95-988c-7e416836c7e3`
- Obviously, the Incident with such title doesn't exist
- Expected: Empty results

- Similarly, you can test with `search_fields` set to: `title`, and `query` set to: `7bdfb3fe-d806-4e95-988c-7e416836c7e3`
- Expected: Empty results

### Querying for non-existing fields
- Searching for a field that doesn't exist should either fail or yield no results
- Setting `query` parameter to: `theres_no_field_like_that:*` - (Return every entity with any value in the field "theres_no_field_like_that")
- Expected: Should return no results

- Another way of testing this is setting `search_fields` param, but since Swagger UI doesn't let you set any arbitrary values for that parameter, the only way of testing this is outside of Swagger, e.g., using curl `curl -X GET "http://localhost:3000/ctia/incident/search?query=*&search_fields=theres_no_field_like_that" -H "accept: application/json"`
- In this case, it's expected to fail with a validation error

### Matching multiple fields
   
If you want to search for Incidents that have specific values in specific fields, you can use a query like this:
- `query`: `"\"title of test incident\" AND \"Log Review\""`, `search_fields`: `["title" "discovery_method"]`

### Looking for the same value in different fields in multiple records

     {
      "query_mode" : "multi_match",
      "query" : "TESTING_SHMESTING",
      "search_fields" : [ "assignees", "title" ]
     }

In this case, it should find all the Incidents where `Title` is `TESTING_SHMESTING`, and there's at least one `Assignee` with the same value. You can modify Bundle from the first test case accordingly and import it.

### Summary

As you can see, there are now many different ways of searching for entities.

d204f02f positive constraint instead of not=
eabf2361 a few comments
3ba1583a adds an example meta to the function
7ea7ed76 renames a field
a3349882 refactors, renames and moves function
ec5ccb0f fixes the query in the test so it works in ES 5 and 7
55566ad5 disable a test that doesn't work in ES v5
3a1ba616 updates tests for passing
8ed4c3c4 makes sure generated bundle has sufficient number of entities
1d44822d De-emphasize "Lucene" in swagger context summary
9eda3fba additional test cases
0620f2ec re-factor generators for performance
02d631b2 first couple of extended fullsearch text query tests
ad401179 adds auxiliary function
fd4d2cd4 extends default /search route, so :fields can be passed
66c1bc01 multi_match queries in ES don't support default_operator
4fd45df3 adjust test for updated function signature
f764413e rename :mode to :query_mode and make the field optional
d519cb9a remove :query_mode from filter-map in search-query
3e842ef7 expose query_mode parameter on default /search route
e9319007 converts positional arguments into maps
33162c3f test_helpers/crud. refactor ns heading
a1488c10 adjust crud-tests: make-search-query-test
0c179e12 adjust crud-tests: handle-query-string-search-count-test
9daafc08 adjust crud-tests: handle-delete-search
01f83b0e adjust crud-tests: handle-aggregate-test
629534c3 fix default operator
f3f8db16 adjusting tests
499fbc37 attempt to remove :query-string field
3c3d94f5 converting full-text-query to ES DSL query
49961931 schema change
3e0c890d rename :date-range key
448b6aa2 trying to leave the key as is


test/ctia/stores/es/crud_test.clj
```diff
@@ -555,15 +571,17 @@
...
                       :filter-map {:confidence "High"}}
               :matched []
               :ident ident-2
               :deleted? false})
 
             (check-fn
              {:msg "matched entities must be properly deleted"
-              :query {:query-string "title1"
+              :query {:full-text {:query "title1"
+                                  :mode  :query_string}
```
ereteog, line 584: there is a lot of boiler plate here perhaps you should factorize it with a helper:
``` clojure
(defn query-string
   ([query] (query-string query nil) 
   ([query options]
      (into {:query query
             :mode  :query_string}
             options))
```

and even derive it into a more generic helper in this [ns](https://github.com/threatgrid/ctia/blob/master/src/ctia/http/routes/common.clj):

``` clojure
(defn full-text
   ([mode query] (query-string mode query  nil) 
   ([mode query  options]
       (into {:query query}
               options)))
             
(def query-string (partial full-text :query_string))
(def simple-query-string (partial full-text :simple_query_string))
(def multi-match (partial full-text :multi_match))
```

At some point I think that we should integrate it into [ductile](https://github.com/threatgrid/ductile/blob/master/src/ductile/query.clj) to promote our Elasticsearch querying practices  :-)

ereteog (commented):
it looks good.
For the test that you will add, you could simply check that the ES queries are properly generated.

test/ctia/stores/es/crud_test.clj
agzam: I'd rather choose not to hide it behind an abstraction and let the callers set query parameters directly on every level. One thing maybe we could do is to make `:mode` field optional and if it's not set - use `:query_string` by default. e.g.:

```
(check-fn
  {:query {:full-text {:query "title1"}
           :filter-map {:confidence "High"}}
```

ereteog: > One thing maybe we could do is to make :mode field optional and if it's not set - use :query_string by default.

Yes, I like it.

test/ctia/test_helpers/crud.clj
```diff
@@ -358,18 +356,19 @@
...
       (when delete-search-tests?
-        (test-delete-search app
-                            entity-str
-                            (keyword (string/replace plural #"-" "_"))
-                            example))
+        (th.search/test-delete-search
+         {:app        app
+          :entity     entity-str
+          :bundle-key (keyword (string/replace plural #"-" "_"))
+          :example    example}))
```
ereteog, line 371: nice that's easier to read this way!

src/ctia/http/routes/common.clj
```diff
@@ -117,19 +117,20 @@
      (cond-> {}
        (seq date-range) (assoc-in [:range date-field] date-range)
        (seq filter-map) (assoc :filter-map filter-map)
```
ereteog, line 119: few lines above filter-map is prepared as below:
``` clojure
filter-map (apply dissoc search-params filter-map-search-options)
```
with `filter-map-search-options` (probably not a good name) containing the search control parameters:
``` clojure
(def search-options [:sort_by
                     :sort_order
                     :offset
                     :limit
                     :fields
                     :search_after])
                     
(def filter-map-search-options
  (conj search-options :query :from :to))
```

We need to add :query_mode to `search-options`

ereteog (commented):
You need to add tests to check this PR:
- extends [`make-search-query-test`](https://github.com/threatgrid/ctia/blob/dbdef19688db0db59805aa071d87d8b812bd17a8/test/ctia/stores/es/crud_test.clj#L229) to check that the proper query is generated for each mode, and that new options are accepted.
- I forsee a bug :-), we finally need an end to end test, I would introduce one for any entity like judgement for instance. The undetected bug is that you do not remove `query_mode` from the filter-map. For Instance, if we have a search on incident that filters on `source` and a `full-text` search with `simple_query_string`, it should fail because `query_mode` parameter will be in the `filter-map`.

test/ctia/http/generative/fulltext_search_test.clj
```diff
@@ -0,0 +1,223 @@
...
+                         (gen/fmap #(dissoc % :id)))
+                    (gen/vector 5 11)))
+              entity-keys)]
+    (gen/bind
+     (apply gen/tuple gens)
+     (fn [entity-maps]
+       (gen/return
+        (zipmap entity-keys entity-maps))))))
+
+(defn- bundle-gen-for
```
ereteog, line 57: why do you need a generator?
From what I red, you only search on given fields for which you set a fix value, don't you?
Using CTIM minimal examples should be enough.

ereteog (commented):
It seems globally ok and well tested. I made few comments, mainly to simplify the tests and one to rename a query parameter.
Can you feel the QA section, please?

test/ctia/http/generative/fulltext_search_test.clj
```diff
@@ -0,0 +1,223 @@
...
+                         (is (= (-> res :parsed-body count)
+                                (-> bundle (get entity) count))))}]
+   (let [bundle   (gen/fmap
+                   (fn [bundle]
+                     (update
+                      bundle :incidents
+                      (fn [incidents]
+                        (apply
+                         utils/update-items incidents
+                         (repeat 3 #(assoc % :title "nunc porta vulputate tellus"))))))
```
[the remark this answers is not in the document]
agzam: Ah, because I need to update only 3, but I see now how I can do it with using clojure.core functions.

test/ctia/http/generative/fulltext_search_test.clj
agzam: The idea was to test the search with random values of different entities, i.e.:
 
- create a few random assets, incidents, etc. 
- change some fields in chosen entities (e.g.: set the Title for two incidents to be specific, while the rest would remain random values)
- issue a query like: "get me entities with specific title"
- ensure that only two relevant incidents come back (even though we have more)

test/ctia/http/generative/fulltext_search_test.clj
```diff
@@ -0,0 +1,223 @@
...
+   ;; Test `AND` (set two different fields that contain the same word in the same entity)
+   (let [bundle        (gen/fmap
+                        (fn [bundle]
+                          (update
+                           bundle :incidents
+                           (fn [incidents]
+                             (utils/update-items
+                              incidents
+                              #(assoc % :discovery_method "Log Review"
+                                      :title "title of test incident")))))
```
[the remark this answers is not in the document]
agzam: Ah, no... that would set the values in every entity. I only need to update a single one, not all.

src/ctia/lib/utils.clj
```diff
@@ -173,3 +173,13 @@
...
           (partition-all 2 selectors)))
+
+(defn update-items
+  "Updates items in a sequential coll by applying functions one by one to each
+  item. Each function takes an old value and returns a new value. The total
+  number of updated items always equals the number of functions passed, and the
+  total number of items in the resulting coll remains unchanged."
+  [coll & fs]
+  (if (sequential? coll)
+    (map #(% %2) (concat fs (repeat identity)) coll)
```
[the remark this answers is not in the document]
agzam: > what's the purpose of `(repeat identity)` ?

As it says in the docstring, if it's called like this `(update-items [1 4 4 5] dec inc)`, it returns: `(2 3 4 5)`. See? It  applies `dec` to the first, then `inc` to the second, and leaves the remaining intact. `(repeat identity)` is for that.

    (concat [inc dec] (repeat identity)) => 

	(#function[clojure.core/inc]
	 #function[clojure.core/dec]
	 #function[clojure.core/identity]
	 #function[clojure.core/identity]
	 #function[clojure.core/identity]
	 #function[clojure.core/identity]
	 #function[clojure.core/identity]
	,,, etc

---
This function allows me to update only some items in a collection and to leave remaining unchanged.
Probably I need to add an example in the docstring.

src/ctia/stores/es/crud.clj
```diff
@@ -316,51 +316,60 @@ It returns the documents with full hits meta data including the real index in wh
...
         filter-terms (-> (ensure-document-id-in-map filter-map)
-                         q/prepare-terms)]
+                            q/prepare-terms)
+        full-text-q  (when full-text
+                       {(get full-text :query_mode :query_string) ;; if no :query-mode passed, use :query_string mode
+                        (merge
+                         (dissoc full-text :query_mode)
+                         (when (and default_operator
+                                    ;; multi_match queries in ES don't support default_operator
+                                    (not= :multi_match (:query_mode full-text)))
```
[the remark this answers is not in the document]
agzam: Initially did it as you suggested, but had to revert. I realized that query_mode sometimes can be `nil`, making it a bit too verbose if I do it the way you prefer, sorry.

src/ctia/stores/es/crud.clj
```diff
@@ -316,51 +316,61 @@ It returns the documents with full hits meta data including the real index in wh
...
```
ereteog, line 337: Ok, like I said, I let it to your appreciation.
Just in case, here is a proposal
```suggestion
        es_query (:query_mode full-text :query_string) ;; if no :query-mode passed, use :query_string mode
        full-text-q  (when full-text
                       {es_query
                        (merge
                         (dissoc full-text :query_mode)
                         (when (and default_operator
                                    ;; multi_match queries in ES don't support default_operator
                                    (contains? #{:simple_query_string :query_string} es_query))
```

test/ctia/http/generative/fulltext_search_test.clj
```diff
@@ -0,0 +1,223 @@
...
+     :bundle-gen       (gen/fmap
+                        (fn [bundle]
+                          (update
+                           bundle :incidents
+                           (fn [incidents]
+                             (apply
+                              utils/update-items incidents
+                              (repeat 3
+                                      #(assoc % :title "Etiam vel neque bibendum dignissim"
+                                              :assignees ["bibendum"]))))))
```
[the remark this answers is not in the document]
ereteog: ok you can forget this suggestion :-)

src/ctia/http/routes/crud.clj
```diff
@@ -163,6 +164,18 @@
...
+                          search-q-params
+                          {(s/optional-key :query_mode)
+                           (describe FullTextQueryMode "Elasticsearch Fulltext Query Mode. Defaults to query_string")
+
+                           ;; We cannot name the parameter :fields, because we already have :fields (part
+                           ;; of search-q-params). That key is to select a subsets of fields of the
+                           ;; retrieved document and it gets passed to the `_source` parameter of
+                           ;; Elasticsearch. For more: www.elastic.co/guide/en/elasticsearch/reference/current/mapping-source-field.html
+                           (s/optional-key :es_query_fields)
+                           (describe (st/get-in search-q-params [:fields]) "'fields' key of Elasticsearch Fulltext Query.")})
```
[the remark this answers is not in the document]
agzam: Let's create an issue, and I'll address this in a separate PR?

src/ctia/http/routes/common.clj
```diff
@@ -106,9 +108,12 @@
                     to :- (s/maybe s/Inst)]
                    (cond-> {}
                      from (assoc :gte from)
-                     to (assoc :lt to)))))
+                     to   (assoc :lt to)))))
   ([date-field
-    {:keys [query from to] :as search-params}
+    {:keys [query
+            from to
```
ereteog, line 114:
```suggestion
            from
            to
```

src/ctia/http/routes/crud.clj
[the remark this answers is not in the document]
ereteog: ok.

PR #1110, opened 2021-04-13
Improves error message of asset_ref resolution

agzam:
During AssetMapping/AssetProperties resolution, when it cannot resolve an
asset_ref with a transientID, it throws an ambiguous message.

This PR is to make it less ambiguous.

> **Epic** #
> Close #https://github.com/threatgrid/iroh/issues/5132
> Related #1089

<a name="qa">[§](#qa)</a> QA
============================

QA instructions are the same as in: https://github.com/threatgrid/ctia/pull/1089

ad517508 * origin/improve_asset_ref_resolution_errmsg__#iroh/5132 refactor set-asset-ref and add tests
c7ca53fc * 1110 improves error message of asset_ref resolution

### DIFF
--- a/src/ctia/entity/asset.clj
+++ b/src/ctia/entity/asset.clj
@@ -3,7 +3,7 @@
    [ctia.domain.entities :refer [default-realize-fn]]
    [ctia.http.routes.common :as routes.common]
    [ctia.http.routes.crud :refer [services->entity-crud-routes]]
-   [ctia.schemas.core :refer [APIHandlerServices def-acl-schema def-stored-schema]]
+   [ctia.schemas.core :as schemas :refer [APIHandlerServices def-acl-schema def-stored-schema]]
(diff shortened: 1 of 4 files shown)


ereteog (commented):
Nice!

I would add a little test that proves the fix (even more if we ask for a cherry pick if we have time) and prevents regressions on this point.
To simply achieve this you can factorize the body of these 2 functions that are identical (perhaps in the asset ns?) and parametrize the only part that changes: the error message. Then you would simply have to add unit test on this function.

In order to help QA team, you should remove the external ids in the QA section of the related PR (bundle import first checks if entities having same external ids exist to avoid duplicates) and update the QA description with attended error message. Another option is to add a QA section in this PR with a scenario that focuses on this error.

ereteog (approved):
Nice!

hisans:
QA Verified in 1.72 Release

PR #1111, opened 2021-04-20
fix random index name clean

ereteog:
When using `fixture-ctia-with-app` we generate random indices to avoid collision between tests. 
We always had trailing indices in our local clusters since, I thought that it was due to failing tests, but indeed the cleaning was not triggered at all.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

### DIFF
--- a/test/ctia/test_helpers/es.clj
+++ b/test/ctia/test_helpers/es.clj
@@ -66,11 +66,11 @@
                                    :as services} :- ESConnServices]
   (let [{:keys [conn index]} (es-init/init-store-conn
                               (es-init/get-store-properties entity get-in-config)
-                              services)]
+                              services)
+        index-wildcard (str index "*")]
     (when conn
-      (doto (str index "*")
-        #(es-index/delete! conn %)
-        #(es-index/delete-template! conn %)))))
+        (es-index/delete! conn index-wildcard)
+        (es-index/delete-template! conn index-wildcard))))
 
 (defn fixture-purge-event-indices-and-templates
   "walk through all producers and delete their indices and templates"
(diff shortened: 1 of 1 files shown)


test/ctia/test_helpers/es.clj
```diff
@@ -66,11 +66,11 @@
...
                               (es-init/get-store-properties entity get-in-config)
-                              services)]
+                              services)
+        index-wildcard (str index "*")]
     (when conn
-      (doto (str index "*")
-        #(es-index/delete! conn %)
-        #(es-index/delete-template! conn %)))))
+        (es-index/delete! conn index-wildcard)
+        (es-index/delete-template! conn index-wildcard))))
```
ereteog, line 73:
```clojure
(doto 1 #(println "first" %) (println "second"))
1 second
;; => 1
```

test/ctia/test_helpers/es.clj
```diff
@@ -207,7 +207,7 @@
            :_id _id)))
 
 (defn load-bulk
-  ([conn docs] (load-bulk conn docs "true"))
+  ([conn docs] (load-bulk conn docs "wait_for"))
```
ereteog, line 210: this ensure that the request actually `wait_for` the index refresh and does not only triggers it.

agzam (approved):
Thank you!

test/ctia/test_helpers/es.clj
frenchy64: Ha! My bad.

frenchy64 (approved):
Excellent, thanks.

PR #1112, opened 2021-04-22
use python 3 in travis

ereteog:
We should try travis build with python 3.
Python 2.7 will not work soon:
https://github.com/pypa/pypi-support/issues/978

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: clojure
 lein: 2.9.1
 dist: trusty
+python: 3.6.1
 
 branches:
   only:


.travis.yml
```diff
@@ -1,6 +1,7 @@
 language: clojure
 lein: 2.9.1
 dist: trusty
+python: 3.6.1
```
ereteog, line 4: https://docs.travis-ci.com/user/reference/trusty/#python-images

vemv (approved):
LGTM, worst-case scenario is reverting 1LOC :)

PR #1121, opened 2021-05-05
Schema helpers for Trapperkeeper graphs

frenchy64:
Introduces a few schema helpers to help dedupe Trapperkeeper-related schemas.

1. `open-service-schema`: like `open-schema` but for service graph schemas
2. `select-service-subschema`: slice up service graph _schemas_

Also changes the syntax for `select-service-subgraph` to take a map rather than kw args and either take a schema as a second argument or a map of sets.

eg.,

```clojure
(s/defschema HTTPShowServices
  (-> APIHandlerServices
      (csu/select-service-subschema
        {:ConfigService #{:get-in-config}
         :CTIAHTTPServerService #{:get-port}})
      csu/open-service-schema))
```

```clojure
(s/defn app->HTTPShowServices :- HTTPShowServices [app]
  (-> app
      app/service-graph
      (select-service-subgraph HTTPShowServices)))
```

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: Schema helpers for Trapperkeeper graphs
```


agzam (approved):
Looks good!

src/ctia/schemas/utils.clj
```diff
@@ -18,55 +19,169 @@
...
+                                                          (s/explicit-schema-key fn-kw))
+                                                  (if-some [svc-fn (get gval (s/explicit-schema-key fn-kw))]
+                                                    {(s/explicit-schema-key fn-kw) svc-fn}
+                                                    (when-not (s/optional-key? fn-kw)
+                                                      (throw (ex-info (str "Missing " service-kw " service function: "
+                                                                           (s/explicit-schema-key fn-kw))
+                                                                      {}))))))
+                                           fn-kws)]
+                     {(s/explicit-schema-key service-kw) service-fns})))))
+        selectors))
```
ereteog, line 82: I am somehow lost in this conditional tree :-)
Could you split it a little bit to ease its reading.

src/ctia/schemas/utils.clj
```diff
@@ -18,55 +19,169 @@
...
+  "
+  [graph-schema :- (s/protocol s/Schema)
+   selectors :- {(s/pred simple-keyword?)
+                 #{(s/pred simple-keyword?)}}]
+  {:pre [(map? graph-schema)]}
+  (into {}
+        (map (fn [[service-kw fn-kws]]
+               (assert (keyword? service-kw)
+                       (pr-str service-kw))
+               (assert (set? fn-kws))
```
ereteog, line 145: aren't these assert a duplicate of the schemas?

ereteog (commented):
Nice! All these helpers are handy!
`service-subgraph` is not easy to read, could you split it, please?
Most of the `assert` appears to be duplicate of the schemas, are there necessary?

src/ctia/schemas/utils.clj
```diff
@@ -18,55 +19,169 @@
...
-     (if (and (map? x) (not (record? x)))
-       (optional-keys x)
-       x))
-   schema))
+  (st/optional-keys-schema schema))
+
+(s/defschema SpecificKey
+  (s/pred s/specific-key?))
+
+(s/defn service-subgraph :- (s/pred map?)
```
yogsototh, line 41: I wonder if it would be easy to test the value of your second argument to have a single function polymorphic in the second argument type.

If a selector `{SpecificKey #{SpecificKey}}` then use this implementation.
If a service schema => then use the implementation from `service-subgraph-from-schema`.

I am generally against this. But here, as the usage will be quite specific, it would make sense to me to use the same function name in both case `service-subgraph` and let it decide for me if it should be a `service-subgraph-from-schema` or `service-subgraph-from-service-subgraph-selector`.

src/ctia/schemas/utils.clj
```diff
@@ -18,55 +19,169 @@
...
+          (map (fn [[service-kw service-fns]]
+                 {:pre [(map? service-fns)]}
+                 (when (s/specific-key? service-kw)
+                   {service-kw
+                    (into #{}
+                          (filter s/specific-key?)
+                          (keys service-fns))})))
+          schema)))
+
+(s/defn service-subschema :- (s/protocol s/Schema)
```
yogsototh, line 120: hmmm I think I would use `service-schema-select` to capture the idea that this is not "check if service-schema-1 is a sub-schema of service-schema-2".

yogsototh (approved):
LGTM! And looks like nice operators on ServiceGraphs and ServiceGraphsSchemas.

Looking at these, I am under the impression you would like to have a few more operator to compose them differently.

Like a specific `merge`/`assoc`/`dissoc`

And probably; `(dissoc-service-fn [service-graph service-name & service-fn-keys]...)` and I think you already have a few helpers like these elsewhere.
The nice property of schema is that some of these functions should have the same API for a `service-graph` and a `service-graph-schema`.

src/ctia/schemas/utils.clj
frenchy64: I named the conditions and added a comment, does that help?

src/ctia/schemas/utils.clj
frenchy64: Yes, I made the schema more specific via `SimpleKeywordSpecificKey` and removed the asserts.

src/ctia/schemas/utils.clj
frenchy64: Thanks, added a `select-*` prefix to all the functions.

src/ctia/schemas/utils.clj
frenchy64: I merged them as it's just a matter of testing for set vs. map. This also solves the performance problem I had commented in `service-subgraph-from-schema` where the schema was traversed twice--now it's a single pass.

src/ctia/schemas/utils.clj
```diff
@@ -18,55 +19,169 @@
...
-   (let [ks? (explicit-key-set ks)]
-     (stu/map-keys
-      (fn [k]
-        (cond
-          (and ks (not (ks? (explicit-key k)))) k
-          (s/specific-key? k) (f (s/explicit-schema-key k))
-          :else k))
-      schema))))
-
+;; TODO delete me
```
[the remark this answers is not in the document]
frenchy64: I went ahead and deleted them--adds a few dozen files to the diff.

ereteog (approved):
LGTM

PR #1124, opened 2021-05-17
Redefine default search fields

agzam:
> **Epic** #
> Close https://github.com/threatgrid/iroh/issues/5151, https://github.com/threatgrid/iroh/issues/5269
> Related #

## Defines default fields for search.

Default fields for the full-text search are now all Entity fields with the exclusion of certain fields, like `id,` `type,` `language`, `tlp`, etc.

Users may choose to include or exclude fields as they see fit.

Basically, now when the search query is posted (and only `:query` parameter of the search provided), the search is performed in all the main (important) fields of the Entity, but excludes some second-class fields.

However, it is still possible to search in fields like `tlp` and `language` (and others) by explicitly including them in `:search_fields` parameter.

ezgif com-gif-maker

<a name="qa">[§](#qa)</a> QA
============================

### Test scenario #1

1. Create one or multiple Entities. [e.g., Incidents] with random values.
2. Set the `title` of one of the Incidents to be of some value, e.g.: `Fusce suscipit wisi nec facilisis`
2. Send a search request 
```
/ctia/incident/search?query=*suscipit*
```
3. **Expected**: It should find the Incident

### Test scenario #2

1. Create one or multiple Entities. [e.g., Incidents] with random values.
2. Set the `language` of one of the Incidents to be of some value, e.g.: `Esperanto`
2. Send a search request 
```
/ctia/incident/search?query=*Esperanto*
```
3. **Expected**: Search should yield no results, since `language` is one of the fields that are ignored by default
4. Send another search request, this time explicitly providing the field
```
/ctia/incident/search?query=*Esperanto*&search_fields=language
```
5. **Expected**: It should return the Incident

### DIFF
--- a/src/ctia/entity/sighting/schemas.clj
+++ b/src/ctia/entity/sighting/schemas.clj
@@ -78,4 +78,5 @@
 
 (def sighting-sort-fields
   (conj sighting-fields
-        "observed_time.start_time,timestamp"))
+        :observed_time.start_time
+        :timestamp))
--- a/src/ctia/http/routes/common.clj
+++ b/src/ctia/http/routes/common.clj
@@ -7,8 +7,6 @@
    [ctia.schemas.search-agg :refer
     [FullTextQueryMode MetricResult RangeQueryOpt SearchQuery]]
    [ctia.schemas.sorting :as sorting]
-   [ctia.schemas.utils :as schemas.utils]
-   [ctia.stores.es.mapping :as em]
    [ring.swagger.json-schema :as json-schema]
    [ring.swagger.schema :refer [describe]]
    [ring.util.codec :as codec]
@@ -28,6 +26,29 @@
 
 (def filter-map-search-options
   (conj search-options :query :from :to))
+
+(def insignificant-search-fields
+  "Fields to be ignored by default when searching"
+  #{:authorized_groups
+    :authorized_users
+    :created
+    :external_ids
+    :external_references
+    :from
+    :groups
+    :id
+    :language
+    :modified
+    :owner
+    :revision
+    :schema_version
+    :source
+    :source_ref
+    :source_uri
+    :timestamp
+    :tlp
+    :to
+    :type})
 
 (s/defschema BaseEntityFilterParams
   {(s/optional-key :id) s/Str
@@ -114,7 +135,7 @@
   [search-query-params :- (s/maybe (s/protocol s/Schema))]
   (let [fields-schema  (st/get-in search-query-params [:fields])
         default-fields (-> fields-schema first :vs
-                           (set/difference (set (keys em/base-entity-mapping)))
+                           (set/difference insignificant-search-fields)
                            vec)]
    (st/merge
     search-query-params
--- a/src/ctia/http/routes/common.clj
+++ b/src/ctia/http/routes/common.clj
@@ -38,9 +38,10 @@
    (s/optional-key :tlp) s/Str})
 
 (s/defschema SourcableEntityFilterParams
-  {(s/optional-key :source) s/Str
+  {(s/optional-key :source) s/Str})
 
-   (s/optional-key :query_mode)
+(s/defschema SearchEntityParams
+  {(s/optional-key :query_mode)
    (describe FullTextQueryMode "Elasticsearch Fulltext Query Mode. Defaults to query_string")})
 
 (s/defschema PagingParams
--- a/src/ctia/http/routes/common.clj
+++ b/src/ctia/http/routes/common.clj
@@ -2,14 +2,19 @@
   (:require
    [clj-http.headers :refer [canonicalize]]
    [clj-momo.lib.clj-time.core :as t]
+   [clojure.set :as set]
    [clojure.string :as str]
    [ctia.schemas.search-agg :refer
     [FullTextQueryMode MetricResult RangeQueryOpt SearchQuery]]
    [ctia.schemas.sorting :as sorting]
+   [ctia.schemas.utils :as schemas.utils]
+   [ctia.stores.es.mapping :as em]
+   [ring.swagger.json-schema :as json-schema]
    [ring.swagger.schema :refer [describe]]
    [ring.util.codec :as codec]
    [ring.util.http-response :as http-res]
    [ring.util.http-status :refer [ok]]
+   [schema-tools.core :as st]
    [schema.core :as s]))
 
 (def search-options [:sort_by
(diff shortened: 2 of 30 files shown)


src/ctia/http/routes/common.clj
```diff
@@ -32,7 +38,10 @@
    (s/optional-key :tlp) s/Str})
 
 (s/defschema SourcableEntityFilterParams
-  {(s/optional-key :source) s/Str})
+  {(s/optional-key :source) s/Str
+
+   (s/optional-key :query_mode)
+   (describe FullTextQueryMode "Elasticsearch Fulltext Query Mode. Defaults to query_string")})
```
ereteog, line 44: "Sourcable" is related to fields that enable to identify the source of the identity, not to search it.
There is only source for now, but I still hope that one day we will get closer to the provenance approach of the Semantic Web 😉

src/ctia/http/routes/common.clj
```diff
@@ -99,6 +108,28 @@
     {:gte from
      :lt to-or-now}))
 
+(s/defn prep-es-fields-schema :- (s/protocol s/Schema)
+  "Conjoins ES :fields onto search-parameters."
+  [search-query-params :- (s/maybe (s/protocol s/Schema))]
+  (let [fields-schema  (st/get-in search-query-params [:fields])
```
ereteog, line 114: We need to review the `fields` value in each entity to ensure that they really cover the fields we want to search in.

src/ctia/http/routes/common.clj
```diff
@@ -99,6 +108,28 @@
...
+        default-fields (-> fields-schema first :vs
+                           (set/difference (set (keys em/base-entity-mapping)))
```
ereteog, line 116: We cannot rely on the store layer in a route layer. Moreover this is specific to the es store. There is a schema for these base fields.

ereteog (commented):
Quick feedback. It seems ok.. However that's only the tests which prove the feature is properly implemented.
If no field is specified, does the search match terms in any field with respect to its mapping?

src/ctia/http/routes/common.clj
agzam: Hmm. I think I misunderstood your comment here: https://github.com/threatgrid/ctia/pull/1109#discussion_r625897294

src/ctia/http/routes/common.clj
agzam: What do you mean? Just a manual audit or something? That's easy, we can open Swagger and it will show all the fields for every entity where the search will be performed (by default)

src/ctia/http/routes/common.clj
agzam: Ah, right. Good point. 👍

src/ctia/http/routes/common.clj
ereteog: > Just a manual audit or something? 

Yes somehow. `fields` was not particularly leveraged to retrieve partial documents on search. So it's very likely that some interesting fields to search in are not available.

src/ctia/http/routes/common.clj
ereteog: Like `SourcableEntityFilterParams` you would have `SearchableEntityParams`

src/ctia/entity/sighting/schemas.clj
```diff
@@ -78,4 +78,5 @@
 
 (def sighting-sort-fields
   (conj sighting-fields
-        "observed_time.start_time,timestamp"))
+        :observed_time.start_time
+        :timestamp))
```
ereteog, line 82: This sort field is intentionally exposed as `"observed_time.start_time,timestamp"` for CTR which needs to sort sigthings first on `observed_time.start_time` then `timestamp`. I do not have the full context, but any sort option exposed like this is a requirement of an API client (SecureX, CTR, IROH, Engine, Ribbon, and still counting ;-) ).

src/ctia/http/routes/common.clj
```diff
@@ -23,6 +28,29 @@
 (def filter-map-search-options
   (conj search-options :query :from :to))
 
+(def unimportant-search-fields
```
ereteog, line 31:
```suggestion
(def ignored-search-fields
```

src/ctia/http/routes/common.clj
```diff
@@ -99,6 +131,96 @@
     {:gte from
      :lt to-or-now}))
 
+(defn- schema->es-fields
+  "Sifts out the keys that can be used as :search_fields."
+  [schema]
+  (->>
+   schema
+   csu/schema->keys
+   (reduce
```
ereteog, line 140: that's a filter, isn't it?

src/ctia/http/routes/common.clj
```diff
@@ -99,6 +131,96 @@
...
+     ;;
+     ;; For backward-compatibility we keep the old name and add new key with a different name
+     (s/optional-key :search_fields)
+     (json-schema/field
+      [(apply s/enum default)]
+      {:default     default
+       :description "'fields' key of Elasticsearch Fulltext Query."})})))
+
+(defn ensure-search-fields
+  "It gurantees that ES :fields is always passed into ES
```
ereteog, line 212:
```suggestion
  "It guarantees that ES :fields is always passed into ES
```

test/ctia/http/generative/fulltext_search_test.clj
```diff
@@ -185,7 +185,61 @@
...
+                           (fn [incidents]
+                             (utils/update-items
+                              incidents
+                              #(assoc % :language "Etiam vel neque bibendum dignissim")))))
+                        (bundle-gen-for :incidents))
+     :check            (fn [_ _ _ res]
+                         ;; it shouldn't find anything, when :search_fields not
+                         ;; provided, it would only search in 'default' fields,
+                         ;; and the :language is excluded
+                         (is (-> res :parsed-body count zero?)))}
```
ereteog, line 226:
```suggestion
                         (is (-> res :parsed-body empty?)))}
```

test/ctia/http/generative/fulltext_search_test.clj
```diff
@@ -185,7 +185,61 @@
...
+                                                          (-> % :discovery_method (= "bibendum")))))]
+                           (is (= 3 (count matching)))))}
+    {:test-description ":search_fields defaults. Should ignore fields like :id, :tlp, :language, etc."
+     :query-params     {:query "*bibendum*"}
+     :bundle-gen       (gen/fmap
+                        (fn [bundle]
+                          (update
+                           bundle :incidents
+                           (fn [incidents]
+                             (utils/update-items
```
ereteog, line 218: please move this function in a test ns.

ereteog (changes requested):
In the issue I initially asked for an allow list per entity.
I am now more convinced that you need to stick to the initial plan:
- If we add any system field in the future we will need to include it here. If we forgot to remove it and a client start using it it will be harder to remove it. It probably looks like an improbable scenario, but it already happened in a similar case. The opposite case would be less painful, if we introduce a new searchable field and forget to add it, we will just have to add it to fix it.
- We have no control with the auto generation of the default field on the search precision per entity, while an allow list per entity enable us to make a full review of relevant fields to include.
- it introduces unnecessary complexity on the code, we prefer easy to read boiler plate.

agzam:
After a discussion, we decided not to try to over-engineer it with the purpose of fixing https://github.com/threatgrid/iroh/issues/5326 and use a more straightforward approach. Closing this in favor of https://github.com/threatgrid/ctia/pull/1127

hisans:
QA Verified in 1.74 Release

PR #1099, opened 2021-03-16
Issue-4744: Add endpoint to search vulnerabilities by CPEs

halcyon:
Replaces pull #1077 

Epic threatgrid/iroh#4744

Related threatgrid/iroh#4789

Add route to search for vulnerabilities configurations which match combinations of CPE Match Strings.

Proposed route:
GET /ctia/vulnerability/cpe_match_strings?cpe23_match_strings=[cpe_match_string]

PR #1127, opened 2021-05-29
Redefine default search fields #iroh/5151  second try

agzam:
Define text search fields for entities

> **Epic** https://github.com/advthreat/iroh/projects/14
> Close https://github.com/threatgrid/iroh/issues/5151
> Related #

<a name="qa">[§](#qa)</a> QA
============================

*QA plan for this ticket should be very similar to one in: https://github.com/threatgrid/ctia/pull/1109 Mostly, we need to make sure there's no regression and the search works as before.*

*We encourage you to think of various query strategies and send as many different queries as possible. Saturate the test cases to flush out things that we may have missed. The following outline is an example of a few different queries.*

Naturally, you're going to need some data to query against it. You can use Bundle Import to create a few entities. 

- Mark your entities with some unique group name, e.g., the following examples set the value of `source` field for every entity to be `pr1127`

- Whenever you need to re-run the tests, it's probably desirable to remove any existing entities, to avoid confusion:

```http
DELETE http://private.intel.int.iroh.site/ctia/incident/search?source=pr1127&query=*&REALLY_DELETE_ALL_THESE_ENTITIES=true

DELETE http://private.intel.int.iroh.site/ctia/actor/search?source=pr1127&query=*&REALLY_DELETE_ALL_THESE_ENTITIES=true
```

- Make sure you don't have any records for entities you're about to test. The following queries should return 0.

```http
GET http://private.intel.int.iroh.site/ctia/incident/search/count&source=pr1127

GET http://private.intel.int.iroh.site/ctia/actor/search/count&source=pr1127
```

### Import a bundle
    
<details>
    <summary>
    Use bundle import to create some entities, the following example bundle contains three incidents, and a single actor:
   </summary>

```
POST http://private.intel.int.iroh.site/ctia/bundle/import
Content-type: application/json

{
    "incidents": [
        {
            "confidence": "High",
            "tlp": "green",
            "language": "language",
            "status": "Open",
            "categories": [
                "Denial of Service",
                "Improper Usage"
            ],
            "intended_effect": "Extortion",
            "source_uri": "http://example.com/incident-one",
            "discovery_method": "Log Review",
            "external_references": [],
            "incident_time": {
                "reported": "2016-02-11T00:40:48Z",
                "remediated": "2016-02-11T00:40:48Z",
                "rejected": "2016-02-11T00:40:48Z",
                "opened": "2016-02-11T00:40:48Z",
                "discovered": "2016-02-11T00:40:48Z",
                "closed": "2016-02-11T00:40:48Z"
            },
            "title": "Lorem Ipsum Test Incident",
            "short_description": "first incident",
            "external_ids": [],
            "source": "pr1127",
            "type": "incident",
            "revision": 1,
            "schema_version": "1.1.3",
            "assignees": [],
            "description": "description of first incident"
        },
        {
            "confidence": "High",
            "tlp": "green",
            "language": "language",
            "status": "Open",
            "categories": [
                "Denial of Service",
                "Improper Usage"
            ],
            "intended_effect": "Extortion",
            "source_uri": "http://example.com/incident-two",
            "discovery_method": "Log Review",
            "external_references": [],
            "incident_time": {
                "reported": "2016-02-11T00:40:48Z",
                "remediated": "2016-02-11T00:40:48Z",
                "rejected": "2016-02-11T00:40:48Z",
                "opened": "2016-02-11T00:40:48Z",
                "discovered": "2016-02-11T00:40:48Z",
                "closed": "2016-02-11T00:40:48Z"
            },
            "title": "Lorem Ipsum Test Incident dos",
            "short_description": "incident numero dos",
            "external_ids": [],
            "source": "pr1127",
            "type": "incident",
            "revision": 1,
            "schema_version": "1.1.3",
            "assignees": [],
            "description": "second incident"
        },
        {
            "confidence": "High",
            "tlp": "green",
            "language": "language",
            "status": "Open",
            "categories": [
                "Denial of Service",
                "Improper Usage"
            ],
            "intended_effect": "Extortion",
            "source_uri": "http://example.com/incident-three",
            "discovery_method": "Log Review",
            "external_references": [],
            "incident_time": {
                "reported": "2016-02-11T00:40:48Z",
                "remediated": "2016-02-11T00:40:48Z",
                "rejected": "2016-02-11T00:40:48Z",
                "opened": "2016-02-11T00:40:48Z",
                "discovered": "2016-02-11T00:40:48Z",
                "closed": "2016-02-11T00:40:48Z"
            },
            "title": "incident numero three",
            "short_description": "incident numero tres",
            "external_ids": [],
            "source": "pr1127",
            "type": "incident",
            "revision": 1,
            "schema_version": "1.1.3",
            "assignees": [],
            "description": "third incident"
        }
    ],
    "actors": [
        {
            "description" : "Legion Of Doom",
            "motivation" : "Ego",
            "valid_time" : {
                "start_time" : "2016-02-11T00:40:48Z",
                "end_time" : "2016-07-11T00:40:48Z"
            },
            "identity" : {
                "description" : "LOD identity",
                "related_identities" : [ ]
            },
            "sophistication" : "Innovator",
            "schema_version" : "1.0",
            "revision" : 1,
            "planning_and_operational_support" : "MUHS 9th Hour",
            "type" : "actor",
            "source": "pr1127",
            "external_ids" : [ ],
            "short_description" : "founded by the hacker Lex Luthor",
            "title" : "LOD Group",
            "external_references" : [ ],
            "source_uri" : "http://example.com/actors/legion-of-doom",
            "intended_effect" : "Fraud",
            "actor_type" : "Hacker",
            "language" : "language",
            "tlp" : "green",
            "confidence" : "High"
        }],
    "tlp": "green",
    "language": "language",
    "source_uri": "http://example.com",
    "title": "title",
    "short_description": "bundle",
    "source": "source",
    "revision": 1,
    "schema_version": "1.1.3",
    "valid_time": {
        "start_time": "2016-05-11T00:40:48Z",
        "end_time": "2016-07-11T00:40:48Z"
    },
    "description": "incidents to test"
}
```

</details>

- After the Import, check the number of imported entities

```http
GET http://private.intel.int.iroh.site/ctia/incident/search/count?source=pr1127
```
Expected: 3

```http
GET http://private.intel.int.iroh.site/ctia/actor/search/count?source=pr1127
```
Expected: 1

#### 1. Query for non-existing values

```http
GET http://private.intel.int.iroh.site/ctia/actor/search?source=pr1127&query=lorem_ipsum_suspendisse_potenti
```

Expected: empty list

#### 2. Query including non-existing fields

Even though the following query expected to succeed:
```http
GET http://private.intel.int.iroh.site/ctia/actor/search?source=pr1127&query=LOD
```

This next one is expected to fail
```http
GET http://private.intel.int.iroh.site/ctia/actor/search?source=pr1127&query=LOD&search_fields=some_strange_field
```

Expected: Error - unexpected search field

#### 3. Search for Incidents that have specific values in specified fields

The following query should fail:

```http
GET http://private.intel.int.iroh.site/ctia/incident/search?source=pr1127&query=first%20Ipsum&search_fields=description
```

However, if we add the `title` field (without changing the query), it should find the record

```http
GET http://private.intel.int.iroh.site/ctia/incident/search?source=pr1127&query=first%20Ipsum&search_fields=description&search_fields=title
```

#### 4. Looking for the same value in different fields in multiple records

The next query returns two incidents, because two incidents have "numero" in their `short_description` fields

```http
GET http://private.intel.int.iroh.site/ctia/incident/search?source=pr1127&query=numero
```

#### 5. Now, if we want to match a record with "numero" in multiple fields, we still can use Lucene syntax, like so:

  This query `short_description:numero AND title:numero` should return only a single incident
  
```http
GET http://private.intel.int.iroh.site/ctia/incident/search?source=pr1127&query=short_description%3Anumero%20AND%20title%3Anumero
```

But changing the query slightly: `short_description:numero OR title:numero`, again gets both incidents
  
```http
GET http://private.intel.int.iroh.site/ctia/incident/search?source=pr1127&query=short_description%3Anumero%20OR%20title%3Anumero
```

```
```

e892f2ca fix typo
a99a6a3c adds additional test case
dac426e5 test for all searchable-fields to exist in the schema
937fd8fc removes unused reference
3eeb0b4e event is not a sourceable entity
fa86133e extracts searchable-fields to a separate var
79c7be20 refactor the test
412b56bc simplify threading
b6d2c6e9 added couple more tests cases
7bf1e0b0 test for status code, not the message
1100ef0e simplify conditional
97440e4f typo fix
6d1ccf2b searchable fields audit, based on PR feedback
e5e1f1f2 put full-text-search-schema-functions test back
2273808c removes schema->all-keys fn, not needed anymore
6ce29d8c simplifies tests by switching from bundle to bulk
c98d46b2 adds a few more tests cases
fa322eed defines searchable fields manually
48792b74 formatting
b297e1f2 more fields to ignore
2f42d7f0 ignore schema keys that are not keywords
8bc5f42d updates schema->all-keys so it properly picks up deeply nested keys
55056f52 subtle bug where schema->all-keys could skip certain keys
849cbfb9 ignore a few nested fields
dbc79b4a temporarily disable failing test
fa3157d2 ignore a few nested fields
5d1eea0c disable nested vulnerability fields
8cdbb1f7 updates schema->all-keys to traverse over nested keys
d89c8bea moves auxiliary fn to dedicated namespace
48b023be fix tests after changing fn signature
3ffdd046 disabling judgement test for searching in nested fields
21ae6771 vulnerability fields to ignore
71d11735 adds some fields to ignore
15336735 attempt to change functions signature
c4f1ea86 adds another test
7505dcaa test and docstrings
2d345300 fixes flawed test
b593e4b0 fixes the stop-word text query test
a7b29028 more intuitive function signature
9303f831 adds a check
e673fc9f temporariy disable failing tests
cf703074 adds whitelisted-search-fields
4b881043 add fields to ignore
380df2df search_fields enforcement
d201d79b move :searchable-fields to entity routes
7df0c62a adds searchable-fields to Entity schema to appease failing validation
d23ddbb0 introduce explicit searchable fields for every entity


src/ctia/entity/campaign.clj
```diff
@@ -124,7 +124,12 @@
     :external-id-capabilities :read-campaign
     :can-aggregate?           true
     :histogram-fields         campaign-histogram-fields
-    :enumerable-fields        campaign-enumerable-fields}))
+    :enumerable-fields        campaign-enumerable-fields
+    :searchable-fields        (routes.common/searchable-fields
+                               campaign-fields
+                               :valid_time.start_time
+                               :valid_time.end_time
+                               :activity.date_time)}))
```
ereteog, line 132: this is indexed a date, just use fields indexed as text and keyword

src/ctia/entity/casebook.clj
```diff
@@ -335,7 +335,10 @@
      :external-id-capabilities :read-casebook
      :hide-delete?             false
      :histogram-fields         casebook-histogram-fields
-     :enumerable-fields        casebook-enumerable-fields})))
+     :enumerable-fields        casebook-enumerable-fields
+     :searchable-fields        (routes.common/searchable-fields
+                                (concat casebook-fields
+                                        casebook-enumerable-fields))})))
```
ereteog, line 341: we said allow list, you do not control anymore this list like this

src/ctia/entity/campaign.clj
agzam: These are the fields to be ignored. I think I should make that clearer by changing the function's signature, so it is called maybe like this:

```
(searchable-fields {:fields campaign-fields :ignore [:valid_time.start_time :activity.date_time]) 
```

agzam: https://github.com/threatgrid/ctia/pull/1127/commits/35174d6947e9a11548cc68ea1d54d4264e9fe9c3

test/ctia/test_helpers/search.clj
```diff
@@ -152,7 +152,7 @@
       (is (empty? (search-ids app entity "wor"))))
 
     ;; test stop word filtering
-    (is (= matched-ids (search-ids app entity "the word"))
+    #_(is (= matched-ids (search-ids app entity "the word"))
```
agzam, line 155: @ereteog I've been trying to understand why this thing is failing (and I think the second case on L:192 is relevant too)

It looks like when `fields` are used, things just don't work (as before) with `AND` operator. Compare these queries:

      "query_string" : {
          "default_operator" : "AND",
          "fields" : ["description",
                      "short_description",
                      "title",
                      "promotion_method"
                      ],
          "query" : "the incident"
      }
      
      ; => no results
      
      ; However,

      "query_string" : {
          "default_operator" : "AND",
          "query" : "the incident"
      }
      
      ; works
      ; and:
      
      "query_string" : {
          "default_operator" : "OR",
          "fields" : ["description",
                      "short_description",
                      "title",
                      "promotion_method"
                      ],
          "query" : "the incident"
       }
      
      ; also works
     
I can cheat here and simply modify the tests so they use "OR" operator instead, but would that impact anything we have in prod?

ereteog: I prefer to understand first.
It's probably related to: https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#query-string-multi-field-parms

agzam: Sounds good. If I dig up something interesting, I'll let you know.

src/ctia/http/routes/common.clj
```diff
@@ -50,6 +53,65 @@
...
+
+(def default-whitelisted-search-fields
+  #{:id
+    :title
+    :short_description
+    :description
+    :external_ids})
+
+(s/defn searchable-fields :- (s/protocol s/Schema)
+  [{:keys [fields ignore]}]
```
ereteog, line 74: why not simply provide the list of searchable fields?
``` clojure
(routes.common/searchable-fields
                               (dissoc (set actor-fields)
                                            :valid_time.start_time
                                            :valid_time.end_time))
```

ereteog (changes requested):
* You need to add dedicated tests, adding a filter on `description` is not enough.
You should add an isolated test and not build on top of the current search test + unit test the function you introduce in https://github.com/threatgrid/ctia/pull/1127/files#diff-1c0b76213a276bc9592c3e82a560f9abed7d0ca4b4cffb29b0a4f26f57f7e3b2.
* The QA section must be deterministic without deleting the whole store.
You could add a unique filter, like adding the value `pr1127` to each document and filter each search query with `source=pr1127`.

src/ctia/http/routes/common.clj
agzam: > why not simply provide the list of searchable fields?

If you remember, I didn't want us to repeat the list of allowed and blacklisted fields for every single entity, 

So it doesn't look like this (in this case for Actor):

    (routes.common/searchable-fields
     {:allowed-fields #{:actor_type
                        :confidence
                        :description
                        :external_ids
                        :id
                        :intended_effect
                        :motivation
                        :short_description
                        :sophistication
                        :title}
      :ignored-fields [:language
                       :revision
                       :schema_version
                       :source
                       :source_uri
                       :timestamp
                       :tlp
                       :type
                       :valid_time.end_time
                       :valid_time.start_time]})

Since we can have a global default whitelist and a global ignore list, we can just use them, and for every entity we can specifically tell which fields we want to whitelist or blacklist à la carte.

But I wanted to make it clear which fields (in addition to default ignore list) need to be ignored. So the function takes a list of fields, and [optionally] fields to ignore.

I guess I need to add some proper docstrings.

src/ctia/http/routes/common.clj
```diff
@@ -50,6 +53,65 @@
...
+    :source_uri
+    :timestamp
+    :tlp
+    :type})
+
+(def default-whitelisted-search-fields
+  #{:id
+    :title
+    :short_description
+    :description
```
ereteog, line 70: some entities are not [describable](https://github.com/threatgrid/ctim/blob/d9462cb83f656f04fc9a764033a3861b4e7dfb94/src/ctim/schemas/common.cljc#L234) like [AssetMapping](https://github.com/threatgrid/ctim/blob/d9462cb83f656f04fc9a764033a3861b4e7dfb94/src/ctim/schemas/asset_mapping.cljc#L23)

src/ctia/http/routes/common.clj
ereteog: I properly understood the process. I am ok with this implementation, but you pass a first collection of fields to add and then remove some of these fields, so IMHO simply pass the search-fields without those that should be ignored. That's the responsibility of the caller to only pass relevant search fields (and you automatically ensure the presence / absence of common fields if necessary).

there are:
- default field to include
- default field to exclude
- specific fields to add

So roughly I would expect
``` clojure
(difference (union default-whitelisted-search-fields search-fields)
                   default-ignored-search-fields)
```

agzam: > simply pass the search-fields without those that should be ignored.

I thought about it. In that case how do we precisely determine which fields should be ignored? If we rely on `default-ignored-search-fields`, then there's no way to explicitly tell not to ignore any specific field. Then we'd say: "what if we remove default-ignored-fields", in that case, for every entity we would have to explicitly list them, like so:

```
(routes.common/searchable-fields (dissoc weakness-fields :language
                                                         :revision
                                                         :schema_version
                                                         :source
                                                         :source_uri
                                                         :timestamp
                                                         :tlp
                                                         :type))
```

Or always ignore the fields in `default-ignored-fields` without a way of overriding that.

src/ctia/http/routes/common.clj
agzam: Ah, you're right. I'm gonna try something different. I'll try to get the fields not from the `fields` (because it doesn't actually include all the fields entity has), I'll try to grab it directly from the Entity schema.

agzam: @ereteog I shuffled a few things, if you can, please take another gander. 

I am now getting the fields from the entity schema, and it works now this way:

- You provide the schema and optionally fields to ignore
-  There's `default-ignored-search-fields`, for fields that always will be ignored. It was tempting to make it possible to explicitly allow  some fields (even if they are in that list), but this is not going to be used right now; If ever needed, the function can be extended later. 
- And we don't need `default-whitelisted-search-fields` anymore.

____

Something new I've discovered that I somehow missed previously, pls. take a look: https://github.com/threatgrid/ctia/pull/1127/files#diff-7d7da115f9a07f82eb46cabf135808b2835014f44605f098f3f8b152cc5d4f82R53

If you agree, I will incorporate that feature separately.

____

And then I will update the QA instructions.

ereteog: If we do it separately we have a risk of regression.

test/ctia/entity/judgement_test.clj
```diff
@@ -50,11 +50,15 @@
...
 
-      (let [term "1.2.3.4"
-            response (GET app
-                          (str "ctia/judgement/search")
-                          :headers {"Authorization" "45c1f5e3f05d0"}
-                          :query-params {"query" term})]
+      ;; TODO: this test no longer works, because apparently, in order to
+      ;; perform search for values within nested fields (without Lucene
+      ;; syntax), we're going to have to implement a way po pass nested fields, e.g.,:
+      ;; "query_string": {"query": "1.2.3.4", "fields": ["observable.*^1.0"] - note the asterisk
```
ereteog, line 56: why `*`?
Assuming you are referring to the wildcard example [here](https://www.elastic.co/guide/en/elasticsearch/reference/current/query-dsl-query-string-query.html#query-string-multi-field), it should rather be ["observable.type",  "observable.value"].
So here, if your code properly works, the final query shall include default fields and for judgement `observable.value` shall be included.

ereteog: moreover the[ initial value](https://github.com/threatgrid/ctia/blob/70dd7d013a5401bc3c93f0c942e221828de57067/test/ctia/entity/judgement_test.clj#L44) for `term` is  `observable.value:\"1.2.3.4\"`
Why did you modify it?

agzam: > Assuming you are referring to the wildcard example here, it should rather be ["observable.type", "observable.value"].

I agree. But that's what I'm saying. The logic in the current PR does not address nested fields (it doesn't even try to ascertain them). If you're saying we are at risk of introducing regression, then I better change it here, so it does recognize nested fields.

ereteog: 99% of test which do not pass are regressions.
If you look at the test which does not pass, it's clearly a regression.

ereteog:
Since the beginning, the main job of this task is to carefully select meaningful fields. I already mentioned it with the first version of the function you made to walk the schemas.
Some of the entities have numerous fields like vulnerability but most of them are very specific values which will be searchable though specific helpers (and [already implemented by scott ](https://github.com/threatgrid/ctia/pull/1099)).
Automatic inclusions of string fields will produce a lot of noise in the search and a lot of false positive. Then you will be back to manual seletion of ignored fields exclusion.
So, just make manual explicit list of of carefully selected searchable fields per entity.

src/ctia/entity/casebook.clj
```diff
@@ -334,7 +334,83 @@
...
+                                 :external_references.external_id
+                                 :external_references.hashes
+                                 :external_references.source_name
+                                 :external_references.url
+                                 :observables.type
+                                 :observables.value
+                                 :short_description
+                                 :texts.text
+                                 :texts.type
+                                 :title}})))
```
ereteog, line 413:
```suggestion
     :searchable-fields        #{:id
                                 :source
                                 :description
                                 :observables.type
                                 :observables.value
                                 :short_description
                                 :texts.text
                                 :texts.type
                                 :title}})))
```

src/ctia/entity/coa.clj
```diff
@@ -131,7 +131,45 @@
...
+                                :open_c2_coa.target.type
+                                :open_c2_coa.type
+                                :related_COAs.COA_id
+                                :related_COAs.confidence
+                                :related_COAs.relationship
+                                :related_COAs.source
+                                :short_description
+                                :stage
+                                :structured_coa_type
+                                :title}}))
```
ereteog, line 172:
```suggestion
    :searchable-fields        #{:id
                                :source
                                :coa_type
                                :cost
                                :description
                                :efficacy
                                :impact
                                :objective
                                :open_c2_coa.action.type
                                :open_c2_coa.actuator.specifiers
                                :open_c2_coa.actuator.type
                                :open_c2_coa.id
                                :open_c2_coa.modifiers.additional_properties.context
                                :open_c2_coa.modifiers.destination
                                :open_c2_coa.modifiers.frequency
                                :open_c2_coa.modifiers.id
                                :open_c2_coa.modifiers.location
                                :open_c2_coa.modifiers.method
                                :open_c2_coa.modifiers.option
                                :open_c2_coa.modifiers.response
                                :open_c2_coa.modifiers.search
                                :open_c2_coa.modifiers.source
                                :open_c2_coa.target.specifiers
                                :open_c2_coa.target.type
                                :open_c2_coa.type
                                :related_COAs.COA_id
                                :related_COAs.confidence
                                :related_COAs.relationship
                                :related_COAs.source
                                :short_description
                                :stage
                                :structured_coa_type
                                :title}}))
```

src/ctia/entity/event.clj
```diff
@@ -143,22 +143,28 @@
...
-     :date-field :timestamp})))
+     :search-capabilities     :search-event
+     :delete-capabilities     #{:delete-event :developer}
+     :date-field              :timestamp
+     :searchable-fields       #{:id
+                                :fields.change.after
+                                :event_type
+                                :fields.action
+                                :fields.change.before
+                                :fields.field}})))
```
ereteog, line 167: only some fields are indexed in events: https://github.com/threatgrid/ctia/blob/43ff86881b099bfd5a86f6a24a7ec0e05e980c04/src/ctia/entity/event.clj#L36
```suggestion
     :searchable-fields       #{:id
                                :source
                                :entity.id
                                :event_type}})))
```

src/ctia/entity/feedback.clj
```diff
@@ -82,23 +82,33 @@
...
+     :searchable-fields        #{:id
+                                 :entity_id
+                                 :external_ids
+                                 :external_references.description
+                                 :external_references.external_id
+                                 :external_references.hashes
+                                 :external_references.source_name
+                                 :external_references.url
+                                 :feedback
+                                 :reason}})))
```
ereteog, line 111:
```suggestion
     :searchable-fields        #{:id
                                 :source
                                 :entity_id
                                 :feedback
                                 :reason}})))
```

src/ctia/entity/identity_assertion.clj
```diff
@@ -112,7 +112,18 @@
...
+                                :external_references.url
+                                :assertions.name
+                                :assertions.value
+                                :external_ids
+                                :external_references.description
+                                :external_references.external_id
+                                :external_references.hashes
+                                :external_references.source_name
+                                :identity.observables.type
+                                :identity.observables.value}}))
```
ereteog, line 126:
```suggestion
    :searchable-fields        #{:id
                                :source
                                :assertions.name
                                :assertions.value
                                :identity.observables.type
                                :identity.observables.value}}))
```

src/ctia/entity/incident.clj
```diff
@@ -231,7 +231,24 @@
...
+                                 :external_references.description
+                                 :external_references.external_id
+                                 :external_references.hashes
+                                 :external_references.source_name
+                                 :external_references.url
+                                 :intended_effect
+                                 :promotion_method
+                                 :short_description
+                                 :status
+                                 :title}})))
```
ereteog, line 251:
```suggestion
     :searchable-fields        #{:description
                                 :source
                                 :id
                                 :assignees
                                 :categories
                                 :confidence
                                 :discovery_method
                                 :intended_effect
                                 :promotion_method
                                 :short_description
                                 :status
                                 :title}})))
```

src/ctia/entity/indicator.clj
```diff
@@ -161,7 +161,29 @@
...
+                                :kill_chain_phases.kill_chain_name
+                                :kill_chain_phases.phase_name
+                                :likely_impact
+                                :producer
+                                :severity
+                                :short_description
+                                :specification
+                                :tags
+                                :test_mechanisms
+                                :title}}))
```
ereteog, line 186:
```suggestion
    :searchable-fields        #{:id
                                :source
                                :confidence
                                :description
                                :indicator_type
                                :kill_chain_phases.kill_chain_name
                                :kill_chain_phases.phase_name
                                :likely_impact
                                :producer
                                :severity
                                :short_description
                                :tags
                                :test_mechanisms
                                :title}}))
```

this is an example of specification https://github.com/threatgrid/ctim/blob/f6ffae8bbc986ba9eed13d7d4fdf1d6f1519f552/src/ctim/examples/indicators.cljc#L42
you cannot add it as a whole
```
   :specification {:type "Judgement"
                   :judgements ["http://example.com/ctia/judgement/judgement-1000efcc-3474-498d-a655-0e0dc57f84ba"]
                   :required_judgements [{:judgement_id "http://example.com/ctia/judgement/judgement-1000efcc-3474-498d-a655-0e0dc57f84ba"
                                          :source "source"
                                          :relationship "relationship"
                                          :confidence "High"}]}

File: src/ctia/entity/investigation.clj [hunk 14]
```diff
@@ -95,7 +95,24 @@
...
+                                :external_references.source_name
+                                :external_references.url
+                                :investigated_observables
+                                :object_ids
+                                :short_description
+                                :targets.observables.type
+                                :targets.observables.value
+                                :targets.os
+                                :targets.type
+                                :title}}))
```
ereteog, line 115:
```suggestion
    :searchable-fields        #{:id
                                :source
                                :description
                                :investigated_observables
                                :object_ids
                                :short_description
                                :targets.observables.type
                                :targets.observables.value
                                :targets.os
                                :targets.type
                                :title}}))
```

this is an example of actions: https://github.com/threatgrid/ctia/blob/43ff86881b099bfd5a86f6a24a7ec0e05e980c04/src/ctia/entity/investigation/examples.clj#L35
```
   "actions" "[{\"uuid\":\"cae31704-9454-4428-afac-711e41a7ae24\",\"state\":\"ok\",\"updated\":\"2019-01-02T17:48:32.099Z\",\"type\":\"collect\", ...}]"})
   ```

File: src/ctia/entity/judgement.clj [hunk 15]
```diff
@@ -108,27 +105,42 @@
...
+                                     :external_references.description
+                                     :external_references.external_id
+                                     :external_references.hashes
+                                     :external_references.source_name
+                                     :external_references.url
+                                     :observable.type
+                                     :observable.value
+                                     :reason
+                                     :reason_uri
+                                     :severity}}]
```
ereteog, line 143:
```suggestion
         :searchable-fields        #{:id
                                     :source
                                     :confidence
                                     :disposition_name
                                     :observable.type
                                     :observable.value
                                     :reason
                                     :severity}}]
```

src/ctia/entity/malware.clj
```diff
@@ -146,7 +146,22 @@
...
+                                :external_references.external_id
+                                :external_references.hashes
+                                :external_references.source_name
+                                :external_references.url
+                                :kill_chain_phases.kill_chain_name
+                                :kill_chain_phases.phase_name
+                                :labels
+                                :short_description
+                                :title
+                                :x_mitre_aliases}}))
```
ereteog, line 164:
```suggestion
    :searchable-fields        #{:id
                                :source
                                :abstraction_level
                                :description
                                :kill_chain_phases.kill_chain_name
                                :kill_chain_phases.phase_name
                                :labels
                                :short_description
                                :title
                                :x_mitre_aliases}}))
```

src/ctia/entity/relationship.clj
```diff
@@ -224,7 +224,20 @@
...
+                                :external_references.description
+                                :external_references.external_id
+                                :external_references.hashes
+                                :external_references.source_name
+                                :external_references.url
+                                :relationship_type
+                                :short_description
+                                :source_ref
+                                :target_ref
+                                :title}}))
```
ereteog, line 240:
```suggestion
    :searchable-fields        #{:id
                                :source
                                :description
                                :relationship_type
                                :short_description
                                :source_ref
                                :target_ref
                                :title}}))
```

src/ctia/entity/sighting.clj
```diff
@@ -65,7 +65,43 @@
...
+                                :sensor_coordinates.observables.value
+                                :sensor_coordinates.os
+                                :sensor_coordinates.type
+                                :severity
+                                :short_description
+                                :targets.observables.type
+                                :targets.observables.value
+                                :targets.os
+                                :targets.type
+                                :title}}))
```
ereteog, line 104:
```suggestion
    :searchable-fields        #{:id
                                :source
                                :confidence
                                :description
                                :observables.type
                                :observables.value
                                :relations.origin
                                :relations.origin_uri
                                :relations.related.type
                                :relations.related.value
                                :relations.relation
                                :relations.source.type
                                :relations.source.value
                                :resolution
                                :sensor
                                :sensor_coordinates.observables.type
                                :sensor_coordinates.observables.value
                                :sensor_coordinates.os
                                :sensor_coordinates.type
                                :severity
                                :short_description
                                :targets.observables.type
                                :targets.observables.value
                                :targets.os
                                :targets.type
                                :title}}))
```

src/ctia/entity/target_record.clj
```diff
@@ -125,7 +125,23 @@
...
+                                :external_references.source_name
+                                :external_references.url
+                                :short_description
+                                :targets.observables.type
+                                :targets.observables.value
+                                :targets.os
+                                :targets.sensor
+                                :targets.source_uri
+                                :targets.type
+                                :title}}))
```
ereteog, line 144:
```suggestion
    :searchable-fields        #{:id
                                :source
                                :description
                                :short_description
                                :targets.observables.type
                                :targets.observables.value
                                :targets.os
                                :targets.sensor
                                :targets.type
                                :title}}))
```

src/ctia/entity/tool.clj
```diff
@@ -91,7 +91,21 @@
...
+                                :external_references.external_id
+                                :external_references.hashes
+                                :external_references.source_name
+                                :external_references.url
+                                :kill_chain_phases.kill_chain_name
+                                :kill_chain_phases.phase_name
+                                :labels
+                                :short_description
+                                :title
+                                :x_mitre_aliases}}))
```
ereteog, line 108:
```suggestion
    :searchable-fields        #{:id
                                :source
                                :description
                                :kill_chain_phases.kill_chain_name
                                :kill_chain_phases.phase_name
                                :labels
                                :short_description
                                :title
                                :x_mitre_aliases}}))
```

src/ctia/entity/vulnerability.clj
```diff
@@ -227,7 +227,76 @@
...
+                                 :impact.cvss_v3.modified_user_interaction
+                                 :impact.cvss_v3.privileges_required
+                                 :impact.cvss_v3.remediation_level
+                                 :impact.cvss_v3.report_confidence
+                                 :impact.cvss_v3.scope
+                                 :impact.cvss_v3.temporal_severity
+                                 :impact.cvss_v3.user_interaction
+                                 :impact.cvss_v3.vector_string
+                                 :short_description
+                                 :title}})))
```
ereteog, line 299:
```suggestion
     :enumerable-fields        vulnerability-enumerable-fields
     :searchable-fields        #{:id
                                 :source
                                 :description
                                 :configurations.nodes.children.cpe_match.cpe23Uri
                                 :configurations.nodes.cpe_match.cpe23Uri
                                 :short_description
                                 :title}})))
```

src/ctia/entity/weakness.clj
```diff
@@ -152,7 +152,57 @@
...
+                                :detection_methods.effectiveness_notes
+                                :architectures.class
+                                :languages.class
+                                :potential_mitigations.description
+                                :common_consequences.note
+                                :detection_methods.effectiveness
+                                :common_consequences.impacts
+                                :operating_systems.class
+                                :modes_of_introduction.note
+                                :detection_methods.description}}))
```
ereteog, line 205:
```suggestion
    :searchable-fields        #{:id
                                :source
                                :paradigms.name
                                :description
                                :technologies.name
                                :background_details
                                :abstraction_level
                                :languages.name
                                :architectures.prevalence
                                :common_consequences.scopes
                                :alternate_terms.term
                                :detection_methods.method
                                :architectures.name
                                :operating_systems.version
                                :potential_mitigations.effectiveness_notes
                                :likelihood
                                :potential_mitigations.effectiveness
                                :paradigms.prevalence
                                :languages.prevalence
                                :modes_of_introduction.phase
                                :external_references.source_name
                                :external_ids
                                :short_description
                                :common_consequences.likelihood
                                :notes.note
                                :title
                                :operating_systems.name
                                :operating_systems.cpe_id
                                :functional_areas
                                :notes.type
                                :operating_systems.prevalence
                                :structure
                                :external_references.description
                                :technologies.prevalence
                                :external_references.hashes
                                :affected_resources
                                :potential_mitigations.phases
                                :alternate_terms.description
                                :external_references.external_id
                                :potential_mitigations.strategy
                                :detection_methods.effectiveness_notes
                                :architectures.class
                                :languages.class
                                :potential_mitigations.description
                                :common_consequences.note
                                :detection_methods.effectiveness
                                :common_consequences.impacts
                                :operating_systems.class
                                :modes_of_introduction.note
                                :detection_methods.description}}))
```

test/ctia/test_helpers/search.clj
```diff
@@ -152,7 +153,7 @@
       (is (empty? (search-ids app entity "wor"))))
 
     ;; test stop word filtering
-    (is (= matched-ids (search-ids app entity "the word"))
+    (is (= matched-ids (search-ids app entity "the word" {:search_fields ["description"]}))
```
ereteog, line 156: this test is supposed to work as before.

src/ctia/entity/attack_pattern.clj
```diff
@@ -112,7 +112,23 @@
...
+                                :external_references.hashes
+                                :external_references.source_name
+                                :external_references.url
+                                :kill_chain_phases.kill_chain_name
+                                :kill_chain_phases.phase_name
+                                :short_description
+                                :title
+                                :x_mitre_contributors
+                                :x_mitre_data_sources
+                                :x_mitre_platforms}}))
```
ereteog, line 131:
```suggestion
    :searchable-fields        #{:id
                                :source
                                :abstraction_level
                                :description
                                :kill_chain_phases.kill_chain_name
                                :kill_chain_phases.phase_name
                                :short_description
                                :title
                                :x_mitre_contributors
                                :x_mitre_data_sources
                                :x_mitre_platforms}}))
```

ereteog (commented):
I did the manual selection of fields (which was the main job of this task).
The QA section is now ok.
I will look at the tests later.

test/ctia/http/generative/fulltext_search_test.clj
```diff
@@ -106,110 +76,138 @@
...
-     nil    ;; external-key-prefixes
-     login
-     services)
+   [{:test-description "searching for non-existing fields, returns errors"
+     :query-params     {:query         "*"
+                        :search_fields ["donec_retium_posuere_tellus"
+                                        "proin_neque_massa"]}
+     :bulk-gen         (bulk-gen-for :incidents)
+     :check            (fn [_ _ _ res]
+                         (is (-> res :parsed-body (contains? :errors))))}
```
ereteog, line 166:
```suggestion
                         (is (= 400 (-> res :parsed-body :status))))}
```

test/ctia/http/routes/common_test.clj
```diff
@@ -34,6 +35,28 @@
               :lt to}
              (sut/coerce-date-range from to))))))
 
+(deftest full-text-search-schema-functions
+  (testing "ctia.http.routes.common/prep-es-fields-schema"
+    (let [enum->set (fn [enum-schema]
+                      (-> (->> enum-schema ffirst (apply hash-map) :vs)))]
```
ereteog, line 41:
```suggestion
                      (->> enum-schema ffirst (apply hash-map) :vs))]
```

test/ctia/http/generative/fulltext_search_test.clj
ereteog: indeed it's already tested above:
"https://github.com/threatgrid/ctia/pull/1127/files#diff-7c52458f6c78a79afff4de6bfb3123a41d720dad0bfb7eb0ef82ea67d0421a2bR102"

test/ctia/http/generative/fulltext_search_test.clj
```diff
@@ -106,110 +76,138 @@
...
 
+      {:test-description "searching in fields that contain the pattern, yields results"
+       :query-params     {:query         "\"first\" \"Ipsum\""
+                          :search_fields ["short_description" "title"]}
+       :bulk-gen         bulk
+       :check            (fn [_ _ _ unparsed]
+                           (let [res (-> unparsed :parsed-body)]
+                             (is (= 1 (count res)))
+                             (is "first incident" (-> res first :short_description))
+                             (is "Lorem Ipsum Test Incident" (-> res first :title))))})]))
```
ereteog, line 194: You tested previous PR here and there is at least one case which is already checked.
Here the task is the default behavior:
``` clojure
{:test-description 'searching in default search fields that contain the pattern, yields results"
       :query-params     {:query         "\"first\" \"Ipsum\""}
       :bulk-gen         bulk
       :check            (fn [_ _ _ unparsed]
                           (let [res (-> unparsed :parsed-body)]
                             (is (= 1 (count res)))
                             (is "first incident" (-> res first :short_description))
                             (is "Lorem Ipsum Test Incident" (-> res first :title))))}
                             
{:test-description 'searching in non default search fields that contain the pattern, shall not return any result"
       ;; green should be the value of :tlp, please do check it
       :query-params     {:query         "green"} 
       :bulk-gen         (assoc bulk
       :check            (fn [_ _ _ unparsed]
                           (let [res (-> unparsed :parsed-body)]
                             (is (empty? res))))}
```

src/ctia/http/routes/common.clj
```diff
@@ -50,6 +54,36 @@
...
 
+(s/defn prep-es-fields-schema :- (s/protocol s/Schema)
+  "Conjoins Elasticsearch fields parameter into search-q-params schema"
+  [{:keys [search-q-params
+           searchable-fields] :as entity-crud-config}]
+  (let [default-fields-schema (->>
+                               searchable-fields
+                               (map name)
+                               (apply s/enum))]
+    (if searchable-fields
```
ereteog, line 65:
```suggestion
    (if (seq searchable-fields)
```

test/ctia/http/routes/common_test.clj
```diff
@@ -34,6 +35,28 @@
               :lt to}
              (sut/coerce-date-range from to))))))
 
+(deftest full-text-search-schema-functions
```
ereteog, line 38: your `testing` description are the name of each function. 
Please, make a `deftest` per function.

test/ctia/http/routes/common_test.clj
```diff
@@ -34,6 +35,28 @@
...
+      (are [fields result]
+          (= result
+             (->
+              (sut/prep-es-fields-schema
+               {:search-q-params {},
+                :searchable-fields fields})
+              (st/get-in [:search_fields])
+              enum->set))
+        #{:foo :bar} #{"foo" "bar"}
+        #{:foo}      #{"foo"})))
```
ereteog, line 51: you need to test `nil` `#{}` empty values since your fn check this emptiness.

test/ctia/http/generative/fulltext_search_test.clj
ereteog: You also had issues with nested fields, you need to check this edge case with sightings for instance: `targets.observable.value`

ereteog (changes requested):
you mostly tested your previous PR (and indeed the QA section introduces some noise with some checks also related to previous PR)

test/ctia/test_helpers/search.clj
```diff
@@ -189,8 +190,7 @@
       (if (= "AND" default_operator)
         (is (= matched-ids found-ids-escaped)
             "escaping reserved characters should avoid parsing errors and preserve behavior of AND")
-        (is (set/subset? (set/union matched-ids unmatched-ids)
-                         found-ids-escaped)
+        (is (set/subset? found-ids-escaped matched-ids)
```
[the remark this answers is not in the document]
agzam: No, this test still flawed from the get-go. The arguments for `set/subset?` need to be flipped.

test/ctia/test_helpers/search.clj
agzam: I'm having to keep it for now. Some tests failed when I removed it.

test/ctia/http/routes/common_test.clj
```diff
@@ -34,6 +37,83 @@
...
+        (is (= result
+               (some-> (sut/prep-es-fields-schema
+                        {:search-q-params   {},
+                         :searchable-fields fields})
+                       (st/get-in [:search_fields])
+                       enum->set))
+            (format "when %s passed, %s expected" fields result))
+      #{:foo :bar} #{"foo" "bar"}
+      #{:foo}      #{"foo"}
+      nil          nil)))
```
ereteog, line 53: In order to properly test this function behavior which does:
``` clojure
    (if (seq searchable-fields)
      (do-stuff)
     search-q-params)))
```

```suggestion
      #{:foo}      #{"foo"})
      (testing "search-q-params shall not be modified with searchable-fields is nil or empty"
          (is (= incident/IncidentSearchParams
                   (sut/prep-es-fields-schema  {:search-q-params   incident/IncidentSearchParams})
                   (sut/prep-es-fields-schema  {:search-q-params   incident/IncidentSearchParams
                                                                    :searchable-fields #{}))))))
```

test/ctia/http/routes/common_test.clj
```diff
@@ -34,6 +37,83 @@
               :lt to}
              (sut/coerce-date-range from to))))))
 
+(deftest prep-es-fields-schema-test
+  (let [enum->set (fn [enum-schema]
+                    (->> enum-schema ffirst (apply hash-map) :vs))]
+    (are [fields result]
+        (is (= result
+               (some-> (sut/prep-es-fields-schema
+                        {:search-q-params   {},
```
ereteog, line 46: please, use a realtistic `:search-q-params ` value like:
https://github.com/threatgrid/ctia/blob/43ff86881b099bfd5a86f6a24a7ec0e05e980c04/src/ctia/entity/incident.clj#L183
```suggestion
                        {:search-q-params    incident/IncidentSearchParams,
```

test/ctia/test_helpers/search.clj
ereteog: If it still fail, there is a bug here since the searchable-fields shall be injected with `description`.

test/ctia/http/generative/fulltext_search_test.clj
```diff
@@ -106,110 +76,156 @@
...
 
+      {:test-description "searching in default fields that contain the pattern, yields results"
+       :query-params     {:query "\"first\" \"Ipsum\""}
+       :bulk-gen         bulk
+       :check            (fn [_ _ _ unparsed]
+                           (let [res (-> unparsed :parsed-body)]
+                             (is (= 1 (count res)))
+                             (is "first incident" (-> res first :short_description))
+                             (is "Lorem Ipsum Test Incident" (-> res first :title))))}
+      {:test-description "searching in non-default seach fields that contain the pattern, shall not return any results"
```
ereteog, line 195:
```suggestion
      {:test-description "searching in non-default search fields that contain the pattern, shall not return any results"
```

ereteog (commented):
you are close. I am concerned by the `description` search-fields you still need to inject. We need a clear explanation because `description` shall be injected with `searchable-fields`.

agzam:
> you are close. I am concerned by the `description` search-fields you still need to inject. We need a clear explanation because `description` shall be injected with `searchable-fields`.

So I did some digging. That test is failing for pretty much every entity that includes running search tests:

- actor
- asset
- attack-pattern
- campaign
- casebook
- coa
- incident
- investigation
- malware
- relationship
- sighting
- tool
- vulnerability
- weakness
- indicator
- investigation
- target-record

The tests pass for all entities that either don't run search tests at all or :search-field (passed into `ctia.test-helpers.crud/entity-crud-test`) is defined, those are:

-  asset-mapping
-  asset-properties
-  event
-  feedback
-  feed
-  identity-assertion
-  judgement

Basically, the problem is that the analyzer does something weird with "the" part when multiple fields are passed into the query, i.e.:

    "query_string": {
        "query": "the word",
        "default_operator": "AND",
        "fields": ["source", "description"]
    }

Doesn't work. But:

    "query_string": {
        "query": "word",
        "default_operator": "AND",
        "fields": ["source", "description"]
    }
    
Works.

It doesn't even matter if additional fields (in the above example it's "source") contain it or not. So I tried for example creating an entity setting up `:source` field to be `"the source"`, and guess what? 


    "query_string": {
        "query": "the word",
        "default_operator": "AND",
        "fields": ["source", "description"]
    }
    
Didn't work. It's almost funny, but when I changed the query to "the source", that didn't work either. Even removing the "description" field didn't work. Only specifiying it in quotes `"\\"the source\""` worked. So, the analyzer recognizes "the" as the stop word after all.

Another way to make it work is to flip the default operator:

    "query_string": {
        "query": "the word",
        "default_operator": "OR",
        "fields": ["source", "description", ... some other fields]
    }
    
---

The bottom line: in order for this test to work we either have to quote the query string, which defeats the purpose of the test; or we have to flip the default_operator (and that doesn't make sense here); or we have to specify the field explicitly. I guess we'd have to do the latter. Perhaps the only additional thing we could do is to add a comment for posterity.

test/ctia/http/generative/fulltext_search_test.clj
agzam: https://github.com/threatgrid/ctia/pull/1127#issuecomment-902391840

ereteog (approved):
The issue with the `description` field is actually a bug.
Here is the analysis: https://github.com/advthreat/iroh/issues/5590
You can merge this PR, since it's not a major search issue in our scenario, but we need to move on this new task.

PR #1128, opened 2021-06-16
Issue-4744: Include `query-params` in CPE search

halcyon:
> **Epic** threatgrid/iroh#4744

> Related #1077 

`query-params` were not being passed into the CPE Search. This PR also resolves paging problems that resulted from not including `query-params`.


ereteog (approved):
LGTM.
You should add a test to check the fix and avoid regressions.

PR #1131, opened 2021-06-29
remove clj-momo http helper

ereteog:
> Related #https://github.com/advthreat/iroh/issues/5385

I experienced similar issues related to large bodies in test submitted not only to ES but also to CTIA test server.
This PR internalizes the clj-momo test-helper and transforms the string body into an input stream to avoid http chuncks issue with big bodies while testing. 
In the path to get rid of `clj-momo` this PR also removes unnecessary use of `deep=` and custom schema validation.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.


agzam (approved):

frenchy64 (approved):
LGTM

PR #1137, opened 2021-07-20
Add recently joined team members as code owners

rplevy:


ereteog (changes requested):
Some parts are too sensitive, we can include sub directories when I am back.

agzam (approved):

ereteog (approved):
I am on PTO and cannot attend ongoing discussions.
Let's stick to my general rule: trust by default but don't mess with it 😉.
I would like to check ES related changes, mapping errors are VERY expensive to fix.

irinarenteria (approved):

wandersoncferreira (approved):

rplevy:
Resolution on CTIA codeowners is that the codeowners file is not actually enforced (we have never had the "require code owner approval" box checked) but @gbuisson says let's leave the file anyway as it pertains to notifying people.

PR #1144, opened 2021-08-30
adds enumerable fields to Incident

agzam:
> **Epic** https://github.com/advthreat/iroh/projects/14
> Close https://github.com/advthreat/iroh/issues/5634
> Related #

<a name="qa">[§](#qa)</a> QA
============================

1. Make sure there exist a few entities of one kind e.g., Incidents
2. Try posting metric queries, on topn and cardinality, with aggregation on `title` field, i.e.:

 `POST /ctia/incident/metric/cardinality?from=2021-01-01&aggregate-on=title`, or/and
 `POST /ctia/incident/metric/topn?from=2021-01-01&aggregate-on=title`
  
Expected: the resulting map should contain a positive number, e.g,:

    "data": {
      "title": 2     <--- this number shouldn't be zero
    },

```
```

25b2c094 * origin/incident_new_topn&cardinality_fiedls__#5634 pr feedback
b6288a79 * adds enumerable fields to Incident

### DIFF
--- a/src/ctia/stores/es/crud.clj
+++ b/src/ctia/stores/es/crud.clj
@@ -535,7 +535,7 @@ It returns the documents with full hits meta data including the real index in wh
   [{:keys [aggregate-on granularity timezone]
     :or {timezone "+00:00"}} :- HistogramQuery]
   {:date_histogram
-   {:field (get enumerable-fields-mapping aggregate-on aggregate-on)
+   {:field aggregate-on
     :interval granularity ;; TODO switch to calendar_interval with ES7
     :time_zone timezone}})
 
--- a/src/ctia/stores/es/crud.clj
+++ b/src/ctia/stores/es/crud.clj
@@ -368,11 +368,11 @@ It returns the documents with full hits meta data including the real index in wh
     :one-of {s/Any s/Any}
     :query s/Str}))
 
-(def sort-fields-mapping
+(def enumerable-fields-mapping
   "Mapping table for all fields which needs to be renamed
-   for the sorting. Instead of using fielddata we can have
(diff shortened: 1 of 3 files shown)


src/ctia/stores/es/crud.clj
```diff
@@ -535,21 +535,21 @@ It returns the documents with full hits meta data including the real index in wh
   [{:keys [aggregate-on granularity timezone]
     :or {timezone "+00:00"}} :- HistogramQuery]
   {:date_histogram
-   {:field aggregate-on
+   {:field (get enumerable-fields-mapping aggregate-on aggregate-on)
```
ereteog, line 538: It should only be applied to topn and cardinality.
Topn and cardinality performs aggregate on keywords field while histograms aggregate on time field which are always indexed as date.
```suggestion
   {:field aggregate-on)
```

test/ctia/stores/es/crud_test.clj
```diff
@@ -377,7 +377,17 @@
                                        :granularity :month
                                        :aggs {:agg-type :cardinality
                                               :agg-key :nb-orgs
-                                              :aggregate-on "groups"}}}))))
+                                              :aggregate-on "groups"}}})))
+  (is (= {:metric
+          {:date_histogram
+           {:field     "title.whole"
+            :interval  :week
+            :time_zone "+02:00"}}}
```
ereteog, line 385: you need an example with topn and another with cardinality but not histogram :-)

ereteog (approved):
Nice.
Please, add a topn example for the QA section.

ereteog:
@agzam the expected result for the topn is not the same. You can directly test it on INT and add a concrete example.

PR #1145, opened 2021-09-03
[POC DO NOT MERGE] high impact incident filter

ereteog:
> **Epic** https://github.com/threatgrid/response/issues/837
> Related https://github.com/advthreat/iroh/issues/5633
> Related https://github.com/advthreat/iroh/issues/5685

This PR implements the last definition of high impact incidents: incidents which come from secure endpoint.
It adds a `high_impact` parameter to the incident search params. When added it modifies the query to filter on the right source value.
Since the processing of this parameter is specific to the incidents, it's processing is done at the store level. Then we implement a dedicated store for incident.
I split incident in store and schema. The non refacto changes are:
- incident store, [ctia.entity.incident.es-store](https://github.com/threatgrid/ctia/pull/1145/files#diff-b90613956baca1c05ec12a9fd21ee3a0cc56dd437de90565864064d6cee8bc47)
- incident store tests, [ctia.entity.incident.es-store-test](https://github.com/threatgrid/ctia/pull/1145/files#diff-bac4023f7516a485c8d958f0549782db892d40d44dd301a38cf4cda5863cd107)
- incident additional search tests, [ctia.entity.incident-test](https://github.com/threatgrid/ctia/pull/1145/files#diff-13536891658e3f88756e547592bfab86d330fe3638956456d02439e9feba34bf)

<a name="qa">[§](#qa)</a> QA
============================

1. create incidents with the `source` value `secure endpoint` and incidents with other `source` values (e.g. `ngfw`, `stealthwatch`).
2. `GET /ctia/incident/search` without any parameter returns all incidents.
3. `GET /ctia/incident/search?high_impact=true` returns only incidents with `source` value EQUAL TO `secure endpoint`.
4. `GET /ctia/incident/search?high_impact=false` returns only incidents with `source` value DIFFERENT FROM `secure endpoint`.

```
intern: high impact incident filter in CTIA
```

### DIFF
--- a/src/ctia/entity/incident/es_store.clj
+++ b/src/ctia/entity/incident/es_store.clj
@@ -5,6 +5,7 @@
             [ctia.stores.es.mapping :as em]
             [ctia.stores.es.store :refer [close-connections!]]
             [ctia.stores.es.crud :as crud]
+            [clojure.string :as string]
             [schema.core :as s]
             [ctia.schemas.search-agg :refer [SearchQuery]]))
 
@@ -37,25 +38,35 @@
 (def handle-aggregate crud/handle-aggregate)
 (def handle-delete-search crud/handle-delete-search)
 
-(s/defn append-query-clause :- SearchQuery
+(s/defn append-query-clauses :- SearchQuery
   [search-query  :- SearchQuery
-   str-lucene-clause :- s/Str]
+   str-lucene-clauses :- [s/Str]]
   (let [query (get-in search-query [:full-text :query])
-        lucene-query (cond->> str-lucene-clause
-                       query (str query " "))]
+        prepared (string/join " OR " str-lucene-clauses)
+        lucene-query (if query
+                       ;; from previous observations the order of filters in lucene matters
+                       ;; prepared is likely more optimized, add it first
+                       (format "(%s) AND (%s)" prepared query)
+                       prepared)]
     (assoc-in search-query [:full-text :query] lucene-query)))
+
+(defn high-impact-sources
+  [conn-state]
+  (let [get-in-config (get-in conn-state [:services :ConfigService :get-in-config])]
+    (some-> (get-in-config [:ctia :incident :high-impact :source])
+            (string/split #","))))
 
 (s/defn prepare-impact :- SearchQuery
   [conn-state
    search-query :-  SearchQuery]
-  (let [get-in-config (get-in conn-state [:services :ConfigService :get-in-config])
-        high_impact_source (get-in-config [:ctia :incident :high-impact :source])
-        high_impact (get-in search-query [:filter-map :high_impact])]
-    (clojure.pprint/pprint get-in-config)
+  (let [sources (high-impact-sources conn-state)
+        high_impact (get-in search-query [:filter-map :high_impact])
+        query-clauses (map #(format "source:\"%s\"" %) sources)
+        neg-query-clauses (map #(str "-" %) query-clauses)]
     (cond-> search-query
       (some? high_impact) (update :filter-map dissoc :high_impact)
-      (false? high_impact) (append-query-clause (format "-source:\"%s\"" high_impact_source))
-      (true? high_impact) (assoc-in [:filter-map :source] high_impact_source))))
+      (false? high_impact) (append-query-clauses neg-query-clauses)
+      (true? high_impact) (append-query-clauses query-clauses))))
 
 (defrecord IncidentStore [state]
   IStore
--- a/src/ctia/entity/incident/es_store.clj
+++ b/src/ctia/entity/incident/es_store.clj
@@ -1,0 +1,95 @@
+(ns ctia.entity.incident.es-store
+  (:require [ctia.entity.incident.schemas
+             :refer [PartialStoredIncident StoredIncident]]
+            [ctia.store :refer [IQueryStringSearchableStore IStore]]
+            [ctia.stores.es.mapping :as em]
+            [ctia.stores.es.store :refer [close-connections!]]
+            [ctia.stores.es.crud :as crud]
+            [schema.core :as s]
+            [ctia.schemas.search-agg :refer [SearchQuery]]))
+
+(def incident-mapping
+  {"incident"
+   {:dynamic false
+    :properties
+    (merge
+     em/base-entity-mapping
+     em/describable-entity-mapping
+     em/sourcable-entity-mapping
+     em/stored-entity-mapping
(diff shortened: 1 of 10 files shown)


src/ctia/entity/incident/es_store.clj
```diff
@@ -0,0 +1,106 @@
...
+  (let [query (get-in search-query [:full-text :query])
+        prepared (string/join " OR " str-lucene-clauses)
+        lucene-query (if query
+                       ;; from previous observations the order of filters in lucene matters
+                       ;; prepared is likely more optimized, add it first
+                       (format "(%s) AND (%s)" prepared query)
+                       prepared)]
+    (assoc-in search-query [:full-text :query] lucene-query)))
+
+(defn high-impact-sources
```
yogsototh, line 53: While I understand what this is doing because I freshly read https://github.com/advthreat/iroh/issues/5685 I think this would be helpful to add a docstring here trying to give a short rationale (probably link to the issue in the docstring?).

Also probably a `s/defn` wouldn't hurt `(s/maybe [s/Str])` I guess.

src/ctia/entity/incident/es_store.clj
```diff
@@ -0,0 +1,106 @@
...
+  [conn-state]
+  (let [get-in-config (get-in conn-state [:services :ConfigService :get-in-config])]
+    (some-> (get-in-config [:ctia :incident :high-impact :source])
+            (string/split #","))))
+
+(s/defn prepare-impact :- SearchQuery
```
yogsototh, line 59: While I totally understand what this is doing and the why, I think we should be safer to add a new query parameter instead of adding a trick in the filter-map, because this is a non-trivial, surprise-specific behavior.

Why not consider:

adding a new query parameter in the search routes for incidents, a boolean called `high-impact`. This way you wouldn't pollute the behavior of `filter-map`.

From a client standpoint, this looks surprising to me.

The difficult part for me is the `(update :filter-map dissoc :high_impact)`.

src/ctia/entity/incident/es_store.clj
```diff
@@ -0,0 +1,106 @@
...
+  [conn-state
+   search-query :-  SearchQuery]
+  (let [sources (high-impact-sources conn-state)
+        high_impact (get-in search-query [:filter-map :high_impact])
+        query-clauses (map #(format "source:\"%s\"" %) sources)
+        neg-query-clauses (map #(str "-" %) query-clauses)]
+    (cond-> search-query
+      (some? high_impact) (update :filter-map dissoc :high_impact)
+      (false? high_impact) (append-query-clauses neg-query-clauses)
+      (true? high_impact) (append-query-clauses query-clauses))))
```
yogsototh, line 69: I would construct the `query-clauses` and `neg-query-clauses` only when needed with `let` for each `cond` case.

src/ctia/entity/incident/es_store.clj
```diff
@@ -0,0 +1,106 @@
...
+  (let [get-in-config (get-in conn-state [:services :ConfigService :get-in-config])]
+    (some-> (get-in-config [:ctia :incident :high-impact :source])
+            (string/split #","))))
+
+(s/defn prepare-impact :- SearchQuery
+  [conn-state
+   search-query :-  SearchQuery]
+  (let [sources (high-impact-sources conn-state)
+        high_impact (get-in search-query [:filter-map :high_impact])
+        query-clauses (map #(format "source:\"%s\"" %) sources)
```
yogsototh, line 64: Shouldn't there also be a reference to some `client-id` here?

test/ctia/entity/incident/es_store_test.clj
```diff
@@ -0,0 +1,47 @@
...
+           (sut/prepare-impact fake-conn-state
+                               {:filter-map {:title "ho no..."
+                                             :high_impact true}})))
+
+    (is (= {:filter-map {:title "ho no..."}
+            :full-text {:query "(source:\"source1\" OR source:\"source2\") AND (lucene)"}}
+           (sut/prepare-impact fake-conn-state
+                               {:filter-map {:title "ho no..."
+                                             :high_impact true}
+                                :full-text {:query "lucene"}})))
```
yogsototh, line 43: with my proposal this should look like:

```suggestion
                               {:high_impact true
                                :filter-map {:title "ho no..."}
                                :full-text {:query "lucene"}})))
```

test/ctia/entity/incident/es_store_test.clj
```diff
@@ -0,0 +1,47 @@
...
+                               {:filter-map {:title "ho no..."
+                                             :high_impact false}})))
+    (is (= {:filter-map {:title "ho no..."}
+            :full-text {:query "source:\"source1\" OR source:\"source2\""}}
+           (sut/prepare-impact fake-conn-state
+                               {:filter-map {:title "ho no..."
+                                             :high_impact true}})))
+
+    (is (= {:filter-map {:title "ho no..."}
+            :full-text {:query "(source:\"source1\" OR source:\"source2\") AND (lucene)"}}
```
yogsototh, line 39: I thing there should be a `client-id` in the `filter-map`.

yogsototh (commented):
I don't think we should change the expected simple behavior of `filter-map` if we can.
Even if we change the output format of incident to have a `high_impact` field this is not a CTIM standard new field if I understood correctly. As this sounds custom I would add a custom mechanism to handle that search.

Also, I think this PR does not handle the filter by client_id.

yogsototh (approved):
LGTM!

client-id will be handled in another PR and the client API will still be the expected one.
Thanks for the clarification @ereteog

src/ctia/entity/incident/es_store.clj
```diff
@@ -0,0 +1,106 @@
...
+(def handle-query-string-search (crud/handle-query-string-search PartialStoredIncident))
+(def handle-query-string-count crud/handle-query-string-count)
+(def handle-aggregate crud/handle-aggregate)
+(def handle-delete-search crud/handle-delete-search)
+
+(s/defn append-query-clauses :- SearchQuery
+  [search-query  :- SearchQuery
+   str-lucene-clauses :- [s/Str]]
+  (let [query (get-in search-query [:full-text :query])
+        prepared (string/join " OR " str-lucene-clauses)
```
ereteog, line 45: This is a bug: this should be an AND in the neg case

agzam (approved):

ereteog:
Here is the final design https://github.com/advthreat/iroh/issues/5710

PR #1146, opened 2021-09-08
Add client id to stored entities

ereteog:
> **Epic** https://github.com/advthreat/iroh/issues/5710
> Related https://github.com/advthreat/iroh/issues/5714

- add client-id to Identity protocol and accordingly update jwt auth.
- add client-id to StoredEntity model and update realize entity fns.

<a name="qa">[§](#qa)</a> QA
============================

1. Create 2 IROH clients
2. Create an incident with jwt created with each client
==> check that created incidents contain the field `client-id` with the value of corresponding client.
3. for earch client: `GET /ctia/incident/search?query=client_id:{client_id}`
==> check that each response only return the incident created with corresponding client id

```
intern: In CTIA, each newly created entity contains now the client-id. .
```

### DIFF
--- a/src/ctia/auth.clj
+++ b/src/ctia/auth.clj
@@ -3,6 +3,7 @@
 
 (defprotocol IIdentity
   (authenticated? [this])
+  (client-id [this])
   (login [this])
   (groups [this])
   (allowed-capabilities [this])
@@ -11,6 +12,8 @@
 
 (defprotocol IAuth
   (identity-for-token [this token]))
+
+(def not-logged-client-id nil)
 
 (def not-logged-in-owner "Unknown")
 
@@ -22,6 +25,8 @@
   IIdentity
   (authenticated? [_]
     false)
+  (client-id [_]
+    not-logged-client-id)
   (login [_]
     not-logged-in-owner)
   (groups [_]
@@ -34,9 +39,14 @@
 
 (def denied-identity-singleton (->DeniedIdentity))
 
-(s/defn ident->map :- (s/maybe {:login (s/maybe s/Str)
-                                :groups (s/maybe [s/Str])})
+(s/defschema IdentityMap
+  {:client-id (s/maybe s/Str)
+   :login (s/maybe s/Str)
+   :groups (s/maybe [s/Str])})
+
+(s/defn ident->map :- (s/maybe IdentityMap)
   [ident]
   (when ident
     {:login (login ident)
-     :groups (groups ident)}))
+     :groups (groups ident)
+     :client-id (client-id ident)}))
--- a/src/ctia/domain/entities.clj
+++ b/src/ctia/domain/entities.clj
@@ -1,14 +1,15 @@
 (ns ctia.domain.entities
   (:require [clj-momo.lib.time :as time]
+            [ctia.auth :as auth]
             [ctia.domain.access-control :refer [properties-default-tlp]]
-            [ctia.properties :refer [get-http-show]]
-            [ctia.schemas.core :as ctia-schemas
-             :refer [GraphQLRuntimeContext
-                     HTTPShowServices
-                     RealizeFnResult
-                     TempIDs]]
-            [ctim.domain.id :as id]
             [ctia.graphql.delayed :as delayed]
+            [ctia.properties :refer [get-http-show]]
+            [ctia.schemas.core
+             :as
+             ctia-schemas
+             :refer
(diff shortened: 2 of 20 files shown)


test/ctia/entity/web_test.clj
```diff
@@ -158,73 +156,77 @@
...
-                   :body new-judgement-1
-                   :headers {"Authorization" "45c1f5e3f05d0"
-                             "origin" "http://external.cisco.com"})
-             judgement-id (id/long-id->id (:id judgement))]
-
-         (is (= 201 status))
-
-         (testing "GET /ctia/judgement/:id with bad JWT Authorization header"
-           (let [response
+     (testing "POST /ctia/judgement with JWT Authorization header"
```
ereteog, line 159: All these changes are maintly a refactoring.
You can directly jump to: `(testing "Search must properly filter on client-id" ...)`

src/ctia/properties.clj
```diff
@@ -198,16 +198,9 @@
                       "ctia.migration.batch-size" s/Int
                       "ctia.migration.buffer-size" s/Int
 
-                      "ctia.store.asset" s/Str
```
ereteog, line 201: it's not related to this PR, but I discovered it. there is a function in the same ns which automatically generates it.

src/ctia/auth.clj
```diff
@@ -34,9 +39,14 @@
 
 (def denied-identity-singleton (->DeniedIdentity))
 
-(s/defn ident->map :- (s/maybe {:login (s/maybe s/Str)
-                                :groups (s/maybe [s/Str])})
+(s/defschema IdentityMap
+  {:client-id (s/maybe s/Str)
+   :login (s/maybe s/Str)
+   :groups (s/maybe [s/Str])})
+
```
agzam, line 46: Why are these nullable? In what cases we'd have the keys but not the values?

src/ctia/domain/entities.clj
```diff
@@ -34,44 +35,52 @@
                     (:end_time prev-valid-time)
                     time/default-expire-date)}})
 
+(defn default-realize
+  [type-name Model new-object id ident-map prev-object services]
```
agzam, line 39: too many positional args

src/ctia/auth.clj
ereteog: - `client-id` is only for the jwt authentication.
- for the others, it deserves more investigation in the code, I just preserved the output schema of `ident->map`.

test/ctia/test_helpers/collections.clj
```diff
@@ -0,0 +1,6 @@
+(ns ctia.test-helpers.collections
+  (:require [clojure.test :refer [is]]))
+
+(defn is-submap? [a b]
```
agzam, line 4: what's with Javanesque name? why not simply `submap?`

src/ctia/domain/entities.clj
```diff
@@ -34,44 +35,52 @@
...
+        with-base-fields (into new-object
+                               {:id id
+                                :type type-name
+                                :owner (or (:owner prev-object) login)
+                                :groups (or (:groups prev-object) groups)
+                                :schema_version schema-version
+                                :created (or (:created prev-object) now)
+                                :modified now
+                                :timestamp (or (:timestamp new-object)
+                                               (:timestamp prev-object)
```
agzam, line 52: :created and :modified I understand, what's :timestamp is for?

ereteog: That's legacy, I do not have the context. It acts like created.
It's in [CTIM](https://github.com/threatgrid/ctim/blob/master/doc/tutorials/modeling-threat-intel-ctim.md#1112-optional-base-fields): "the time this object was created at, or last modified."
`created` and `modified` are specific to CTIA to distringuish both cases.

test/ctia/test_helpers/collections.clj
ereteog: because it performs a `is` and a submap? :-)
I initially splitted it in 2 funcitons (`submap?` and `is-submap?`) with `(is (submap? a b))` but it breaks the diff rendering in cider. Since it's a test helper, I implemented only this version which actually helps to debug :-)

src/ctia/domain/entities.clj
ereteog: I wrap these parameters in a map, but the point was to simply wrap the body of the delayed fn to simplify its testing.

agzam (approved):

PR #1148, opened 2021-09-14
bulk patch routes

ereteog:
> **Epic** #https://github.com/advthreat/iroh/issues/2249

The PR implements bulk patch routes and most of the logic of bulk update.

<a name="qa">[§](#qa)</a> QA
============================

1. create 3 incidents and note their ids 
2. `PATCH /ctia/bulk`
``` javascript
{
  "incidents":
    [{"id": "incident-id-1", "source": "patched"},
     {"id": "incident-id-2", "source": "patched"},
     {"id": "incident-id-2", "source": "patched"}]
}
```
3. bulk export these patched incidents and check they are actually updated
4. check that update events are created for these incidents.

```
intern: bulk patch route in CTIA
```

### DIFF
--- a/src/ctia/bulk/core.clj
+++ b/src/ctia/bulk/core.clj
@@ -12,7 +12,8 @@
    [ctia.store :as store]
    [ring.util.http-response :refer [bad-request]]
    [schema-tools.core :as st]
-   [schema.core :as s]))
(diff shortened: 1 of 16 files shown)


msprunck (approved):
LGTM!

PR #1154, opened 2021-10-13
Fix flaky test

frenchy64:
Close https://github.com/advthreat/iroh/issues/5863

Would be cleaner to use a time service, but we don't have one.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: Fix flaky test
```

### DIFF
--- a/test/ctia/domain/entities_test.clj
+++ b/test/ctia/domain/entities_test.clj
@@ -36,7 +36,7 @@
       (is (= (:modified realized-create)
              (:created realized-create)
              (:timestamp realized-create))))
-    (Thread/sleep 50) ;; allow time to pass so :modified != :created below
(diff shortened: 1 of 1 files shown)


test/ctia/domain/entities_test.clj
```diff
@@ -36,7 +36,7 @@
       (is (= (:modified realized-create)
              (:created realized-create)
              (:timestamp realized-create))))
-
+    (Thread/sleep 50) ;; allow time to pass so :modified != :created below
```
agzam, line 39: Is that enough?

ereteog, line 39: 1 ms is enough :-)
```suggestion
    (Thread/sleep 1) ;; allow time to pass so :modified != :created below
```

ereteog (approved):
LGTM!

PR #1164, opened 2021-10-28
Enforce search-fields when none given

agzam:
Explicitly set :search_fields when a search query doesn't include any

> Close https://github.com/advthreat/iroh/issues/5953
> Related https://github.com/advthreat/iroh/issues/5938

<a name="qa">[§](#qa)</a> QA
============================
No QA is needed. Will be covered within https://github.com/advthreat/iroh/issues/5929

### DIFF
--- a/src/ctia/http/routes/common.clj
+++ b/src/ctia/http/routes/common.clj
@@ -73,7 +73,8 @@
   "Guarantees that ES fields parameter always passed to ES instance"
   [{:keys [search_fields] :as query-params}
    searchable-fields]
-  (cond-> query-params
+  query-params
+  #_(cond-> query-params
     (empty? search_fields) (assoc :search_fields (mapv name searchable-fields))))
 
 (def paging-param-keys
--- a/src/ctia/http/routes/common.clj
+++ b/src/ctia/http/routes/common.clj
@@ -73,9 +73,9 @@
   "Guarantees that ES fields parameter always passed to ES instance"
   [{:keys [search_fields] :as query-params}
    searchable-fields]
-  query-params
-  #_(cond-> query-params
-    (empty? search_fields) (assoc :search_fields (mapv name searchable-fields))))
+  (cond-> query-params
+    (empty? search_fields)
+    (assoc :search_fields (mapv name searchable-fields))))
 
 (def paging-param-keys
   "A list of the paging and sorting related parameters, we can use
--- a/src/ctia/http/routes/common.clj
+++ b/src/ctia/http/routes/common.clj
@@ -75,7 +75,7 @@
    searchable-fields]
   (cond-> query-params
     (empty? search_fields)
-    (assoc :search_fields searchable-fields)))
+    (assoc :search_fields (mapv name searchable-fields))))
 
 (def paging-param-keys
   "A list of the paging and sorting related parameters, we can use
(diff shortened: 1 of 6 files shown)


src/ctia/http/routes/crud.clj
```diff
@@ -331,14 +333,18 @@
...
-                   (search-query date-field params)
-                   identity-map
-                   (select-keys params routes.common/search-options))
-                 (ent/page-with-long-id services)
-                 ent/un-store-page
-                 routes.common/paginated-ok))
+             (let [params* (if (= "true" (flag-value :enforce-search-fields))
+                             (routes.common/enforce-search-fields
+                              params searchable-fields)
+                             params)]
```
agzam, line 339: This part here is only temporarily. Semantically, enforcing fields should be happening at the lower level, somewhere within `ctia.stores.es.crud/make-search-query`, making it happen down at that level requires changing the signature of every instance of `IQueryStringSearchableStore/query-string-search`, so parts of `entity-crud-config` could hike down there as well (in this case `searchable-fields` and `FeaturesService` instance). 

This may initially create more mess than just sticking it in here. But eventually that's what needs to be done.

test/ctia/http/generative/fulltext_search_test.clj
```diff
@@ -360,3 +362,40 @@
...
+              (whoami-helpers/set-whoami-response
+               app "45c1f5e3f05d0" "foouser" "foogroup" "user")
+              (let [results (atom nil)]
+                (with-redefs [crud/make-search-query
+                              (fn [_ q _]
+                                (reset! results
+                                        (->> q :full-text first :fields
+                                             (map keyword)
+                                             set))
+                                (throw (ex-info "Early termination" {})))]
```
agzam, line 383: This should be done through extending the protocol with a mocked fn, but in my attempts to do it the "right way" were not very successful, and since I'm planning to change the way how search-fields are enforced anyway, this entire test block will have to change sometime soon.

src/ctia/http/routes/common.clj
```diff
@@ -69,6 +69,14 @@
         (describe [default-fields-schema] "'fields' key of Elasticsearch Fulltext Query.")})
       search-q-params)))
 
+(s/defn enforce-search-fields
+  "Guarantees that ES fields parameter always passed to ES instance"
+  [{:keys [search_fields] :as query-params}
+   searchable-fields]
+  (cond-> query-params
+    (empty? search_fields)
+    (assoc :search_fields (mapv name searchable-fields))))
```
ereteog, line 78: Test it, but keywords should be ok
```suggestion
    (assoc :search_fields searchable-fields)))
```

test/ctia/http/generative/fulltext_search_test.clj
```diff
@@ -360,3 +362,40 @@
              (-> res first
                  (select-keys (keys expected))
                  (= expected))))))))))
+
+(deftest enforcing-fields-with-feature-flag-test
+  (testing "feature flag set? fields should be enforced"
+    (are [properties query expected-fields]
+         (helpers/with-properties properties
+           (helpers/fixture-ctia-with-app
```
ereteog, line 370: this starts CTIA each time. Any alternative?
The feature flag service is tested, then I would unit test `enforce-search-fields` (even if its code is trivial), and keep only one test case here to check that default search fields are properly disabled: `["ctia.feature-flags" "enforce-search-fields:false"]`. The 2 other test cases should be covered with the all the other search tests.

ereteog (approved):
lgtm

PR #1165, opened 2021-10-28
Extend FeaturesService

agzam:
Need to extend FeaturesService to handle feature flags.

`ctia.features-service/FeaturesService` in addition to disabling entities (`[:ctia :features]` key) now can handle arbitrary feature flags.

e.g., `ctia.feature-flags=easter-egg:on, experimental-ratio:42`

Ideally, it would be nice if `[:ctia :features]` could be modified to also handle feature flags, but since the key already used on various environments for disabling entities, changing the logic that handles it would be risky. 
Therefore I had to add a new key.
Also, to minimize confusion I renamed `enabled?` method to `entity-enabled?`.

> Close #
> Related https://github.com/advthreat/iroh/issues/5953

<a name="qa">[§](#qa)</a> QA
============================

no QA required

0aa8f9f7 disambiguates `FeaturesService::enabled?`, renames to `entity-enabled?`
34e26d5b try increasing timeout time for docker
a95863d2 extends FeaturesService to handle feature flags
655e4041 adds a simple test case

### DIFF
--- a/dev/ctia/dev/split_tests.clj
+++ b/dev/ctia/dev/split_tests.clj
@@ -204,7 +204,7 @@
                                              "\nOut:\n" out
                                              "\nErr:\n" err)
                                         {:res res}))))
-          timeout-seconds 120
+          timeout-seconds 240
           exit-if-too-long (str "if [[ " timeout-seconds " -le $SECONDS ]]; then "
                                 "  echo \"timeout during connection attempt (waited ${SECONDS} seconds)\"; "
                                 "  exit 1; "
--- a/src/ctia/features_service.clj
+++ b/src/ctia/features_service.clj
@@ -57,16 +57,3 @@
        (contains? x key)
        (not x)))))
 
-(comment
-
-  (require '[ctia.test-helpers.core :as th])
-
-  (th/with-properties
-    ["ctia.features.disable" "asset,actor,sighting"
-     "ctia.feature-flags" "experimental_ratio:35, bubble-gum:50"]
-    (th/fixture-ctia-with-app
-     (fn [app]
-       (let [{:keys [feature-flags]} (th/get-service-map app :FeaturesService)]
-         (println (feature-flags))))))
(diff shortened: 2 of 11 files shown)


dev/ctia/dev/split_tests.clj
```diff
@@ -204,7 +204,7 @@
                                              "\nOut:\n" out
                                              "\nErr:\n" err)
                                         {:res res}))))
-          timeout-seconds 120
+          timeout-seconds 240
```
ereteog, line 207: I saw you had some issues with the CI, do we still need to increase this timeout?

src/ctia/features_service.clj
```diff
@@ -34,3 +56,17 @@
...
+
+  (th/with-properties
+    ["ctia.features.disable" "asset,actor,sighting"
+     "ctia.feature-flags" "experimental_ratio:35, bubble-gum:50"]
+    (th/fixture-ctia-with-app
+     (fn [app]
+       (let [{:keys [feature-flags]} (th/get-service-map app :FeaturesService)]
+         (println (feature-flags))))))
+
+  )
```
ereteog, line 72:
```suggestion
```

ereteog (approved):
lgtm

dev/ctia/dev/split_tests.clj
agzam: Yes, I believe for the time being this is unfortunately the best we could do. Unless @frenchy64 thinks differently, I'd like to increase the timeout.

src/ctia/features_service.clj
agzam: Of course! I just kept it there for the reviewers.

dev/ctia/dev/split_tests.clj
frenchy64: Yes, we must increase the limit. master builds are failing.

PR #1155, opened 2021-10-15
adds default searchable fields

agzam:
Adds searchable-fields in all searchable entities. Users allowed to use subset of those fields with search queries.

> **Epic** https://github.com/advthreat/iroh/issues/5929
> Close https://github.com/advthreat/iroh/issues/5938
> Related #

<a name="qa">[§](#qa)</a> QA
============================

*QA plan for this ticket should be very similar to one in: https://github.com/threatgrid/ctia/pull/1109 Mostly, we need to make sure there's no regression and the search works as before.*

*We encourage you to think of various query strategies and send as many different queries as possible. Saturate the test cases to flush out things that we may have missed. The following outline is an example of a few different queries.*

Naturally, you're going to need some data to query against it. You can use Bundle Import to create a few entities. 

- Mark your entities with some unique group name, e.g., the following examples set the value of `source` field for every entity to be `pr1155`

- Whenever you need to re-run the tests, it's probably desirable to remove any existing entities, to avoid confusion:

```http
DELETE http://ctia-instance/ctia/incident/search?source=pr1155&query=*&REALLY_DELETE_ALL_THESE_ENTITIES=true

DELETE http://ctia-instance/ctia/actor/search?source=pr1155&query=*&REALLY_DELETE_ALL_THESE_ENTITIES=true
```

- Make sure you don't have any records for entities you're about to test. The following queries should return 0.

```http
GET http://ctia-instance/ctia/incident/search/count&source=pr1155

GET http://ctia-instance/ctia/actor/search/count&source=pr1155
```

### Import a bundle
    
<details>
    <summary>
    Use bundle import to create some entities, the following example bundle contains three incidents, and a single actor:
   </summary>

```
POST http://ctia-instance/ctia/bundle/import
Content-type: application/json

{
    "timestamp": "2016-02-11T00:40:48Z",
    "incidents": [
        {
            "confidence": "High",
            "timestamp": "2016-02-11T00:40:48Z",
            "tlp": "green",
            "language": "language",
            "status": "Open",
            "categories": [
                "Denial of Service",
                "Improper Usage"
            ],
            "intended_effect": "Extortion",
            "source_uri": "http://example.com/incident-one",
            "discovery_method": "Log Review",
            "external_references": [],
            "incident_time": {
                "reported": "2016-02-11T00:40:48Z",
                "remediated": "2016-02-11T00:40:48Z",
                "rejected": "2016-02-11T00:40:48Z",
                "opened": "2016-02-11T00:40:48Z",
[...]

PR #1167, opened 2021-11-17
adds rename-search-fields fn

agzam:
Automatic remapping `source` to `source.text` in queries

> Close https://github.com/advthreat/iroh/issues/5999
> Related https://github.com/advthreat/iroh/issues/5938

<a name="qa">[§](#qa)</a> QA
============================
No QA is needed. Will be covered within https://github.com/advthreat/iroh/issues/5929

### DIFF
--- a/src/ctia/entity/incident.clj
+++ b/src/ctia/entity/incident.clj
@@ -130,15 +130,15 @@
      em/describable-entity-mapping
      em/sourcable-entity-mapping
      em/stored-entity-mapping
-     {:confidence       em/searchable-token
+     {:confidence       em/token
       :status           em/token
       :incident_time    em/incident-time
-      :categories       em/searchable-token
-      :discovery_method em/searchable-token
-      :intended_effect  em/searchable-token
-      :assignees        em/searchable-token
-      :promotion_method em/searchable-token
-      :severity         em/searchable-token})}})
+      :categories       em/token
+      :discovery_method em/token
+      :intended_effect  em/token
+      :assignees        em/token
+      :promotion_method em/token
+      :severity         em/token})}})
 
 (def-es-store IncidentStore :incident StoredIncident PartialStoredIncident)
 
@@ -216,14 +216,7 @@
   #{:description
     :source
     :id
-    :assignees
-    :categories
-    :confidence
-    :discovery_method
-    :intended_effect
-    :promotion_method
     :short_description
-    :status
     :title})
 
 (s/defn incident-routes [services :- APIHandlerServices]
--- a/src/ctia/entity/incident.clj
+++ b/src/ctia/entity/incident.clj
@@ -246,8 +246,7 @@
      :search-capabilities      :search-incident
      :external-id-capabilities :read-incident
      :histogram-fields         incident-histogram-fields
-     :enumerable-fields        incident-enumerable-fields
-     :searchable-fields        searchable-fields})))
+     :enumerable-fields        incident-enumerable-fields})))
 
 (def IncidentType
   (let [{:keys [fields name description]}
@@ -300,4 +299,5 @@
    :can-patch?            true
    :patch-capabilities    :create-incident
    :fields                incident-fields
-   :sort-fields           incident-fields})
+   :sort-fields           incident-fields
+   :searchable-fields     searchable-fields})
--- a/src/ctia/entity/incident.clj
+++ b/src/ctia/entity/incident.clj
@@ -130,15 +130,15 @@
      em/describable-entity-mapping
      em/sourcable-entity-mapping
      em/stored-entity-mapping
-     {:confidence em/token
-      :status em/token
-      :incident_time em/incident-time
-      :categories em/token
-      :discovery_method em/token
-      :intended_effect em/token
-      :assignees em/token
-      :promotion_method em/token
-      :severity em/token})}})
+     {:confidence       em/searchable-token
+      :status           em/token
+      :incident_time    em/incident-time
+      :categories       em/searchable-token
+      :discovery_method em/searchable-token
+      :intended_effect  em/searchable-token
+      :assignees        em/searchable-token
+      :promotion_method em/searchable-token
+      :severity         em/searchable-token})}})
 
 (def-es-store IncidentStore :incident StoredIncident PartialStoredIncident)
 
--- a/src/ctia/stores/es/crud.clj
+++ b/src/ctia/stores/es/crud.clj
@@ -453,6 +453,18 @@ It returns the documents with full hits meta data including the real index in wh
                                            access-control-filter-list
                                            ident
                                            get-in-config))))))
+(s/defn rename-search-fields :- (s/maybe {s/Keyword [s/Str]})
+  "Automatically translates keyword fields to use underlying text field.
+
+   ES doesn't like when different types of tokens get used in the same query. To deal with
+   that, we create a nested field of type 'text', see:
+   `ctia.stores.es.mapping/searchable-token`. This should be opaque - caller shouldn't
+   have to explicitly instruct API to direct query to the nested field."
+  [fields :- (s/maybe [s/Any])]
+  (let [search-fields-mapping {"source" "source.text"}]
+    (when fields
+      {:fields
+       (mapv (comp #(get search-fields-mapping % %) name) fields)})))
 
 (s/defn refine-full-text-query-parts :- [{s/Keyword ESQFullTextQuery}]
   [full-text-terms :- [FullTextQuery]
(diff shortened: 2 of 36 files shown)


src/ctia/stores/es/crud.clj
```diff
@@ -453,6 +453,18 @@ It returns the documents with full hits meta data including the real index in wh
...
                                            get-in-config))))))
+(s/defn rename-search-fields :- (s/maybe {s/Keyword [s/Str]})
+  "Automatically translates keyword fields to use underlying text field.
+
+   ES doesn't like when different types of tokens get used in the same query. To deal with
+   that, we create a nested field of type 'text', see:
+   `ctia.stores.es.mapping/searchable-token`. This should be opaque - caller shouldn't
+   have to explicitly instruct API to direct query to the nested field."
+  [fields :- (s/maybe [s/Any])]
+  (let [search-fields-mapping {"source" "source.text"}]
```
ereteog, line 464: you need to include all fields which are in default search and mapped as searchable-token: https://github.com/threatgrid/ctia/pull/1155/files

ereteog (commented):
We need to change the mappings of token default fields to searchable-token and rename them all. There is at least `id` and `observable.value` in the [common mapping definition](https://github.com/threatgrid/ctia/blob/master/src/ctia/stores/es/mapping.clj).

src/ctia/stores/es/query.clj
```diff
@@ -53,3 +55,81 @@
...
+           res))
+       (if (map? val)
+         (let [path (if (not= :properties key)
+                      (conj prev (name key))
+                      prev)]
+           (searchable-fields-map-impl path val res))
+         res)))
+   result m))
+
+(s/defn searchable-fields-map :- {s/Str s/Str}
```
DeLaGuardo, line 87: Do we have a test for that function somewhere? It is a bit hard to review the function without some example of usage 😢

ereteog: In addition, it's sensitive enough to be unit tested and check the edge cases.

ereteog (commented):
Do not forget to define searchable-token mappings

> We need to change the mappings of token default fields to searchable-token and rename them all. There is at least id and observable.value in the common mapping definition.

We need unit tests on the fields rewrite to gain confidence on it (now I cannot tell if it will works with edge cases)

src/ctia/stores/es/query.clj
```diff
@@ -53,3 +55,81 @@
...
+
+(s/defschema ESConnStateProps
+  (st/optional-keys
+   {:config {s/Any s/Any}
+    :props {s/Keyword s/Any}
+    :index s/Any
+    :conn s/Any
+    :services s/Any}))
+
+(s/defn rename-search-fields :- (s/maybe {s/Keyword [s/Str]})
```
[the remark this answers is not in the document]
agzam: Roger that. Planning to add some test(s)

src/ctia/stores/es/query.clj
agzam: @DeLaGuardo @ereteog I've added tests: test/ctia/stores/es/query_test.clj

DeLaGuardo: It is a bit hard to follow the logic in searchable-fields-map-impl. So far I wrap it in my head as "Walk through data structure until value is a map without `:type "keyword"` collecting on the way all the keys except :properties as a path. Then if the leaf node has at least one field of type text - keep the pair, path and field. Later collection of pair will be turned into hash-map." 
Based on that I wrote alternative implementation:

```clj
(defn- searchable-fields-map-impl
  ([m] (searchable-fields-map-impl [] m))
  ([path m]
   (if (and (map? m) (not= (:type m) "keyword"))
     (let [step-fn (fn [k]
                     (let [path (if (= k :properties) path (conj path k))]
                       (searchable-fields-map-impl path (m k))))]
       (mapcat step-fn (keys m)))
     (let [fields (reduce-kv
                   (fn [acc field {:keys [type]}]
                     (if (= type "text")
                       (conj acc field)
                       acc))
                   []
                   (:fields m))]
       (map #(conj [path] %) fields)))))

(s/defn searchable-fields-map :- {s/Str s/Str}
  "Walks through entity's mapping properties and finds all fields with secondary searchable token.
Returns a map where key is path to a field, and value - path to the nested text token."
  [properties :- (s/maybe {s/Keyword s/Any})]
  (into {}
        (map (fn [[path field]]
               (let [path (str/join "." (map name path))]
                 [path
                  (str/join "." [path (name field)])])))
        (searchable-fields-map-impl properties)))
```

@agzam What do you think?

ereteog:
> Do not forget to define searchable-token mappings
> 
> > We need to change the mappings of token default fields to searchable-token and rename them all. There is at least id and observable.value in the common mapping definition.

src/ctia/stores/es/query.clj
agzam: Very nice @DeLaGuardo, your version looks a lot nicer. Thank you for spending time to figure it out. Switched to your implementation.

rplevy (approved):
LGTM!

agzam:
> We need to change the mappings of token default fields to searchable-token and rename them all. 

Wait, wait. Are you saying that we need to change `https://github.com/threatgrid/ctia/blob/rewrite-map_for_searchable-tokens/src/ctia/stores/es/mapping.clj#L58`, replacing:
 
```
(def base-entity-mapping
  {:id token
```
with:
```
(def base-entity-mapping
  {:id searchable-token
```
Is that right, @ereteog? Whoa, that's scary.

ereteog:
> Is that right, @ereteog? Whoa, that's scary.

Yes and other tokens fields that will be used as default search field. And yes, that's scary, [that's why iroh services does it :-) ](https://www.youtube.com/watch?v=wMEq1mGpP5A)

agzam:
@ereteog I've checked `searchable-fields` per entity and re-mapped every searchable field to use searchable-token [here](https://github.com/threatgrid/ctia/pull/1167/commits/1da4390f1a237e1c5452fc53e893bb5bbd8ff584)

src/ctia/entity/incident.clj
```diff
@@ -130,15 +130,15 @@
...
-     {:confidence em/token
-      :status em/token
-      :incident_time em/incident-time
-      :categories em/token
-      :discovery_method em/token
-      :intended_effect em/token
-      :assignees em/token
-      :promotion_method em/token
-      :severity em/token})}})
+     {:confidence       em/searchable-token
```
ereteog, line 133:
```suggestion
     {:confidence      searchable-token
```

ereteog (commented):
I detected 2 changes on searchable-fields that I detected with the mapping.
In order to gain confidence on this we need a full text search on nested fields. Sightings are a good case, add a full text search on a value of `targets.observables.value`of a sighting maximal.

src/ctia/stores/es/query.clj
```diff
@@ -53,3 +55,76 @@
...
+        term->es-query-part (fn [{:keys [query_mode fields] :as text-query}]
+                              (hash-map
+                               (or query_mode :query_string)
+                               (-> text-query
+                                   (dissoc :query_mode)
+                                   (merge
+                                    (when (and default_operator
+                                               (not= query_mode :multi_match))
+                                      {:default_operator default_operator})
+                                    (rename-search-fields es-conn-state fields)))))]
```
ereteog, line 129: We need to add the condition of the feature flag and turn it on after the refresh of the mappings.

agzam: Same flag or should I create new one 🤔 https://github.com/threatgrid/ctia/blob/master/src/ctia/http/routes/crud.clj#L336

agzam: New flag for now https://github.com/threatgrid/ctia/blob/rewrite-map_for_searchable-tokens/src/ctia/stores/es/query.clj#L131

rplevy (approved):
LGTM!

test/ctia/http/generative/fulltext_search_test.clj
```diff
@@ -140,13 +140,13 @@
                              (utils/update-items
                               incidents
                               ;; update only the first, leave the rest unchanged
-                              #(assoc % :discovery_method "Log Review"
+                              #(assoc % :short_description "Log Review"
```
ereteog, line 143: why did you change these tests?
I prefer the version with mixed tokens and text.

ereteog, line 143: please preserve original version, the changes you made should not impact this.

ereteog (approved):
LGTM

test/ctia/http/generative/fulltext_search_test.clj
agzam: > changes you made should not impact this.

They do though, remember, we've limited Incident searchable_fields, `discovery_method` and `assignees` are no longer in `ctia.entity.incident/searchable-fields`, I had to change it.

ereteog: then it would be nice to replace a token field by another token field

agzam: Eh? I'm not sure what you mean, would you elaborate?

PR #1171, opened 2021-11-19
Bump schema version (uses cljc instead of cljx)

frenchy64:
Bump schema version (uses cljc instead of cljx).

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

3a0e0fdb bump schema version (uses cljc instead of cljx)
44b9077f trying to disable a test to check CI status
517fb112 add service-overrides to fixture-ctia-with-app
71234f2b attempt to fix a flaky test

### DIFF
--- a/test/ctia/test_helpers/core.clj
+++ b/test/ctia/test_helpers/core.clj
@@ -279,10 +279,11 @@
   to make them explicit."
   ([t-with-app :- (s/=> s/Any
                         (s/named s/Any 'app))]
-   (fixture-ctia-with-app t-with-app true))
+   (fixture-ctia-with-app t-with-app true {}))
   ([t-with-app :- (s/=> s/Any
                         (s/named s/Any 'app))
-    enable-http?]
+    enable-http? :- s/Bool
+    services-override :- (s/maybe {s/Keyword s/Any})]
    ;; Start CTIA
    ;; This starts the server on an available port (if enabled)
    (let [http-port 0]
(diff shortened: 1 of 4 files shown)


agzam:
It looks like one of the tests I wrote earlier failing here. Allow me to re-write that test block and then we can go ahead with merging this PR. Thank you.

agzam:
@frenchy64, can you please take a look, I made some changes:

- I changed the signature of `ctia.test-helpers.core/fixture-ctia-with-app` so I could override services. At some point (probably right after merging this) I'd like to refactor it, so it takes the map of parameters as the first arg, and 'the-app' part is always at the end of it.

- I restructured the test, but I don't like much how it turned out; I'd be happy to do it differently if you could spare a couple of suggestions.

test/ctia/test_helpers/core.clj
```diff
@@ -279,10 +279,11 @@
   to make them explicit."
   ([t-with-app :- (s/=> s/Any
                         (s/named s/Any 'app))]
-   (fixture-ctia-with-app t-with-app true))
+   (fixture-ctia-with-app t-with-app true {}))
   ([t-with-app :- (s/=> s/Any
                         (s/named s/Any 'app))
-    enable-http?]
+    enable-http? :- s/Bool
+    services-override :- (s/maybe {s/Keyword s/Any})]
```
frenchy64, line 286: Nice, I think this is the first time the `{Kw Service}` representation comes in handy in CTIA!

I suggest:
1. keep the 2-argument arity (make backwards compatible).
2. (Later) Creating a new function that switches the arguments to your liking.
3. (Later) Gradually moving to that function.

agzam: > keep the 2-argument arity (make backwards compatible).
> (Later) Creating a new function/macro that switches the arguments to your liking.

Initially, I thought about doing just that, but then I looked through every invocation, and it 2-arity version used only in a couple of places (already fixed in this PR).

I'm not very fond of this function because it acts like a macro, albeit a weird one, where the body isn't in the tail position. I'll make a PR deprecating it with a new version to use instead.

agzam: Are you suggesting to limit the schema to contain only service keys, like here https://github.com/advthreat/iroh/blob/master/services/status/src/iroh/status/schemas.clj#L26

frenchy64: > Are you suggesting to limit the schema to contain only service keys?

No, it's fine. I just came up with the service map representation as an _input_ in the context of CTIA, but we've only ever used it in user code in our other project until now.

frenchy64:
@agzam go ahead and approve this.

agzam (approved):

PR #1172, opened 2021-11-25
Save few polar bears

ereteog:
Limit the search test to few entities. 
For more context, I made few tests locally, and crud test takes 2 more minutes with search tests which are very repetitive across entities.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.


DeLaGuardo (approved):
LGTM

PR #1173, opened 2021-11-29
signature change of fixture-ctia-with-app

agzam:
`ctia.test-helpers.core/fixture-ctia-with-app` feels like a weird macro where the body is not in the tail
position. Let's change that.

Per discussion [in this PR](https://github.com/threatgrid/ctia/pull/1171#discussion_r755405226)

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

### DIFF
--- a/test/ctia/test_helpers/core.clj
+++ b/test/ctia/test_helpers/core.clj
@@ -1,11 +1,10 @@
 (ns ctia.test-helpers.core
   (:require
-   [clojure.pprint :refer [pprint]]
    [clj-momo.lib.clj-time.coerce :as mcljtime-coerce]
    [clj-momo.lib.clj-time.core :as mcljtime]
    [clj-momo.lib.time :as time]
    [clj-momo.properties :refer [coerce-properties read-property-files]]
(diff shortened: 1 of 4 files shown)


rplevy (approved):
LGTM!

test/ctia/test_helpers/core.clj
```diff
@@ -274,16 +275,18 @@
       all-stores
       get-in-config)))
 
+(s/defschema WithAppOptions
+  (st/optional-keys
+   {:enable-http? s/Bool
```
frenchy64, line 280: How about changing it to `:disable-http?` so absent == false?

frenchy64 (approved):

test/ctia/test_helpers/core.clj
agzam: I've attempted that, but the behavior is controlled by "ctia.http.enabled" property; I think my delicate dance of negating 'disable-http?' to enable the property looked a bit weird, so I didn't do it.

PR #1177, opened 2021-12-03
Bundle import race condition

ereteog:
This PR adds a test which reproduces the race condition in the bundle/import to deduplicate data on external ids.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.


agzam (approved):
Interesting.

rplevy (approved):
LGTM!

PR #1181, opened 2021-12-08
Fix ES7 migration rollover and strengthen migration parameter checks

ereteog:
The templates of the target indices contain the "rolled over" alias, which should only be set on the index. ES5 has some tolerance over it but ES7 does not.
I also strenghten the parameter check to improve error messages on misconfiguration.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.


frenchy64 (approved):
LGTM!

PR #1182, opened 2021-12-09
Move enforce-search-fields fn inward

agzam:
This refactoring is part of [threatgrid/ctia#1167 "Adds rename-search-fields fn"](https://github.com/threatgrid/ctia/pull/1167). 

Separate PR to make it easier to review.

Even though it looks like many changes, it is simply moving the call to `enforce-search-fields` inward  - from `ctia.http.routes.crud` to `ctia.stores.es.query` - enforcing search fields should happen within the boundaries of `IQueryStringSearchableStore.query-string-search`. 
I couldn't do it before and had to keep it at the root level, until now.

To make it right, I had to:
- add `:searchable-fields` to every entity ~(I probably will remove the key from the routes later)~ (removed)
- ~and extend `FeaturesService` so it can simply return all entities. I couldn't directly use `ctia.entity.entities/all-entities` due to a circular dependency problem.~

update: Based on suggestion by @ereteog, changed that, see the comments below. 

<a name="qa">[§](#qa)</a> QA
============================

No QA required, since everything would be tested within the [epic](https://github.com/advthreat/iroh/issues/5929)

### DIFF
--- a/src/ctia/stores/es/query.clj
+++ b/src/ctia/stores/es/query.clj
@@ -132,8 +132,7 @@ Returns a map where key is path to a field, and value - path to the nested text 
 (s/defn refine-full-text-query-parts :- [{s/Keyword ESQFullTextQuery}]
   [es-conn-state :- ESConnStateProps
    full-text-terms :- [FullTextQuery]]
-  (let [{{:keys [default_operator]} :props
-         {{:keys [flag-value]} :FeaturesService} :services} es-conn-state
+  (let [{{:keys [default_operator]} :props} es-conn-state
         term->es-query-part (fn [{:keys [query_mode fields] :as text-query}]
                               (let [fields* (->> fields
                                                  (enforce-search-fields es-conn-state)
--- a/src/ctia/stores/es/query.clj
+++ b/src/ctia/stores/es/query.clj
@@ -102,12 +102,9 @@ Returns a map where key is path to a field, and value - path to the nested text 
 (s/defn enforce-search-fields :- (s/maybe [s/Str])
   [es-conn-state :- ESConnStateProps
    fields :- (s/maybe [s/Str])]
-  (let [{{{:keys [flag-value entities]} :FeaturesService} :services
-         {:keys [entity]} :props} es-conn-state
-        searchable-fields (some->> (entities)
-                                   entity
-                                   :searchable-fields
-                                   (mapv name))]
+  (let [{:keys [searchable-fields]
+         {{:keys [flag-value]} :FeaturesService} :services} es-conn-state
+        searchable-fields (mapv name searchable-fields)]
     (if (and (empty? fields)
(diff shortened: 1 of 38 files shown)


ereteog (commented):
It's great to inject it.
However, it seems like a "hack" with the feature service.
The proper way to inject it would be at the[ store creation](https://github.com/threatgrid/ctia/blob/1c84e16a639abc242329dce8a87e415c6b166706/src/ctia/stores/es/store.clj#L19) (check custom definitions) or even better in the [ESConnState](https://github.com/threatgrid/ctia/blob/1c84e16a639abc242329dce8a87e415c6b166706/src/ctia/stores/es/init.clj#L157).

agzam:
> ... or even better in the [ESConnState]

Hmm., I'm confused, isn't that what I'm doing here? https://github.com/threatgrid/ctia/blob/move-enforce-search-fields/src/ctia/store_service.clj#L21-L25
`ConfigService` was already there, I just added `FeaturesService` to that map.

Is there a better way?

ereteog:
> I just added `FeaturesService` to that map.

FeatureService is related to features. It's ok to inject it in the connection state but it's a little bit hacky to inject it to access entities.
However, at the creation of the connection state, you could directly inject the corresponding entity. When you init the entity n you could inject in the connection state the searchable-fields.

agzam:
@ereteog I did it as you suggested. I removed `(entities)` method from `FeaturesService` and added a key in `src/ctia/stores/es/init.clj`. Also, removed `:searchable-fields` from the routes - no need to keep the key in both places for every entity.

src/ctia/stores/es/query.clj
```diff
@@ -96,9 +96,22 @@ Returns a map where key is path to a field, and value - path to the nested text
     :props {s/Keyword s/Any}
     :index s/Any
     :conn s/Any
-    :services s/Any}))
+    :services s/Any
+    :searchable-fields (s/maybe #{s/Keyword})}))
 
-(s/defn rename-search-fields :- (s/maybe {s/Keyword [s/Str]})
+(s/defn enforce-search-fields :- (s/maybe [s/Str])
```
ereteog, line 102: I would unit test this function.

ereteog (approved):
LGTM, but I am a little bit surprise to not see any impact on `refine-full-text-query-parts-test`. Could you check and explain why?

src/ctia/stores/es/query.clj
agzam: https://github.com/threatgrid/ctia/blob/move-enforce-search-fields/test/ctia/http/generative/fulltext_search_test.clj#L410

agzam: >  I am a little bit surprise to not see any impact on refine-full-text-query-parts-test

Oh, you are right - I need to extend the test, taking into account that enforcing fields now happens inside this function. My bad. I'll add some test cases.

PR #1183, opened 2021-12-10
Disable test for Bundle import race condition

ereteog:
> Close #https://github.com/advthreat/iroh/issues/6098

Disable the race condition test which is more a probe of the race condition that can be used to address it later.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: Disable test for Bundle import race condition
```

### DIFF
--- a/test/ctia/bundle/routes_test.clj
+++ b/test/ctia/bundle/routes_test.clj
@@ -1126,40 +1126,41 @@
          (is (= 200 (:status bundle-get-res)))
          (is (seq (-> bundle-get-res :parsed-body :casebooks))))))))
 
-(deftest race-condition-test
-  (helpers/with-properties
-    ["ctia.store.es.default.refresh" "false"
-     "ctia.store.es.default.refresh_interval" "1s"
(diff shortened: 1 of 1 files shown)


test/ctia/bundle/routes_test.clj
```diff
@@ -1125,3 +1125,42 @@
          (is (= 201 (:status link-res)))
          (is (= 200 (:status bundle-get-res)))
          (is (seq (-> bundle-get-res :parsed-body :casebooks))))))))
+
+(comment
+  (deftest race-condition-test
```
msprunck, line 1130:
```suggestion
  (deftest ^:disabled race-condition-test
```

Usually we use the [`disabled`](https://github.com/threatgrid/ctia/blob/32857663cdd7ac385161103dbafa8dc4f98febf0/dev-resources/circleci_test/config.clj#L19) selector to disable a test

msprunck (approved):
LGTM!

frenchy64 (approved):

PR #1190, opened 2022-01-28
Logging refinements

msprunck:
- Fixes a warning message by adding the dependency [`log4j-over-slf4j`](https://www.slf4j.org/legacy.html#log4j-over-slf4j). It contains replacement classes that redirect all work to their corresponding SLF4J classes.
- Changes the logback encoder pattern to replace the the class name by the logger name in the log message

`2022-01-28T15:14:12.057Z ERROR (qtp1714253032-39) [c.tools.logging$eval451$fn__456] - Request validation failed:`

->

`2022-01-28T15:14:12.057Z ERROR (qtp1714253032-39)  [ctia.http.exceptions] - Request validation failed:`

Usually the logger name matches the namespace from which the log fn is called. 

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

<a name="release-notes">[§](#release-notes)</a> Release Notes
=============================================================

```
intern: Logging refinements
```

PR #1189, opened 2022-01-27
Add logback configuration and remove unused slf4j implementations

msprunck:
Since the switch to Trapperkeeper, [Logback](https://logback.qos.ch) is used as logging system in CTIA but there is no way to configure it.

Changes in this PR:
- Adds a configuration file for logback with the right date format (ISO8601)
- Removes the unused log4j.properties file
- Disable unused slf4j implementations (log4j12 and nop)

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

<a name="release-notes">[§](#release-notes)</a> Release Notes
=============================================================

```
intern: Add logback configuration and remove unused slf4j implementations
```

PR #1192, opened 2022-02-01
Reduce noise in logs when running tests

msprunck:
> Related https://github.com/threatgrid/ctia/pull/1189
> Related https://github.com/threatgrid/ctia/pull/1190#issuecomment-1025980043

Adds a new configuration file for logback that change the configuration during the tests. By default, it disables logs in the root logger and only allows logger that start with `ctia`.

During its initialization, logback tries to find a file logback-test.xml in the classpath. If no such file is found, it checks for the file logback.xml in the classpath.

This PR adds a `logback-test.xml` file which is only available in the test resource folder. It will not be deployed by `uberjar` or used with the command `lein run -m ctia.main`.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: Disable logs during tests
```


yogsototh (approved):
LGTM!

frenchy64:
@msprunck I think there's a problem with the classpath, the CI is still noisy.

I think we need to move `test/resources` into its own folder `test-resources` and then add `:resource-paths ["test-resources"]` to the `:test-impl` alias.

I think `test/resources` being under `test` is a mistake.

msprunck:
> @msprunck I think there's a problem with the classpath, the CI is still noisy.
The goal of this PR is to retrieve the same level of logs we had before I added the configuration file `logback.xml`. It provides a new configuration file `logback-test.xml` which is specific to the test profile.

I tried to disable logs during tests with the `OFF` level or reduce them with `ERROR` or `WARN` level but it produces: 
- race conditions in the migration tests.
- errors in a [test](https://github.com/threatgrid/ctia/blob/6a08126da103c7ec54aa4c967c374664e832cea8/test/ctia/logger_test.clj#L47) that depends on INFO log level.
 
> I think we need to move `test/resources` into its own folder `test-resources` and then add `:resource-paths ["test-resources"]` to the `:test-impl` alias.
Why `:test-impl` and not `test`?
 
> I think `test/resources` being under `test` is a mistake
You are right, I'm not confortable with this folder structure. I changed it.

project.clj
```diff
@@ -204,12 +203,11 @@
                        :main ctia.main
                        :uberjar-name "ctia.jar"
                        :uberjar-exclusions [#"ctia\.properties"]}
-             :test {:jvm-opts ["-Dlog.console.threshold=WARN"]
```
msprunck, line 207: This property is specific to log4j which is no longer used

ereteog (approved):
LGTM

frenchy64:
> Why :test-impl and not test?

Whoops, never mind. Might be worth leaving it in `:dev` too https://github.com/technomancy/leiningen/blob/master/doc/PROFILES.md#default-profiles

> I tried to disable logs during tests with the OFF level or reduce them with ERROR or WARN levels but it produces...

Ah ok. I think the INFO logs make the output about 5-10x longer than before the `logback.xml` file, that's why I thought there was a classpath issue. Thanks for clarifying!

msprunck:
> Whoops, never mind. Might be worth leaving it in `:dev` too https://github.com/technomancy/leiningen/blob/master/doc/PROFILES.md#default-profiles

I don't think so. If we leave it in the `:dev` profile, logback will use `logback-test.xml` when running CTIA locally. Usually `logback-test.xml` is dedicated for tests. You can change the log configuration with `logback.xml`when running CTIA locally with `logback.xml`

agzam (approved):
nice!

PR #1195, opened 2022-02-07
pmap init store

ereteog:
> **Epic** #
> Close #
> Related #

The initialization of stores are independent. Parallelizing it should speed up all test that init CTIA.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.


ereteog:
this increases the build time in actions because there is on core per split.

PR #1196, opened 2022-02-08
threat context

ereteog:
This PR refactor bundle routes to extract threat context helpers.
It's a bases to enable future helpers to paginate this theat context querying.

<a name="ops">[§](#ops)</a> QA

No QA


test/ctia/test_helpers/core.clj
```diff
@@ -65,6 +65,7 @@
          "ctia.metrics.riemann.enabled"               false
          "ctia.metrics.console.enabled"               false
          "ctia.metrics.jmx.enabled"                   false
+         "ctia.store.es.identity.refresh"             "true"
```
ereteog, line 68: It forces the identity index refresh. It fixes a race condition when we set a new identity with `set-capabilities` in test but the test can start before the index is refreshed and make calls fails with `:missing_capability` error.
I discovered it with a test that failed only in gh and not on my local environment.

src/ctia/observable/core.clj
```diff
@@ -0,0 +1,139 @@
...
+  [filters
+   edge-end :- (s/enum :source_ref :target_ref)
+   entity-type
+   identity-map
+   services]
+  (let [relationships (get-relationships filters identity-map services)]
+    (->> (map edge-end relationships)
+         (map #(id/long-id->id %))
+         (filter #(= (str entity-type) (:type %)))
+         (map #(id/long-id %))
```
agzam, line 79: Looks like you can drop some symbols here:

```suggestion
         (map id/long-id->id)
         (filter #(= (str entity-type) (:type %)))
         (map id/long-id))
```

src/ctia/observable/core.clj
```diff
@@ -0,0 +1,139 @@
...
+         set)))
+
+(s/defn get-target-ids
+  [entity-ids
+   relationship_type
+   entity-type
+   identity-map
+   services]
```
agzam, line 87: Too many positional arguments. Up to you, of course, I'd wrapped them into a map.

agzam (approved):

Issue #1135, opened 2021-07-16
Error docker elasticsearch image on Apple Silicon chip


wandersoncferreira:
Executing `docker-compose ... up` for dev environment produces an error on Apple Silicon chip with the Elasticsearch image:

```
ERROR: no matching manifest for linux/arm64/v8 in the manifest list entries
```

The Docker Apple Silicon states that [not all images are available for arm64](https://docs.docker.com/docker-for-mac/apple-silicon/#known-issues) .

A possible solution is to add `platform: linux/x86_64`[here](https://github.com/threatgrid/ctia/blob/e92263ff7a021335c6a4990602f6b12159aa566e/containers/dev/docker-compose.yml#L7).

ereteog:
FYI @turbodog99 [investigated it in IROH](https://github.com/advthreat/iroh/pull/4883).
In CTIA there is one more constraint, we need ES5 and ES7 until the migration (there are breaking changes).
If the solution you propose works here it's perfect.

wandersoncferreira:
Thanks for pointing that out @ereteog , unfortunately the solution proposed only enabled the docker compose to proceed and spin the images, however, the ES5 breaks when docker try to start it. Then, this should not be considered a solution to this issue.

Definitely a good fix is to create a new Dockerfile targeting M1 for ES5. (ES7 already has support for aarm64)

agzam:
@wandersoncferreira Since not all of us on M1 Macs (I'm not), I think you'd have to test it out and ask @turbodog99 to check as well

DeLaGuardo:
Please check this [PR](https://github.com/threatgrid/ctia/pull/1288). Some changes allow to build ES5 docker image locally and using it on M1 Macs.

TLDR;

Run once — `docker compose -f containers/dev/docker-compose.yml -f containers/dev/m1-docker-compose.yml build`
Then just use the command `docker compose -f containers/dev/docker-compose.yml up` to spin up the cluster.

gbuisson:
I think we can close this issue now, let me know if you still have issues, thanks @DeLaGuardo !

PR #1201, opened 2022-03-04
es7 utils

frenchy64:
Related: https://github.com/threatgrid/ctia/issues/1135

ES7 focused dev tooling.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: es7 utils
```

### DIFF
--- a/dev/user.clj
+++ b/dev/user.clj
@@ -1,15 +1,17 @@
 (ns user
   (:require
-   [clojure.tools.namespace.repl :refer [set-refresh-dirs]]
    [cheshire.core :as json]
    [clj-http.client :as http]
    [clj-momo.lib.time :as time]
+   [clojure.repl :refer :all]
+   [clojure.tools.namespace.repl :refer [set-refresh-dirs refresh]]
    [ctia.init :as init]
+   [ctia.properties :as p]
    [ctim.schemas.vocabularies :as vocab]
    [puppetlabs.trapperkeeper.app :as app]
    [schema.core :as s]))
 
-(set-refresh-dirs "src" "dev" "test")
+(set-refresh-dirs "src" "dev" "test" "checkouts")
 
 ;;;;;;;;;;;;;;;;;;;;;;;
 ;; Lifecycle management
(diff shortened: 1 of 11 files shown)


dev/user.clj
```diff
@@ -67,16 +69,26 @@
...
 
+(defn -es7-init-config []
+  (-> (p/build-init-config)
+      (assoc-in [:ctia :store :es :default :version] 7)
+      (assoc-in [:ctia :store :es :default :port] 9207)))
+
+(defn start7
+  "Start CTIA with ES7"
+  []
+  (start :config (-es7-init-config)))
```
DeLaGuardo, line 90: I think `start` and `start7` should have the same interface to work as a replacement for each other. WDYT?

```suggestion
(defn start7
  "Start CTIA with ES7"
  [& {:as opts}]
  (apply start (update opts :config merge (-es7-init-config))))
```

this ^ suggestion is just to illustrate the approach. Looks like `merge` is not enough

test/ctia/test_helpers/core.clj
```diff
@@ -538,7 +542,7 @@
   "Returns entity map for given plural form of the entity key"
   [entity-key]
   (into {}
-        (filter (fn [e] (= entity-key (:plural (val e)))))
+        (filter #(= entity-key (:plural (val %))))
```
DeLaGuardo, line 545: 👍

test/ctia/test_helpers/es.clj
```diff
@@ -88,18 +88,24 @@
   (doseq [entity (keys (all-stores))]
     (purge-index-and-template entity services)))
 
+(defn -es-port []
+  (if ((h/get-es-versions-to-test) 5) "9205" "9207"))
```
DeLaGuardo, line 92: How about a more explicit variant?

```suggestion
  (if (contains? (h/get-es-versions-to-test) 5) "9205" "9207"))
```

and does it need to be a function?

DeLaGuardo (approved):
That is so good to see that you like to introduce new scripts instead of extra options for one "run-it-all" type of script 👍 

I have just a few questions/suggestions but even without it LGTM

dev/user.clj
frenchy64: I made `-es7-init-config` a function because there's isn't a good answer here IMO.

test/ctia/test_helpers/es.clj
frenchy64: Seemed useful to be able to set the `ctia.test.es-versions` system property at runtime, so I made it a function.

frenchy64: I like the interface of a set, so I renamed then fn `set-of-es-versions-to-test`.

frenchy64:
@DeLaGuardo glad you like it. The `config` calculation is so complex in CTIA that I'm ok keeping `start7` zero-arity until we come up with a better idea.

PR #1217, opened 2022-03-21
Add events-service to http server context

DeLaGuardo:
> Related advthreat/iroh#4958

Introduce RiemannService tk service. Then riemann events can be used to collect information about some important metrics related to bundle export functionality. Those events are grouped together via a correlation id set for each bundle export call. There are three main interesting events:

* Number of ids to export
* Overall timing
* Timings for ES requests to fetch entity, relationships and relationships targets

Those metrics are needed as a temporary extension and should be removed after proper performance improvements for export-bundle function.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: calls to send-event in ctia.bundle.core namespace should be removed after advthreat/iroh#4958 is fixed
public: new RiemanService with single method send-event
```

### DIFF
--- a/src/ctia/bundle/core.clj
+++ b/src/ctia/bundle/core.clj
@@ -333,10 +333,16 @@
        (filter (comp seq second))
        (into {})))
 
+(def ^:dynamic correlation-id nil)
+
+(defn- get-epoch-second []
+  (.getEpochSecond (java.time.Instant/now)))
+
 (s/defn fetch-relationship-targets
   "given relationships, fetch all related objects"
   [relationships identity-map
-   services :- APIHandlerServices]
+   {{:keys [send-event]} :RiemannService
+    :as services} :- APIHandlerServices]
   (let [all-ids (->> relationships
                      (map (fn [{:keys [target_ref source_ref]}]
                             [target_ref source_ref]))
@@ -350,7 +356,13 @@
                           (map (fn [[k v]]
                                  {(bulk/bulk-key
                                    (keyword k)) v}) by-type))
+        start (System/currentTimeMillis)
         fetched (bulk/fetch-bulk by-bulk-key identity-map services)]
+    (send-event {:service "export-bundle"
+                 :correlation-id correlation-id
+                 :time (get-epoch-second)
+                 :event-type "fetch-relationships-targets"
+                 :metric (- (System/currentTimeMillis) start)})
     (clean-bundle fetched)))
 
 (defn relationships-filters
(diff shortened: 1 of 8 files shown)


src/ctia/lib/riemann.clj
```diff
@@ -177,3 +169,17 @@
...
+                 riemann/tcp-client
+                 (riemann/batch-client
+                  (or (:batch-size config) 10)))
+        service-prefix (or (:service-prefix config) "CTIA")]
+    (assoc context
+           :conn conn
+           :service-prefix service-prefix)))
+
+(defn stop [{:keys [^RiemannClient conn]}]
+  (.close conn))
```
msprunck, line 185: I would check if conn is not nil. I think that the stop function can be invoked after the TK `init` and before the `start`

src/ctia/bundle/core.clj
```diff
@@ -442,9 +470,23 @@
...
    ident
    params
-   services :- APIHandlerServices]
+   {{:keys [send-event]} :RiemannService
+    :as services} :- APIHandlerServices]
   (if (seq ids)
-    (->> (map #(export-entities % identity-map ident params services) ids)
-         (reduce #(deep-merge-with coll/add-colls %1 %2))
-         (into empty-bundle))
+    (binding [correlation-id (str (java.util.UUID/randomUUID))]
```
msprunck, line 476: Could you import `java.util.UUID`?

src/ctia/bundle/core.clj
```diff
@@ -333,10 +333,16 @@
        (filter (comp seq second))
        (into {})))
 
+(def ^:dynamic correlation-id nil)
+
+(defn- get-epoch-second []
+  (.getEpochSecond (java.time.Instant/now)))
```
msprunck, line 339: Could you import `java.time.Instant`?

src/ctia/bundle/core.clj
```diff
@@ -442,9 +470,23 @@
...
   (if (seq ids)
-    (->> (map #(export-entities % identity-map ident params services) ids)
-         (reduce #(deep-merge-with coll/add-colls %1 %2))
-         (into empty-bundle))
+    (binding [correlation-id (str (java.util.UUID/randomUUID))]
+      (send-event {:service "export-bundle"
+                   :correlation-id correlation-id
+                   :time (get-epoch-second)
+                   :event-type "start"
+                   :metric (count ids)})
```
msprunck, line 481: The `service` field should be different for each event type. The metric is directly related to the service. We cannot mix the number of ids and the processing time of the `export-bundle` fn in the metric field if the service is the same. Here, we could rename the service as `Export bundle`. In the [Riemann documentation](http://riemann.io/concepts.html), they use `API port 8000 reqs/sec` as an example.

```suggestion
      (send-event {:service "Export bundle start"
                   :correlation-id correlation-id
                   :time (get-epoch-second)
                   :event-type "export-bundle"
                   :metric (count ids)})
```

You can keep a field like `event-type` to group all events related to the export bundle.

Nothing is wrong but it's more related to keep the same approach for all our Riemann events

src/ctia/lib/riemann.clj
DeLaGuardo: good catch, thanks

src/ctia/bundle/core.clj
DeLaGuardo: You're right, will change it

msprunck (approved):
LGTM!

PR #1219, opened 2022-03-21
Make ES7 auth configurable

frenchy64:
Close https://github.com/advthreat/iroh/issues/6385

Support `api-key` and `basic-auth` (also `headers` for tests). Reviewers please hide whitespace changes.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: Make ES7 auth configurable
```


gbuisson:
@frenchy64 did you investigate the idea of creating another level of structure under `ctia.store.es.default.auth.params`?

for example expose: `ctia.store.es.default.auth.params.id` , `ctia.store.es.default.auth.params.api-key` and
recombine that as a map in the output (which I think properties.clj) does for you already.

Otherwise this might be the tipping point where we should consider moving to an edn/json config format.

resources/ctia-default.properties
```diff
@@ -60,6 +60,12 @@ ctia.store.es.default.port=9205
 #ctia.store.es.default.host=es-storage.int.iroh.site
 #ctia.store.es.default.port=443
 #ctia.store.es.default.protocol=https
+#ctia.store.es.default.auth.type=api-key
+#ctia.store.es.default.auth.params.id=ngkvLnYB4ZehGW1qU-Xz
+#ctia.store.es.default.auth.params.api-key=6HMnACPRSVWSMvZCf9VcGg
```
ereteog, line 65: for dev purpose I assume we will mainly use basic auth with the default password defined for the container.
```suggestion
```

ereteog (commented):
Tests are only duplicated for ES 5 and 7 at the store level.
We of course do not want to duplicate all tests :-).
Could you introduce a basic end to end test with ES7, like a bundle import + any other read route?

ereteog (approved):

PR #1223, opened 2022-03-25
New RiemannService and metrics for export bundle function

DeLaGuardo:
> Related advthreat/iroh#4958

Introduce RiemannService tk service. Then riemann events can be used to collect information about some important metrics related to bundle export functionality. Those events are grouped together via a correlation id set for each bundle export call. There are three main interesting events:

* Number of ids to export
* Overall timing
* Timings for ES requests to fetch entity, relationships and relationships targets

Those metrics are needed as a temporary extension and should be removed after proper performance improvements for export-bundle function.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: calls to send-event in ctia.bundle.core namespace should be removed after advthreat/iroh#4958 is fixed
public: new RiemanService with single method send-event
```

### DIFF
--- a/.dir-locals.el
+++ b/.dir-locals.el
@@ -1,0 +1,11 @@
+((clojure-mode . ((eval . (progn
+                            ;; Define custom indentation for functions inside metabase.
+                            ;; This list isn't complete; add more forms as we come across them.
+                            ;; Described here:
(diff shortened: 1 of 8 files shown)


msprunck (approved):
LGTM!

hisans:
Cherry-Picked on TEST in release 1.93

PR #1239, opened 2022-04-04
fix ES7 rename-search-field.

ereteog:
> Related #https://github.com/advthreat/iroh/issues/5953

The search on source as a text by default does not work on INT after ES7 migration.
After further investigation, the remapping of textual search for searchable token does not work on ES7 because the implementation did not consider the difference between ES5 and ES7 mappings.

<a name="qa">[§](#qa)</a> QA
============================

After ES7 migration on TEST, In the incident manager, search for `Endpoint`, all `Secure Endpoint` incidents should be returned.

### DIFF
--- a/src/ctia/features_service.clj
+++ b/src/ctia/features_service.clj
@@ -24,6 +24,7 @@
 (tk/defservice features-service
   FeaturesService
   [[:ConfigService get-in-config]]
+  ;;TODO build a proper init to prepare fields only once
   (feature-flags
    [this]
    (when-let [fgs-str (some-> [:ctia :feature-flags]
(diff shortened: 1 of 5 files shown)


src/ctia/features_service.clj
```diff
@@ -24,6 +24,7 @@
 (tk/defservice features-service
   FeaturesService
   [[:ConfigService get-in-config]]
+  ;;TODO build a proper init to prepare fields only once
```
ereteog, line 27: @agzam thoughts?

marioaquino (approved):
LGTM

src/ctia/features_service.clj
agzam: Just to avoid reading it from the config every time? What's the bottleneck here?

agzam (approved):

jagrutpa:
This is not available in TEST. This is INT hence marked as QA - Verify later.

PR #1242, opened 2022-04-07
update-index-state task

frenchy64:
Close https://github.com/advthreat/iroh/issues/6341

Creates a new task to remove the performance hit of multiple instances of CTIA refreshing/updating mappings simultaneously.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: update-index-state task
```

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -102,7 +102,7 @@ jobs:
             touch dev-resources/ctia_test_timings.edn
           fi
       - name: Upload current test timing
-        uses: actions/upload-artifact@v2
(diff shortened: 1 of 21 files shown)


ereteog (approved):
LGTM,
I like the principle to be able to activate tasks.

Question: how do you plan to deploy and configure it?

My understanding is that we ask to launch this task before any release, and we set to true the update (template, settings, refresh-mappings) that we want to happen, right?

frenchy64:
I'm going to configure the deployment to run this task to completion before fanning CTIA out to multiple regions.

PR #1268, opened 2022-05-04
fix rollover task

ereteog:
This PR fixes the rollover task that missed the feature service dependency.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

The rollover task has been failing for months.

### DIFF
--- a/test/ctia/task/rollover_test.clj
+++ b/test/ctia/task/rollover_test.clj
@@ -6,7 +6,8 @@
             [ctia.task.rollover :as sut]
             [ctia.test-helpers.core :as helpers]
             [ctia.test-helpers.es :as es-helpers]
-            [ctia.test-helpers.fixtures :as fixt]))
+            [ctia.test-helpers.fixtures :as fixt]
+            [puppetlabs.trapperkeeper.app :as app]))
 
 (deftest rollover-aliased-test
   (es-helpers/for-each-es-version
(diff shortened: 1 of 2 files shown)


msprunck (approved):
LGTM!

obarbeau (approved):
LGTM!

test/ctia/task/rollover_test.clj
```diff
@@ -105,3 +106,6 @@
             (is (every? nil? [ko-type-1 ko-type-2]))
             (is (every? (fn [k] (some? (:rolled_over k)))
                         [ok-type-1 ok-type-2 ok-type-3]))))))))
+
+(deftest mk-app!-test
+  (is (map? (app/service-graph (sut/mk-app!)))))
```
DeLaGuardo, line 111: There is a predicate [service-graph?](https://github.com/puppetlabs/trapperkeeper/blob/3e5e7e286287d75e7fdf7eb1dabb2fa534091329/src/puppetlabs/trapperkeeper/internal.clj#L39-L46) defined in trapperkeeper. I don't think it adds any difference to the test but it could make it a little bit more explicit, wdyt?

Also, is this test suppose to test that no exception is thrown during building an app?

DeLaGuardo (commented):
I have a small comment but besides that - LGTM

test/ctia/task/rollover_test.clj
ereteog: I was about to add it, but the namespace is `puppetlabs.trapperkeeper.internal`, I am not keen on using functions that are not meant for external use.
For your question, I quickly wrapped it to avoid regressions and check that we actually return a trapper keeper app with valid services.

PR #1271, opened 2022-05-05
severity and created sorting

ereteog:
> Related #https://github.com/advthreat/iroh/issues/6679

sort by severity and date.

<a name="qa">[§](#qa)</a> QA
============================

 
Search incidents with new sort_by filter: `severity:desc,timestamp:desc`. Check that results are ordered by descending severity and for same severity values by descending created date.


NeelSaini:
Verified in rel v1.97

PR #1279, opened 2022-05-14
reproduce and fix flaky test

ereteog:
> Close #https://github.com/advthreat/iroh/issues/6016

I could reproduce deterministically the flaky test by logging the bundle in test failure.
I could limit it to a minimal example and it fails deterministically.
I removed the generative part for now which is not necessary for this regression test.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.


frenchy64 (approved):
LGTM

PR #1285, opened 2022-06-28
Performance improvements for bundle-export function

DeLaGuardo:
> Related https://github.com/advthreat/iroh/issues/4958

Optimizations for bundle-export functionality:

- [x] Extend `IStore` protocol with `read-records` method to fetch multiple records using a single ES search request.
- [x] Add the ability to fetch relationships related to the bunch of records using two ES requests instead of doing an ES request per id.
- [x] Replace iteration over provided ids with bulk operations in `bundle-export` logic.

## `scroll-with-limit` implementation details

The result of each invocation of `list-records` can have two main possible cases:

1. all results belong to the same record. The next call to `list-records` must use `search_after` to 'fast forward' the cursor to the next record id.
2. results list begins with records belonging to the same id, but then the tail might have 1-N partitions. That means the last partition in the result list is incomplete, and we must use `offset` to fetch the next page.

### `search_after` optimization

The second element must contain a numeric value guaranteed not to be contained in any document. For the timestamp, the value is -1. Only in this case, using "search_after" will ensure that the cursor moves to the next identifier.

<a name="qa">[§](#qa)</a> QA
============================

1. Import that bundle [bundle-0.json.zip](https://github.com/threatgrid/ctia/files/8463390/bundle-0.json.zip). It contains two sightings. Note their ids after a successful import.
2. Replace in every file from that archive [bundles.zip](https://github.com/threatgrid/ctia/files/8463400/bundles.zip) all occurrences of substrings `[[0]]` and `[[1]]` with sighting's ids from the result of step 1.
3. Import all 5 bundles.
4. Call bundle export with ids from the result of step 1. It should retrieve corresponding sightings together with related indicators. The result should not contain more than 1000 indicators per sighting.

```
intern: preffer using `ctia.store/read-records` method instead of `ctia.store/read-record` in the loop to fetch multiple records by its ids.
```

### DIFF
--- a/src/ctia/bundle/core.clj
+++ b/src/ctia/bundle/core.clj
@@ -340,162 +340,213 @@
 (defn- get-epoch-second []
   (.getEpochSecond (Instant/now)))
 
-(s/defn fetch-relationship-targets
-  "given relationships, fetch all related objects"
-  [relationships identity-map
-   {{:keys [send-event]} :RiemannService
-    :as services} :- APIHandlerServices]
-  (let [all-ids (->> relationships
-                     (map (fn [{:keys [target_ref source_ref]}]
-                            [target_ref source_ref]))
-                     flatten
-                     set
-                     (filter #(local-entity? % services))
-                     set)
-        by-type (dissoc (group-by
-                         #(ent/long-id->entity-type %) all-ids) nil)
-        by-bulk-key (into {}
-                          (map (fn [[k v]]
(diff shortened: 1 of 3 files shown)


msprunck (approved):
LGTM!

jagrutpa:
Tested with release v1.102 in TEST env.
Tested based on QA instructions.

PR #1288, opened 2022-07-12
Developer environment for Apple M1

DeLaGuardo:
* m1-docker-compose.yml with only changes needed for M1
* update kafka and zookeeper images
* build elasticsearch 5.6.16 from Dockerfile (only M1)
* build riemann and riemann-dash from Dockerfile (only M1)
* add a one-shot task to reset the default password for ES5
* set basic-auth as a default ES authentication method

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.


containers/dev/elasticsearch5-password/wait-for-it.sh
```diff
@@ -0,0 +1,182 @@
+#!/usr/bin/env bash
```
msprunck, line 1: I would add a reference to https://github.com/vishnubob/wait-for-it

msprunck (approved):
LGTM!

PR #1291, opened 2022-07-25
Fix bundle export

DeLaGuardo:
> Related advthreat/iroh#4958 #1285 

Performance improvements PR contains a bug where identity is passed as a record instead of a map into a function to fetch related nodes. Unfortunatelly, implementation silently delists records as inaccessible, making response almost empty. This PR fixes that by normalising arglist for bundle-export function.

<a name="qa">[§](#qa)</a> QA
============================

QA instructions are in #1285


agzam (approved):
nice.

jagrutpa:
Tested as part of this ticket: https://github.com/threatgrid/ctia/pull/1285

PR #1294, opened 2022-08-12
Sort incidents by tactics

frenchy64:
**Epic** https://github.com/advthreat/iroh/issues/6969
Close https://github.com/advthreat/iroh/issues/6974
Depends on https://github.com/threatgrid/ctim/pull/371

Incidents has a new field `tactics` that is a list of mitre tactic identifiers. Each tactic has a (proprietary) score. We want to sort incidents on the highest scored tactic it contains.

We do the sort via a painless script similar to severity sorting, except it's more like a `(sort-by #(apply max ...) ...)`.

The other implicit requirement is that `tactics` sorting can be combined with other fields. Previously, for `severity` sorting, we literally hardcoded half-dozen comma separated `sort_by` values like `severity:asc`, `severity:desc`, `severity:asc,timestamp:desc`, `severity:asc,timestamp:asc`...

Instead, this PR enhances `sort_by` to allow arbitrary combinations of the `field1:asc,field2:desc` syntax, and removed the half-dozen hardcodings of incident query parameters that we did in a pinch.

It would have been nice to turn `sort_by` into an array but we're stuck with the comma-separated syntax due to those hardcodings and combining them doesn't make sense to me. eg., `["severity:asc,timestamp:asc" "title:desc"]` doesn't feel right.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: Sort incidents by tactics/techniques
```

### DIFF
--- a/containers/dev/m1-docker-compose.yml
+++ b/containers/dev/m1-docker-compose.yml
@@ -3,18 +3,6 @@ services:
 
   elasticsearch5:
     build: ./m1-images/elasticsearch5
-  #elasticsearch7:
-  #  image: elasticsearch:7.17.0
-  #  environment:
-  #    - cluster.name=elasticsearch7
-  #    - discovery.type=single-node
-  #    - xpack.security.enabled=true
-  #    - xpack.security.authc.api_key.enabled=true
-  #    - xpack.security.authc.anonymous.roles=superuser # allows no auth for test purpose
-  #    - ELASTIC_PASSWORD=ductile
-  #  ports:
-  #    - "9207:9200"
-  #    - "9307:9300"
 
   docker-riemann-dev:
     build:
--- a/containers/dev/m1-docker-compose.yml
+++ b/containers/dev/m1-docker-compose.yml
@@ -1,19 +1,20 @@
 version: '2'
 services:
 
-  # delete me -- temporary
-  elasticsearch7:
-    image: elasticsearch:7.17.0
-    environment:
-      - cluster.name=elasticsearch7
-      - discovery.type=single-node
-      - xpack.security.enabled=true
-      - xpack.security.authc.api_key.enabled=true
-      - xpack.security.authc.anonymous.roles=superuser # allows no auth for test purpose
-      - ELASTIC_PASSWORD=ductile
-    ports:
-      - "9207:9200"
-      - "9307:9300"
+  elasticsearch5:
+    build: ./m1-images/elasticsearch5
+  #elasticsearch7:
+  #  image: elasticsearch:7.17.0
+  #  environment:
+  #    - cluster.name=elasticsearch7
+  #    - discovery.type=single-node
+  #    - xpack.security.enabled=true
+  #    - xpack.security.authc.api_key.enabled=true
+  #    - xpack.security.authc.anonymous.roles=superuser # allows no auth for test purpose
+  #    - ELASTIC_PASSWORD=ductile
+  #  ports:
+  #    - "9207:9200"
+  #    - "9307:9300"
 
   docker-riemann-dev:
     build:
--- a/containers/dev/m1-docker-compose.yml
+++ b/containers/dev/m1-docker-compose.yml
@@ -1,8 +1,19 @@
 version: '2'
 services:
 
-  elasticsearch5:
-    build: ./m1-images/elasticsearch5
+  # delete me -- temporary
+  elasticsearch7:
+    image: elasticsearch:7.17.0
+    environment:
+      - cluster.name=elasticsearch7
+      - discovery.type=single-node
+      - xpack.security.enabled=true
+      - xpack.security.authc.api_key.enabled=true
+      - xpack.security.authc.anonymous.roles=superuser # allows no auth for test purpose
+      - ELASTIC_PASSWORD=ductile
+    ports:
+      - "9207:9200"
+      - "9307:9300"
 
   docker-riemann-dev:
     build:
--- a/dependabot/dependency-tree.txt
+++ b/dependabot/dependency-tree.txt
@@ -49,7 +49,7 @@ ctia:ctia:jar:1.1.1-SNAPSHOT
 +- metosin:schema-tools:jar:0.12.2:compile
 +- threatgrid:flanders:jar:0.1.23:compile
 |  \- org.clojure:core.match:jar:0.3.0:compile
-+- threatgrid:ctim:jar:eb3169dfe544ee22358eccdfe7506b3e1a905224:compile
++- threatgrid:ctim:jar:1.1.12:compile
 |  +- org.mozilla:rhino:jar:1.7.7.1:compile
 |  \- kovacnica:clojure.network.ip:jar:0.1.3:compile
 |     \- org.clojure:clojurescript:jar:1.7.122:compile
--- a/dependabot/dependency-tree.txt
+++ b/dependabot/dependency-tree.txt
@@ -49,7 +49,7 @@ ctia:ctia:jar:1.1.1-SNAPSHOT
 +- metosin:schema-tools:jar:0.12.2:compile
 +- threatgrid:flanders:jar:0.1.23:compile
 |  \- org.clojure:core.match:jar:0.3.0:compile
-+- threatgrid:ctim:jar:1.1.11:compile
++- threatgrid:ctim:jar:eb3169dfe544ee22358eccdfe7506b3e1a905224:compile
 |  +- org.mozilla:rhino:jar:1.7.7.1:compile
 |  \- kovacnica:clojure.network.ip:jar:0.1.3:compile
 |     \- org.clojure:clojurescript:jar:1.7.122:compile
@@ -267,7 +267,7 @@ ctia:ctia:jar:1.1.1-SNAPSHOT
 +- org.slf4j:log4j-over-slf4j:jar:1.7.20:compile
 +- puppetlabs:trapperkeeper:jar:test:3.2.0:test
 +- puppetlabs:kitchensink:jar:test:3.2.0:test
-+- org.clojure:test.check:jar:1.1.0:test
++- org.clojure:test.check:jar:1.1.1:test
 +- com.gfredericks:test.chuck:jar:0.2.13:test
 +- clj-http-fake:clj-http-fake:jar:1.0.3:test
 |  \- robert:hooke:jar:1.3.0:test
--- a/src/ctia/entity/incident.clj
+++ b/src/ctia/entity/incident.clj
@@ -139,7 +139,8 @@
       :intended_effect  em/token
       :assignees        em/token
       :promotion_method em/token
-      :severity         em/token})}})
+      :severity         em/token
+      :tactics          em/token})}})
 
 (def-es-store IncidentStore :incident StoredIncident PartialStoredIncident)
 
@@ -159,7 +160,8 @@
            :intended_effect
            :assignees
            :promotion_method
-           :severity]))
+           :severity
+           :tactics]))
 
 (s/def sort-extension-templates :- SortExtensionTemplates
   {;; override :severity field to sort semantically
@@ -217,7 +219,8 @@
    :source
    :status
    :title
-   :severity])
+   :severity
+   :tactics])
 
 (def incident-histogram-fields
   [:timestamp
--- a/src/ctia/entity/incident.clj
+++ b/src/ctia/entity/incident.clj
@@ -192,22 +192,12 @@
               "TA0010" 50
               "TA0040" 50}}})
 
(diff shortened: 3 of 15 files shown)


dependabot/dependency-tree.txt
```diff
@@ -49,7 +49,7 @@ ctia:ctia:jar:1.1.1-SNAPSHOT
 +- metosin:schema-tools:jar:0.12.2:compile
 +- threatgrid:flanders:jar:0.1.23:compile
 |  \- org.clojure:core.match:jar:0.3.0:compile
-+- threatgrid:ctim:jar:1.1.11:compile
++- threatgrid:ctim:jar:eb3169dfe544ee22358eccdfe7506b3e1a905224:compile
```
marioaquino, line 52: Non-release jar version?

frenchy64: That would be https://github.com/threatgrid/ctim/pull/371

src/ctia/http/routes/common.clj
```diff
@@ -57,6 +57,33 @@
...
+(s/defn prep-sort_by-param-schema :- (s/protocol s/Schema)
+  "Generalize sort_by enum schema from (enum fields...) to s/Str
+  so it can accept the `field1:asc,field2:desc` syntax to sort by
+  multiple fields. Add the original enums to the doc for the route,
+  that also explains how this syntax works."
+  [search-q-params]
+  (cond-> search-q-params
+    (st/get-in search-q-params [:sort_by])
+    (st/update :sort_by
+               (fn [sort_by-enum]
```
marioaquino, line 69: Would doing an `s/fn` and expressing the type in the param work as an alternative the `assert` below?

marioaquino (commented):
The `src/ctia/stores/es/sort.clj` namespace is the only part of this PR that I can't provide useful review for.  Not sure what to do about that.  If no one else w more ES experience can weight in, I can approve just without meaningful feedback on that part.

containers/dev/m1-docker-compose.yml
```diff
@@ -3,6 +3,18 @@ services:
...
+  #  environment:
+  #    - cluster.name=elasticsearch7
+  #    - discovery.type=single-node
+  #    - xpack.security.enabled=true
+  #    - xpack.security.authc.api_key.enabled=true
+  #    - xpack.security.authc.anonymous.roles=superuser # allows no auth for test purpose
+  #    - ELASTIC_PASSWORD=ductile
+  #  ports:
+  #    - "9207:9200"
+  #    - "9307:9300"
```
j0ni, line 17: If this needs to stay here, can you comment why in there? Or remove it altogether if it is actually dead.

src/ctia/entity/incident.clj
```diff
@@ -159,7 +160,50 @@
...
+                          (group-by :risk_score)
+                          (sort-by key)
+                          (map second)))
+          out (into (sorted-map-by (fn [id1 id2]
+                                     (< (tactic->pos id1 0) (tactic->pos id2 0))))
+                    (map (fn [score group]
+                           (zipmap (map :id group) (repeat score)))
+                         (next (range)) groups))
+
+          _ (assert (= relevant-tactics (keys out))
```
j0ni, line 202: I know this is totally not production code, but it's the last binding in the block man, how can that not hurt your eyes?

j0ni:
In the other PR threatgrid/iroh#6983 I made a comment about deep nesting - lots of code here I'd make similar comments about. Do you all find it easier to work with those closures than the inevitably long-arity named alternatives?

src/ctia/http/routes/common.clj
j0ni: I like @marioaquino's idea too.

I wonder if there's a combinator we could have to make those `_` bindings go away in a more graceful way where a nice alternative isn't obvious. Something that takes the rhs of the binding and an assertion predicate.

```clojure
(let [valid-fields (chitty!! (:vs sort_by-enum) #(every? (some-fn string? simple-ident?) %))
      ...]
  ...)
```

But `s/fn` is grand if it covers the case.

j0ni: (It could be either-monad polymorphic, in case @yogsototh's work makes it further into the codebase.)

src/ctia/entity/incident.clj
frenchy64: I like making side effects stand out :) It hurts my brain less.

containers/dev/m1-docker-compose.yml
frenchy64: I don't understand why this was removed in the first place, but I'll figure that out another time.

src/ctia/http/routes/common.clj
frenchy64: My intention here was to communicate to any poor dev in the future that violates these invariants as directly as possible. An assert with a custom msg seems to work.

frenchy64:
> Do you all find it easier to work with those closures than the inevitably long-arity named alternatives?

This is probably just my style. I tend to use terse, procedural code below a certain abstraction threshold so I can hold the invariants in my head. I usually benefit from style feedback if it becomes unreadable.

src/ctia/http/routes/common.clj
frenchy64: @j0ni Maybe something like this, but I prefer the `_` version.

```clojure
(let [valid-fields (doto (:vs sort_by-enum)
                     (as-> valid-fields (assert (some-fn string? simple-ident?) valid-fields) (pr-str valid-fields)
...
```

containers/dev/m1-docker-compose.yml
DeLaGuardo: That was removed to keep compose files DRY. With docker compose you can merge multiple compose files together, stacking them into a chain: `docker compose -f docker-compose-1.yml -f docker-compose-2.yml`. In that case, `docker-compose-2.yml` file might contain only configuration options that need to be redefined and additional services not present in `docker-compose-1.yml` file.

The main idea for `m1-docker-compose-file.yml` is to help build images required for `aarch64` architecture (Mac M1/2):
```sh
# Execute this command only once
docker compose -f containers/dev/docker-compose.yml -f containers/dev/m1-docker-compose.yml build
```
After its done, you can start the development environment using the main docker compose file only and forget about `m1-docker-compose.yml`:
```sh
docker compose -f containers/dev/docker-compose.yml up -d
```

src/ctia/stores/es/sort.clj
```diff
@@ -3,6 +3,17 @@
...
+;; https://www.elastic.co/guide/en/elasticsearch/painless/8.1/painless-walkthrough.html#_missing_keys
+(defn- painless-return-default-if-no-field [field-name default-param]
+  (format "if (!doc.containsKey('%s') || doc['%s'].size() != 1) { return params.%s }"
+          field-name field-name default-param))
+
+(defn- normalize-remappings [remappings]
+  (into {} (map (fn [e]
+                  (update e 0 #(cond-> %
+                                 (string? %) str/lower-case))))
+        remappings))
```
DeLaGuardo, line 15: Looks like `clojure.core/update-keys` to me.

src/ctia/stores/es/sort.clj
```diff
@@ -18,6 +29,35 @@
...
+      ;;  :remappings {"TA0043" 1
+      ;;               "TA0042" 2}
+      ;;  :sort_order :asc}
+      :remap-list-max
+      (let [{:keys [remap-default remappings]} params
+            remappings (normalize-remappings remappings)]
+        (assert (every? (complement neg?) (vals remappings)) ":remap-list-max :remappings values must be non-negative")
+        {:_script
+         {:type "number"
+          :script {:lang "painless"
```
DeLaGuardo, line 47: painless script isn't available for ES5, is it?

test/ctia/entity/incident_test.clj
```diff
@@ -106,18 +106,20 @@
...
-        (assoc :title (str (java.util.UUID/randomUUID))
-               :revision (or order 0)))))
+(defn gen-new-incident
+  ([] (gen-new-incident "High"))
+  ([severity]
+   (let [order (ctim-severity-order severity)
+         _ (if (some? severity)
+             (assert (number? order)
+                     (str "Unmapped severity " (pr-str severity)))
+             (assert ((some-fn nil? number?) order)))]
```
DeLaGuardo, line 116: `ctim-severity-order` is a map of string-> number, I can't imagine a situation when it could return something different from nil or number. Then instead of assert it could be a bit more explicit `(or (ctim-severity-order severity) (throw (ex-info (str "Unmapped severity" severity) {})))`

marioaquino (approved):
LGTM - I don't totally understand the context around where the changes in this PR apply but I can't find anything to suggest improvement.

src/ctia/stores/es/sort.clj
frenchy64: It is https://www.elastic.co/blog/painless-a-new-scripting-language

test/ctia/entity/incident_test.clj
frenchy64: 100% agree, that's why it's an assertion :)

src/ctia/stores/es/sort.clj
```diff
@@ -18,6 +27,35 @@
...
+      ;; chooses the maximum value in field (a list) based on remappings,
+      ;; and sorts based on that value for each entity. Remappings must
+      ;; be positive integers.
+      ;; eg
+      ;; {:op :remap-list-max
+      ;;  :field-name "tactics"
+      ;;  :remappings {"TA0043" 1
+      ;;               "TA0042" 2}
+      ;;  :sort_order :asc}
+      :remap-list-max
```
yogsototh, line 39: As this bloc of code seems big, and also, it contains non trivial text (the painless script), perhaps it would be nice to define a function instead.

The function could have the comment as docstring.

src/ctia/stores/es/sort.clj
```diff
@@ -18,6 +27,35 @@
...
+      (let [{:keys [remap-default remappings]} params
+            remappings (normalize-remappings remappings)]
+        (assert (every? (complement neg?) (vals remappings)) ":remap-list-max :remappings values must be non-negative")
```
yogsototh, line 42: I guess `remappings` is set by the developers, but if this `remappings` is constructed somehow, the assertion could fail in a deployed env, and not in the developer env.

So perhaps would it be safer to replace this `assert` by throwing an Exception with detailed logs, but not much detail in the Exception itself, so the HTTP error could return a 500 without much detail about the issue. 

I'm not sure the middleware already handle correctly or not these kind of exception. So I could be totally wrong here.

src/ctia/stores/es/sort.clj
```diff
@@ -18,6 +27,35 @@
...
+        {:_script
+         {:type "number"
+          :script {:lang "painless"
+                   :inline (str/join
+                             "\n"
+                             ["if (!doc.containsKey(params.fieldName)) { return params.default }"
+                              "double score = params.default;"
+                              "for (int i = 0; i < doc[params.fieldName].length; ++i) {"
```
yogsototh, line 50: I could totally imagine a future dev to introduce a typo here, for example writing `params.fiedlName` instead of `params.fieldName`.
Would that be catch before deployment or `params.fiedlName` would be silently `null`?

If that's the case, perhaps it would be nice to have something that could make a quick and dirty parsing and check that every `params.X` has an associated field in the `:params`.

yogsototh (approved):
LGTM!

src/ctia/stores/es/sort.clj
frenchy64: I believe painless catches non-existent fields. It's similar to Java in that way.

https://www.elastic.co/guide/en/elasticsearch/painless/master/painless-operators-reference.html#field-access-operator

frenchy64: Hopefully the future dev will write unit tests like I have here, that's the only way I have confidence in these scripts.

src/ctia/stores/es/sort.clj
frenchy64: I removed the assert since it didn't make much sense anyway (it reflected a previous iteration).

src/ctia/stores/es/sort.clj
frenchy64: Combined the `:remap` and `:remap-list-max` cases, and moved comments to doc.

test/ctia/entity/incident_test.clj
DeLaGuardo: but then, why is assertion needed? you wouldn't assert that `str` always returns a string. How is it different from the mapping from string to number? I don't mean to be picky, but I still don't understand why these four lines of code with branching logic are needed.

PR #1298, opened 2022-08-23
Fix :test-selectors configuration

gbuisson:
Close [#6999](https://github.com/advthreat/iroh/issues/6999)

Investigating https://github.com/advthreat/iroh/issues/6999 I first focused on a potential issue involving ES, I tested improving the ductile library to support default timeouts without any effect.

I then noticed that the `test-selectors` config reads a file, it turns out that the `require` in "dev-resources/circleci_test/config.clj" fails the `read-string` call after and the selectors config return nil.

This has the effect of disabling selectors altogether, when you run `lein test` it runs everything including `:disabled`, `:generative` or `:sleepy`.

Fixing the configuration allows me to run the default suite locally in about 23-24min on my M1 8GB which is a great improvement! 

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -124,8 +124,8 @@
                  [metrics-clojure-ring ~metrics-clojure-version]
                  [clout "2.2.1"]
                  [slugger "1.0.1"]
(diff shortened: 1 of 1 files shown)


yogsototh (approved):
LGTM!

PR #1299, opened 2022-08-24
Fix some ES related tests

gbuisson:
> **Epic** #
> Close #
Related #6999

Since the tests are now running again locally, I spotted some failing items:

1. ES Authentication
some issues related to the default ES configuration change as we are now recommending to use authenticated connections by default for the local setup.

2. System indexes

new ES versions seem to be storing more items in some system indices, this fails tests relying on `/_cat/indices`

This PR fixes both classes of issues and should allow you to get a successful local test suite.

### DIFF
--- a/test/ctia/stores/es/crud_test.clj
+++ b/test/ctia/stores/es/crud_test.clj
@@ -277,6 +277,9 @@
    :aliased true
    :rollover {:max_docs 3}
    :refresh "true"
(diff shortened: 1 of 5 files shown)


msprunck (approved):
LGTM!

NeelSaini:
No QA required - comments from @Maksym after discussing with devs

PR #1301, opened 2022-09-05
Respect Judgement expiration in feed views

gbuisson:
Close[ #3470](https://github.com/advthreat/iroh/issues/3470)

Fix the Feed view route in CTIA by filtering expired judgements from the output.

couple things to note:

- I switched to the `read-records` store implementation instead of reading each record sequentially, this should in theory dramatically improve performance for large blocklists

- as a consequence from the change I need to increase the fetch limit of the relationships because we don't know in advance  if some judgements are expired, as I understand this limit was set and forgotten at the POC stage, we should monitor the production systems and determine if we need to keep it.
 

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
public: CTIA Feeds should now honor Judgement expiration
```

### DIFF
--- a/src/ctia/entity/feed.clj
+++ b/src/ctia/entity/feed.clj
@@ -27,6 +27,7 @@
             query-string-count
             query-string-search
             read-record
+            read-records
             update-record]]
    [ctia.stores.es.mapping :as em]
    [ctia.stores.es.store :refer [def-es-store]]
@@ -161,13 +162,18 @@
       (not feed) :not-found
       (not (valid-lifetime? lifetime)) :not-found
       (not= s (decrypt secret)) :unauthorized
-      :else (let [;; VERY IMPORTANT! inherit the identity from the Feed!
+      :else (let [ ;; VERY IMPORTANT! inherit the identity from the Feed!
                   feed-identity
                   {:login owner
                    :groups groups
                    :capabilities #{:read-judgement
                                    :read-relationship
                                    :list-relationships}}
+                  read-judgements (fn [ident ids]
+                                    (read-records (get-store :judgement)
+                                                  ids
+                                                  ident
+                                                  {}))
                   feed-results
                   (some->> (list-all-pages
                             :relationship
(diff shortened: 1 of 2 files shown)


test/ctia/entity/feed_test.clj
```diff
@@ -188,12 +193,14 @@
...
-                                 :body feed
-                                 :headers {"Authorization" "45c1f5e3f05d0"}))))))))))
+            (is (= 200 (:status
+                        (helpers/PUT app
+                                     (str "ctia/feed/" (:short-id feed-id))
+                                     :body feed
+                                     :headers {"Authorization" "45c1f5e3f05d0"}))))))))))
+(use-fixtures :once
+  (join-fixtures [validate-schemas
+                  whoami-helpers/fixture-server]))
```
msprunck, line 203: Why did you move the `use-fixtures` block?

src/ctia/entity/feed.clj
```diff
@@ -139,7 +140,7 @@
           (fn [feeds]
             (map #(decrypt-feed % services) feeds))))
 
-(def fetch-limit 200)
+(def fetch-limit 10000)
```
msprunck, line 143: The ElasticSearch documentation recommends to not request too many results at once with `from` and `size`. It can significantly increase memory and CPU usage. Why did you change this limit in the context of this PR?

msprunck (approved):
LGTM!

src/ctia/entity/feed.clj
msprunck, line 143: The ElasticSearch documentation recommends to not request too many results at once as it can significantly increase memory and CPU usage. If I'm not wrong, the `fetch-limit` set the page size used in the function `list-all-pages`. All pages are always fetched. I don't think you need to change this value to retrieve all judgements.

ereteog: I second Matt, you can raise it to limit the number of queries but it will retrieve all the relationships (which is potentially dangerous, we had trouble with it).

gbuisson: OK I will keep the current value

test/ctia/entity/feed_test.clj
gbuisson: oops! fixed that!

ereteog (approved):
lgtm

maksym-bielyshev:
QA Verified in release v1.105 in TEST Environment

PR #1303, opened 2022-09-14
Revert optimisations for bundle export

DeLaGuardo:
> Related 7cd8fe176b49036fd78285e2fa4dd80580d1ea42

This PR reverts changes for the bundle export function. After deployment on production, there were a few problems related to those optimisations. As a first step to handling those problems, it is better to revert the changes and rethink the optimisation strategy.
[commit](https://github.com/threatgrid/ctia/commit/7cd8fe176b49036fd78285e2fa4dd80580d1ea42#diff-20cea95187ae74e5a3f1cfd3770299e92b9ee526c8bf6cbdb5c791fd41627e69)

<a name="qa">[§](#qa)</a> QA
============================

1. Import that bundle [bundle-0.json.zip](https://github.com/threatgrid/ctia/files/8463390/bundle-0.json.zip). It contains two sightings. Note their ids after a successful import.
2. Replace in every file from that archive [bundles.zip](https://github.com/threatgrid/ctia/files/8463400/bundles.zip) all occurrences of substrings `[[0]]` and `[[1]]` with sighting's ids from the result of step 1.
3. Import all 5 bundles.
4. Call bundle export with ids from the result of step 1. It should retrieve corresponding sightings together with related indicators. The result should not contain more than 1000 indicators per sighting.


msprunck (approved):
LGTM!

NeelSaini:
Cherry Picked on TEST in rel v1.105

skriths:
Verified in TEST for Release 1.106

PR #1306, opened 2022-09-28
Stop using single-segment ns `java-time`, migrate to `java-time.api`

frenchy64:
There is a new namespace java-time.api which is exactly the same as java-time and coexists peacefully. Both are automatically generated and will stay in sync in the future. It was created because of this issue: https://github.com/dm3/clojure.java-time/issues/91

I like java-time.api better because it's not a single-segment name. What do we think?

```
intern: Stop using single-segment ns `java-time`, migrate to `java-time.api`
```


ereteog (approved):
I am totally cool with that :-)

PR #1309, opened 2022-10-10
Migration task: limit memory usage

ereteog:
> related #https://github.com/advthreat/iroh/issues/7166

This PR is an attempt to fix a memory leak.
It seems that the `reduce` of the batches and slice queries do not release the head of the queue and causes an OOM.

### DIFF
--- a/src/ctia/task/migration/migrate_es_stores.clj
+++ b/src/ctia/task/migration/migrate_es_stores.clj
@@ -185,14 +185,13 @@
              (pr-str query))
   (let [read-params (assoc migration-params :query query)
         data-queue (seque buffer-size
-                          (read-source read-params))]
-    (loop [total migrated-count
-           [batch & batches] data-queue]
-      (if batch
-        (recur (write-target total batch services) batches)
-        (assoc migration-params
-               :migrated-count
-               total)))))
+                          (read-source read-params))
+        new-migrated-count (reduce #(write-target %1 %2 services)
+                                   migrated-count
+                                   data-queue)]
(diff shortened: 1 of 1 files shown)


src/ctia/task/migration/migrate_es_stores.clj
```diff
@@ -150,7 +150,7 @@
    services :- mst/MigrationStoreServices]
   (let [migrated (transduce migrations conj documents)
         {:keys [data errors]} (list-coerce migrated)
-        new-migrated-count (+ migrated-count (count data))]
+        new-migrated-count (+ (count data) migrated-count)]
```
ereteog, line 153: I suspect that the reduce hold the queue due to this ([from this discussion](https://stackoverflow.com/questions/44586291/how-to-forget-headgcd-for-lazy-sequences-in-clojure))

gbuisson: it's not a very important remark but I think it would feel more natural if the state of the migrated count reside within the loop and not in the underlying function

src/ctia/task/migration/migrate_es_stores.clj
```diff
@@ -129,7 +129,7 @@
   [read-params :- (s/maybe BatchParams)]
   (lazy-seq
    (when-let [batch (read-source-batch read-params)]
-     (cons batch (read-source batch)))))
+     (cons batch (read-source (dissoc batch :documents))))))
```
ereteog, line 132: when reading the next batch, we do not need the documents anymore.

PR #1313, opened 2022-10-14
Upgrade ductile version 0.4.4 -> 0.4.5

DeLaGuardo:
> Related https://github.com/advthreat/iroh/issues/7056

Version 0.4.5 fixes encoding for the bulk insert operation. Those changes should fix bundle-import related issues when non-ASCII symbols got replaced with `?`.

<a name="qa">[§](#qa)</a> QA
============================

1. Import bundle with entities containing non-ASCII characters somewhere in the description.
2. Fetch that entity
3. Non-ASCII symbols should not be replaced with `?` or altered in any other way.

```
public: entities fields can have non-ASCII symbols
```

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -158,6 +158,58 @@ jobs:
         # poor man's travis_retry
         run: for i in {1..${ACTIONS_RETRY_TIMES}}; do lein warm-ci-deps && break; done
       - run: ./scripts/check-dependabot
+
+  test-encoding:
+    runs-on: ubuntu-20.04
+    needs: [setup]
+    steps:
+
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Start Docker environment
+        run: docker compose -f containers/dev/docker-compose.yml up --no-color --quiet-pull -d
+
+      - name: Maven Cache
+        id: maven-cache
+        uses: actions/cache@v3
+        with:
+          path: |
+            ~/.m2/repository
+            ~/.gitlibs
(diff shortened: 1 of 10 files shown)


.github/workflows/pr.yml
```diff
@@ -158,6 +158,58 @@ jobs:
         # poor man's travis_retry
         run: for i in {1..${ACTIONS_RETRY_TIMES}}; do lein warm-ci-deps && break; done
       - run: ./scripts/check-dependabot
+
+  test-encoding:
```
ereteog, line 162: @frenchy64 I let you approve this part :-)

ereteog (approved):
LGTM

frenchy64 (approved):
Nice, the test selectors live [here](https://github.com/threatgrid/ctia/blob/master/dev-resources/circleci_test/config.clj), they could also be updated in the same way as ductile's to exclude these tests if that's the intention.

DeLaGuardo:
> Nice, the test selectors live [here](https://github.com/threatgrid/ctia/blob/master/dev-resources/circleci_test/config.clj), they could also be updated in the same way as ductile's to exclude these tests if that's the intention.

Thanks, but I don't think encoding tests should be disabled via test selectors for "normal" test runs. They are speedy even if we run them twice.

frenchy64:
@DeLaGuardo Ah, I was more thinking the problem would be that `with-profile +test-encoding` is not enabled in normal runs.

skriths:
Verified Incident creation and updating with non-ASCII symbols in TEST for release 1.108

PR #1319, opened 2022-11-07
Delegate handling of 404 response to outer router

DeLaGuardo:
1. /doc/ routes now catch all paths instead of only those with extension
2. if a requested file can not be found in /resources/ctia/public/, return nil to indicate that the handler should return a 404 response to the client.
3. 404 responses for URLs like /doc/foo and /doc/foo.md now identical

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,9 +3,9 @@
 
 A Pragmatic, Operationalized Threat Intel Service and Data Model
 
-For full documentation see [doc](resources/public/doc/)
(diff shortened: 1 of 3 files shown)


msprunck (approved):

NeelSaini:
Cherry Picked on TEST in rel v1.109

PR #1322, opened 2022-11-25
Ensure pagination for fetching judgements of the feed

DeLaGuardo:
> Close advthreat/iroh#7281

- Add `ctia.stores.es.store/all-pages-iteration` function to iterate lazily over the result of ES query
- Ensure feed will not try to fetch more than `max_result_window` judgements or relationships at once

<a name="qa">[§](#qa)</a> QA
============================

1. Import a single indicator and a lot of judgements (more than 10000) related to that indicator via related relationships with `"source_ref": "%indicator-id%"`
2. Create a feed using indicator id from the result of the first step
3. Get a view of the feed `GET /ctia/feed/:feed-id/view?s=:secret`. Response should contain all related judgements and there are should be pagination headers (`X-Limit` and `X-Search_after`) present in the response.

```
intern: implement ctia.store/IPaginateableStore protocol for the store that needs pagination capabilities
```


src/ctia/entity/feed.clj
```diff
@@ -218,226 +218,251 @@
      :responses {404 {:schema s/Str}
                  401 {:schema s/Str}}
      :query-params [s :- (describe s/Str "The feed share token")]
-     (let [{:keys [output]
-            :as feed} (fetch-feed id s services)]
+     :header-params [{x-search_after :- (s/maybe s/Str) "nil"}
+                     {x-limit :- (s/maybe s/Str) "nil"}]
```
msprunck, line 222: Could you provide the same [pagination interface](https://github.com/threatgrid/ctia/blob/master/src/ctia/http/routes/common.clj#L51) we already expose in the other API endpoints? They use `limit`, `offset` and `search_after` as query parameters.

src/ctia/stores/es/store.clj
```diff
@@ -15,6 +20,23 @@
...
 
+(defn all-pages-iteration
+  "Returns lazy iteration of consecutive calls to `query-fn` with pagination params.
+
+  Resulted data is a sequence of responses of a shape `{:data [,,,] :paging {:next {,,,}}}`"
+  ([query-fn] (all-pages-iteration query-fn {}))
+  ([query-fn {:keys [limit]
+              :or {limit default-limit}
+              :as params}]
+   (let [limit (if (> limit max-result-window)
```
msprunck, line 31: `max-result-window` is a default value which can be less or greater than the value in the index settings. Why not use the value that is in the settings?

src/ctia/store.clj
```diff
@@ -27,16 +25,19 @@
   (delete-identity [this org-id role]))
 
 (defprotocol IEventStore
-  (read-event [this id ident params])
-  (create-events [this new-events])
-  (list-events [this filtermap ident params]))
+  (create-events [this new-events]))
```
msprunck, line 28: `create-record` accepts multiple records. I'm wondering if we could get rid of the `IEventStore`. Not in the scope of this PR

src/ctia/entity/feed.clj
DeLaGuardo: done

src/ctia/store.clj
DeLaGuardo: The only reason I kept IEventStore protocol is that the signatures of create-events and create-records are incompatible. This could be solved by adding an arity-2 variant for create-records of IStore protocol. But as you mention, it would be good to keep it in separate PR.

src/ctia/stores/es/store.clj
DeLaGuardo: because this function intentionally doesn't know about settings. There is a place to provide a limit value explicitly via `params` and only if missing — fallback to the "very default" limit provided by ductile library.

src/ctia/stores/es/store.clj
```diff
@@ -15,6 +20,23 @@
...
+                 max-result-window
+                 limit)]
```
msprunck, line 33:
```suggestion
   (let [limit (min limit max-result-window)]
```

> only if missing — fallback to the "very default" limit provided by ductile library.
I'm not sure about this behaviour. If `limit` is `>` than the default-max-result-window but less than the `max-result-window`, `default-max-result-window` is used

DeLaGuardo: I identify those cases:
* limit wasn't provided — set it to `ductile/default-limit` which is 100
* limit is provided and its value less than `ductile/max-result-window` — keep the value
* limit is provided but more than `ductile/max-result-window` — use max-result-window

PS: I can see your point after I wrote this ^
Probably it would be better to handle only the case when the limit isn't set in params and keep it as is overwise. WDYT?

msprunck: yes sounds good

msprunck (approved):
LGTM!

ereteog (approved):
that's really nice! thank you!

trcollier:
QA verified in TEST release v1.112

PR #1325, opened 2022-12-01
Standardize event store

DeLaGuardo:
Refactoring to remove `ctia.store/IEventStore` protocol as obsolete in favour of standard `ctia.store/IStore` protocol and `ctia.stores.es.store/def-es-store` macro.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: please use `ctia.stores.es.store/def-es-store` macro for every new Entity implementation based on Elasticsearch
```


ereteog (approved):
thx!

PR #1331, opened 2023-01-24
Add techniques to enumerable fields

ereteog:
> Related #https://github.com/advthreat/iroh/issues/7455

This PR adds `techniques` mapping and adds this field as enumerable fields for topn metric.

<a name="qa">[§](#qa)</a> QA
============================

1) request topn techniques for incidents `GET /ctia/incident/metric/topn?from=2022-06-01T00.00.00.000Z&aggregate-on=techniques`
2) validate that the response do have MITRE techniques ordered by their frequency in incidents.

### DIFF
--- a/src/ctia/entity/incident.clj
+++ b/src/ctia/entity/incident.clj
@@ -142,6 +142,7 @@
       :promotion_method em/token
       :severity         em/token
       :tactics          em/token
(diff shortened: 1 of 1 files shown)


DeLaGuardo (approved):
LGTM

PR #1336, opened 2023-01-27
Default "no-pagination" for feed

DeLaGuardo:
> **Epic** #
> Close #
> Related #

Return all related entities in case of no explicit pagination params.

<a name="qa">[§](#qa)</a> QA
============================

1. Import a single indicator and a lot of judgements (more than 10000) related to that indicator via related relationships with `"source_ref": "%indicator-id%"`
2. Create a feed using indicator id from the result of the first step
3. Get a view of the feed `GET /ctia/feed/:feed-id/view?s=:secret`. The response should contain all related judgements without pagination headers.

```
```

### DIFF
--- a/src/ctia/entity/feed.clj
+++ b/src/ctia/entity/feed.clj
@@ -133,7 +133,7 @@
   ([id share-token services]
    (fetch-feed id share-token {} services))
   ([id share-token
-    page-params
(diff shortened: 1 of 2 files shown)


msprunck (approved):
LGTM!

PR #1340, opened 2023-02-17
Scores dynamic mapping

frenchy64:
Epic: https://github.com/advthreat/iroh/issues/7341

We want incident `scores` to be an open map, which can be achieved via dynamic mapping. See explanation in https://github.com/threatgrid/ctim/pull/398 for context.

Will require a migration for INT.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: Scores dynamic mapping
```

### DIFF
--- a/src/ctia/entity/incident.clj
+++ b/src/ctia/entity/incident.clj
@@ -30,14 +30,12 @@
 
 (def incident-bundle-default-limit 1000)
 
-(def isIncident (st/assoc is/Incident :scores {s/Keyword s/Num}))
-
 (def-acl-schema Incident
-  isIncident
+  is/Incident
   "incident")
 
 (def-acl-schema PartialIncident
-  (fu/optionalize-all isIncident)
+  (fu/optionalize-all is/Incident)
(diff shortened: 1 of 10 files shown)


src/ctia/entity/incident.clj
```diff
@@ -126,6 +126,12 @@
 (def incident-mapping
   {"incident"
    {:dynamic false
+    :date_detection false
+    :numeric_detection true
+    :dynamic_templates
+    [{:num
+      {:match_mapping_type "long"
```
ereteog, line 133: Scores could be float.
```suggestion
      {:match_mapping_type "float"
```

frenchy64: I thought the point of this was to coerce long=>float? IIUC your change is a float=>float coercion?

ereteog: yes you are totally right, ignore my comment :-)

ereteog (approved):
FOR THE WIN!

PR #1342, opened 2023-02-28
Note Entity API changes

gbuisson:
This PR accommodates for the changes in the CTIM Note entity, namely:

- remove `entity_id` and introduce `related_entities` in the schema
- introduce `note_class` in the schema
- updated the entity search query parameters

<a name="qa">[§](#qa)</a> QA
============================

You can test the API following all the sample calls in the design documentation

  

```
public: updated Note Entity API
```

### DIFF
--- a/src/ctia/entity/note/schemas.clj
+++ b/src/ctia/entity/note/schemas.clj
@@ -37,8 +37,8 @@
 (def note-fields
   (concat sorting/base-entity-sort-fields
           sorting/sourcable-entity-sort-fields
-          [:related_entities.id
-           :related_entities.type
+          [:related_entities.entity_id
+           :related_entities.entity_type
            :note_class
            :author
            :content]))
@@ -49,8 +49,8 @@
 (def searchable-fields
   #{:id
     :source
-    :related_entities.id
-    :related_entities.type
+    :related_entities.entity_id
+    :related_entities.entity_type
     :note_class
     :content})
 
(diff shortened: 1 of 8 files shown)


yogsototh (approved):
LGTM!

src/ctia/entity/note/schemas.clj
```diff
@@ -47,11 +49,13 @@
 (def searchable-fields
   #{:id
     :source
-    :entity_id
+    :related_entities.entity_id
+    :related_entities.entity_type
+    :note_class
```
ereteog, line 54: searchable fields are for fields that we want to sort on by default, without actively filtering on it. I do not think that's desired here.

src/ctia/stores/es/mapping.clj
```diff
@@ -256,6 +256,12 @@
     :columns {:enabled false}
     :rows {:enabled false}}})
 
+(def note-related-entity
+  {:type "object"
+   :properties
+   {:entity_type token
+    :entity_id token}})
```
ereteog, line 263: We tend to put entity specific mappings in the corresponding ns.

src/ctia/entity/note/schemas.clj
gbuisson: I removed the unneeded fields

src/ctia/stores/es/mapping.clj
gbuisson: done

PR #1343, opened 2023-03-02
Fix bulk relationships between transient asset mappings/fields

frenchy64:
Close https://github.com/advthreat/iroh/issues/7611

We need to realize _all_ entities before creating relationships.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: Fix bulk relationships between transient asset mappings/fields
```

### DIFF
--- a/src/ctia/bulk/core.clj
+++ b/src/ctia/bulk/core.clj
@@ -290,41 +290,32 @@
     {{:keys [get-in-config]} :ConfigService :as services} :- APIHandlerServices]
    (let [{:keys [refresh]
           :or   {refresh (bulk-refresh? get-in-config)}} params
-         new-entities (gen-bulk-from-fn
-                       create-entities
-                       (dissoc
-                        bulk
-                        :relationships
-                        ;; AssetMapping and AssetProperties have asset-ref field
-                        ;; that needs to be resolved, so we delay the creation
-                        ;; of these entities to be after we create Assets
-                        :asset_mappings :asset_properties)
-                       tempids
-                       login
-                       {:refresh refresh}
-                       services)
-         entities-tempids (into tempids (merge-tempids new-entities))
-         new-linked-ents (gen-bulk-from-fn
-                          create-entities
-                          (select-keys
-                           bulk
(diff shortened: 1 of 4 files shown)


agzam (approved):

src/ctia/bulk/core.clj
```diff
@@ -290,41 +290,32 @@
...
+                                                (-> acc
+                                                    (update :entities merge entities)
+                                                    (update :tempids into (merge-tempids entities)))))
+                                            {:entities {}
+                                             :tempids tempids}
+                                            [(dissoc bulk :relationships :asset_mappings :asset_properties)
+                                             ;; resolve asset_ref on AssetMapping and AssetProperties
+                                             (select-keys bulk [:asset_mappings :asset_properties])
+                                             ;; resolve transient ids on relationships. all other entities must be realized.
+                                             (select-keys bulk [:relationships])])
```
ereteog, line 310: that's a lot easier to read now, thx!
Just a matter of taste, but for reduce I like to name the `fn` to summarize what it does and potentially unit test it: 

``` clojure
(s/defn EntitiesTempIds
    {:entities s/Map :tempids s/Map}

(s/defn concat-create-entities :- EntitiesTempIds
    [entities-tempids  :- EntitiesTempIds
     new-entities :- [s/Map]]
    ...
    )

(s/defn create-bulk
    ...
    (reduce concat-create-entities
            {:entities {} :tempids tempids}
             [(dissoc bulk :relationships :asset_mappings :asset_properties)
              ;; resolve asset_ref on AssetMapping and AssetProperties
             (select-keys bulk [:asset_mappings :asset_properties])
             ;; resolve transient ids on relationships. all other entities must be realized.
             (select-keys bulk [:relationships])
     ...)
```

ereteog (approved):
2 years old bug :-).
Good catch!

src/ctia/bulk/core.clj
frenchy64: It was more effective to unit test by abstracting the `reduce`, so that's what I did.

PR #1346, opened 2023-03-08
bump snakeyaml to address CVE-2022-38751

yogsototh:
Bump snakeyaml version to fix a security issue.

See [this conversation](url) to confirm we are not vulnerable to CVE-2022-1471.

The only place that CTIA might parse yaml is during trapperkeeper's [bootstrap](https://github.com/puppetlabs/trapperkeeper/blob/5ac89448d149838f19649856b17178c7d943d1da/src/puppetlabs/trapperkeeper/config.clj#L61). But we don't use yaml configs and we'd always trust our own configuration anyway.

### DIFF
--- a/dependabot/dependency-tree.txt
+++ b/dependabot/dependency-tree.txt
@@ -27,8 +27,6 @@ ctia:ctia:jar:1.1.1-SNAPSHOT
 |  +- clj-commons:fs:jar:1.6.307:compile
 |  |  +- org.apache.commons:commons-compress:jar:1.20:compile
 |  |  \- org.tukaani:xz:jar:1.8:compile
-|  +- clj-commons:clj-yaml:jar:0.7.2:compile
-|  |  \- org.flatland:ordered:jar:1.5.9:compile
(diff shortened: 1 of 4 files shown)


msprunck (approved):
LGTM!

ereteog (approved):

frenchy64:
I think we need 2.0, right? https://github.com/advthreat/iroh/security/dependabot/31

frenchy64:
As long as we don't use snakeyaml directly, apparently clj-yaml is immune to this CVE by default. I think we just need to check if we use clj-yaml safely. TK also uses it. https://github.com/clj-commons/clj-yaml/pull/83

frenchy64:
Actually, there are 2 vulns:
- https://nvd.nist.gov/vuln/detail/CVE-2022-38751 fixed in 1.31
- https://nvd.nist.gov/vuln/detail/CVE-2022-1471 fixed in 2.0

We might not need the latter due to clj-yaml, but I think 1.33 is a good version to fix the former.

PR #1349, opened 2023-03-13
Bump to CTIM 1.3.5

msprunck:
<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

### DIFF
--- a/src/ctia/entity/sighting/es_store.clj
+++ b/src/ctia/entity/sighting/es_store.clj
@@ -36,7 +36,9 @@
       :internal em/boolean-type
       :observables em/observable
       :observables_hash em/token
-      :relations em/observed-relation})}})
+      :relations em/observed-relation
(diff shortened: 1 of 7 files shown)


src/ctia/entity/sighting/es_store.clj
```diff
@@ -36,7 +36,9 @@
       :internal em/boolean-type
       :observables em/observable
       :observables_hash em/token
-      :relations em/observed-relation})}})
+      :relations em/observed-relation
+      :context {:type "object"
+                :enabled false}})}})
```
ereteog, line 41: Dynamic being false at the top level of the entity, it should not be indexed if you do not define this mapping.

ereteog (approved):
LGTM

NeelSaini:
Verified on TEST in rel v1.117 by @James

PR #1351, opened 2023-03-16
Exception handling for bundle export

DeLaGuardo:
> Related advthreat/iroh#7635

* rethrow cause instead of java.util.concurrent.Exception
* allow handle-read to suppress access control error as in handle-read-many
* bundle-export should ignore inaccessible entities instead of throwing.

<a name="qa">[§](#qa)</a> QA
============================

1. TBD
2.
3.

```
intern: handle-read now can accept `:suppress-access-control-error?` option. 
public: `bundle-export` will ignore unaccessible entities instead of throwing exception.
```

### DIFF
--- a/src/ctia/bundle/core.clj
+++ b/src/ctia/bundle/core.clj
@@ -7,8 +7,8 @@
    [ctia.auth :as auth]
    [ctia.bulk.core :as bulk]
    [ctia.bundle.schemas :refer
-    [BundleImportData BundleImportResult
-     EntityImportData FindByExternalIdsServices]]
+    [BundleImportData BundleImportResult EntityImportData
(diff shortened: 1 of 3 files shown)


frenchy64 (approved):
LGTM!

ereteog (commented):
what about using `handle-read-many` to remove the pmap and that already has the logic for the exception
https://github.com/threatgrid/ctia/blob/9ba6e8466212a37fa3294d124dc0109776dff7d3/src/ctia/stores/es/crud.clj#L209

ereteog (approved):
Nice!

NeelSaini:
Verified on TEST in rel v1.117 by @James

PR #1353, opened 2023-03-20
Incident score schema check

ereteog:
> Related #https://github.com/threatgrid/ctia/pull/1340

Enable to dynamically generate schemas from config and apply it to incident scores.

<a name="qa">[§](#qa)</a> QA
============================

Check that the incident creation and update schemas is (check it on Bulk and Bundle too)
``` json
{
  "asset": 0,
  "ttp": 0,
  "probability": 0,
  "global": 0
}
```


frenchy64 (approved):
Nice!

PR #1356, opened 2023-03-27
add tactics/techniques to incident search filters

ereteog:
> Related #https://github.com/advthreat/iroh/issues/7455

Simplifies tactics and techniques filtering in search and metrics for incidents.

<a name="qa">[§](#qa)</a> QA
============================

Check that you can filter incidents on `tactics`  and `techniques`


gbuisson:
I would change the filters to accept arrays of Strings instead of a single string.

frenchy64 (approved):

PR #1359, opened 2023-04-03
Allow setting `allow_partial_search_results` in ES search queries

gbuisson:
> **Epic** #
> Close https://github.com/advthreat/iroh/issues/3057
> Related #

- Expose a new configuration property at the ES Store level `allow_partial_search_results`
- This boolean is transmitted in the ES query params if specified

This ES option is documented here: https://www.elastic.co/guide/en/elasticsearch/reference/7.17/search-search.html

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
public: allow setting the `allow_partial_search_results` as an ES Store option
```

### DIFF
--- a/src/ctia/stores/es/crud.clj
+++ b/src/ctia/stores/es/crud.clj
@@ -453,13 +453,21 @@ It returns the documents with full hits meta data including the real index in wh
    :props s/Any
    (s/optional-key :sort-extension-definitions) SortExtensionDefinitions})
 
+(defn es-seven-configured?
+  "given ES store properties, check it is configured to use ES7"
+  [{:keys [version]}]
+  (<= 7 version))
+
 (s/defn make-query-params :- {s/Keyword s/Any}
   [{:keys [params props sort-extension-definitions]} :- MakeQueryParamsArgs]
-  (cond-> (-> params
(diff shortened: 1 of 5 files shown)


src/ctia/stores/es/crud.clj
```diff
@@ -453,13 +453,19 @@ It returns the documents with full hits meta data including the real index in wh
    :props s/Any
    (s/optional-key :sort-extension-definitions) SortExtensionDefinitions})
 
+(defn seven? [{:keys [version]}]
```
ereteog, line 456: the name of this function partially reflects what it does ;-)

ereteog (commented):
I would not make it configurable and hardcode it to force the behaviour that we prefer.
However I wonder if we should not inject it when we call the store.
In some cases, we might prefer diffrent behavious:
- search route: should we favor partial results than a blank page?
- bulk / bundle operations: we do not want partial results when we lookup external references.

wdyt?

src/ctia/stores/es/crud.clj
gbuisson: @ereteog I renamed the function and added a docstring

ereteog (approved):

PR #1358, opened 2023-04-03
Average aggregation

frenchy64:
Related https://github.com/advthreat/iroh/issues/7380

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

<a name="release-notes">[§](#release-notes)</a> Release Notes
=============================================================

```
intern: one-line description for internal eyes only.
public: one line description for public release notes.
```

<a name="squashed-commits">[§](#squashed-commits)</a> Squashed Commits
======================================================================

PR #1361, opened 2023-04-18
Push sighting store's coercion pattern into def-es-store

frenchy64:
Related https://github.com/advthreat/iroh/issues/7380
Related https://github.com/threatgrid/ctia/pull/1358

Sighting store has ESStoredSighting (the Sighting stored in ES) and StoredSighting (user-facing schema) so we can store internal non-CTIM fields. We will soon need this for other stores, so this pushes the logic into `def-es-store`.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: Start pushing sighting store's coercion pattern into stores abstraction
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -463,6 +463,15 @@ or from source with leiningen:
 lein run -m ctia.task.update-index-state
 ```
 
+# Frequently Asked Questions (FAQ)
+
+Q: How do I stop seeing `Unauthorized ES request` when running CTIA locally?
+A: Uncomment the following lines in `resources/ctia-default.properties`
+```
+ctia.store.es.default.auth.type=basic-auth
+ctia.store.es.default.auth.params.user=elastic
+ctia.store.es.default.auth.params.pwd=ductile
+```
 
 ## License
 
--- a/test/ctia/stores/es/store_test.clj
+++ b/test/ctia/stores/es/store_test.clj
@@ -110,39 +110,35 @@
     (let [g (gensym)]
       (is (= g (-> g ->SightingStore :state)))))
   (testing ":store-opts transformers"
-    (helpers/fixture-ctia-with-app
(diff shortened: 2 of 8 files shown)


ereteog (commented):
This implementation is one step deeper in the abstraction. Why not fully implementing it in def-es-store?
Even if it might be somehow tested in sightings test, considering how central is the store implementation, I would secure it with dedicated unit tests.

README.md
```diff
@@ -463,6 +463,15 @@ or from source with leiningen:
...
 
+# Frequently Asked Questions (FAQ)
+
+Q: How do I stop seeing `Unauthorized ES request` when running CTIA locally?
+A: Uncomment the following lines in `resources/ctia-default.properties`
+```
+ctia.store.es.default.auth.type=basic-auth
+ctia.store.es.default.auth.params.user=elastic
+ctia.store.es.default.auth.params.pwd=ductile
+```
```
ereteog, line 474: :-)

test/ctia/stores/es/store_test.clj
```diff
@@ -1,14 +1,27 @@
...
             [ctia.test-helpers.fixtures :as fixt]
             [ctia.test-helpers.core :as helpers]
             [ctim.examples.incidents :refer [incident-minimal]]
-            [clojure.test :refer [deftest testing is]]
+            [clojure.test :refer [deftest testing is use-fixtures]]
             [schema.core :as s]
             [schema-tools.core :as st]))
 
+(use-fixtures :once
+              mth/fixture-schema-validation
```
ereteog, line 21: At some point it would be good to get rid of `clj-momo`.
Please use [schema.test/validate-schemas](https://plumatic.github.io/schema/schema.test.html#var-validate-schemas)

PR #1364, opened 2023-05-03
Revert "Push sighting store's coercion pattern into def-es-store"

frenchy64:
Reverts threatgrid/ctia#1361

For some reason this commit fails to queue a build on the master branch and v1.120. Let's try reverting and then investigating after the v1.120 cut.


obarbeau (approved):
approve the revert

PR #1371, opened 2023-05-26
Fix X-Total-Hits in incident average metric

frenchy64:
**Epic** https://github.com/advthreat/iroh/issues/7380

X-Total-Hits should be the divisor in the incident average metric. Instead, it's all the incidents created in the from/to range of the query. This fixes the problem by adding extra filters to restrict such queries to only incidents that have the relevant field in the aggregation.

For example, if getting the average of the intervals.new_to_opened field, we add a filter to the aggregation query that filters incidents that only have the intervals.new_to_opened field.

To test this, I create extra garbage incidents within the from/to range of the existing unit test. This fails the existing test and makes the count higher than it should. Then, by adding extra filters, the test passes again.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: Fix X-Total-Hits in incident average metric
```

### DIFF
--- a/src/ctia/stores/es/crud.clj
+++ b/src/ctia/stores/es/crud.clj
@@ -590,23 +590,29 @@ It returns the documents with full hits meta data including the real index in wh
 
 (s/defn make-search-query :- {s/Keyword s/Any}
   "Translate SearchQuery map into ES Query DSL map"
-  [es-conn-state :- ESConnState
-   search-query :- SearchQuery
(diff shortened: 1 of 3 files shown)


rplevy (approved):
LGTM!

ereteog (commented):
Great! 
We need to generalize it to other metrics https://github.com/advthreat/iroh/issues/7928

PR #1374, opened 2023-06-15
Further restrict scope of incident average metric

frenchy64:
**Epic** https://github.com/advthreat/iroh/issues/7380

To prevent outliers, the incident average time to investigate/resolve should only include incidents that are both created and acted (i.e., opened or closed) within the timeframe the user provides via the from/to query params. Before this, we could have incidents that are created in the current timeframe, but closed long in the future. These cases yield averages that are larger than the current timeframe, which was not useful for our use case.

Conceptually, the main extension in this PR is to allow multiple `range` queries on different fields. They must be of the same date range. We could be a little smarter in separating `from` to created and `to` to opened/closed time, but there's not much benefit compared to the code necessary to enable that.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: Further restrict incident average metric
```


ereteog (changes requested):
IMO that's a custom API client need that can be adressed with search filters

PR #1377, opened 2023-07-05
Filter incidents on timestamp not created

agzam:
> Close [advthreat/iroh#7917 — CTIA: filter incidents on timestamp instead of created](https://github.com/advthreat/iroh/issues/7917)
> Close https://github.com/advthreat/iroh/issues/8519

<a name="qa">[§](#qa)</a> QA
============================

Testing this feature essentially entails creating an incident (or multiple) and ensuring that its timestamp remains unchanged after modifying it.

1. Create a minimal Incident

```
POST https://intel.test.iroh.site/ctia/incident
Accept: application/json
Content-Type: application/json
Authorization: Bearer TOKEN

    {
    "incident_time" : {
        "opened" : "2016-02-11T00:40:48Z"
    },
    "status" : "Open",
    "confidence" : "High",
    "severity" : "High",
    "title" : "Some incident For Testing",
    "revision" : 3
    }
```
    
 This request returns an entity with an Id field that would be something like: 
`https://intel.test.iroh.site/ctia/incident/ctia/incident/incident-66c2b363-42a4-4e76-9e0b-00317ba1a29e`. We call these "long-ids".

2. Grab `incident-66c2b363-42a4-4e76-9e0b-00317ba1a29e` part of it. We internally call them "short-ids"

If you're testing with curl in the command line, you can pipe it to `jq` and `sed` like so:

```
curl 'https://intel.test.iroh.site/ctia/incident' \
-H 'Accept: application/json' \
-H 'Content-Type: application/json' \
-H 'Authorization: Bearer TOKEN' \
-X POST \
--data-raw '{
"incident_time" : {
    "opened" : "2016-02-11T00:40:48Z"
},
"status" : "Open",
"confidence" : "High",
"severity" : "High",
"title" : "Some Incident For Testing",
"revision" : 3
}
' | jq .id | sed 's|.*/||' | sed 's/"$//'
```

*jq is a cmd-line JSON processor that basically tells it to retrieve `id:` field value from the query result. 
sed - stands for "stream editor" and here it says to remove everything but the last segment of the long id url, and remove the trailing double quote as well*

3. Note the timestamp of the Incident you created

```
curl 'https://intel.test.iroh.site/ctia/incident/search/incident-66c2b363-42a4-4e76-9e0b-00317ba1a29e' \
-H 'Accept: application/json' \
-H 'Authorization: Bearer TOKEN' \
-H 'Content-Type: application/json' | jq .timestamp
```

*Remember, the last segment in the request should be the short id of the incident created* 

4. Modify created Incident

Patch the incident you just created, adjusting one of the fields, e.g., `status`. It was an Incident with the status `Open`, now change it to `Incident Reported`. Different possible values for Incident Status are [listed in CTIM](https://github.com/threatgrid/ctim/blob/master/src/ctim/schemas/vocabularies.cljc#L346)

```
    curl 'https://intel.test.iroh.site/ctia/incident/incident-66c2b363-42a4-4e76-9e0b-00317ba1a29e' \
    -H 'Accept: application/json' \
    -H 'Content-Type: application/json' \
    -H 'Authorization: Bearer TOKEN' \
    -X PATCH \
    --data-raw '{
    "status" : "Incident Reported"
    }'
```

5. Finally, compare the timestamp of the modified Incident with the timestamp at the time of the initial creation.

**The timestamps should match**

```
```

### DIFF
--- a/src/ctia/entity/incident.clj
+++ b/src/ctia/entity/incident.clj
@@ -1,7 +1,6 @@
 (ns ctia.entity.incident
   (:require
    [clj-momo.lib.clj-time.core :as time]
-   [clj-time.core :as clj-time]
    [clojure.string :as str]
    [ctia.domain.entities
     :refer [default-realize-fn un-store with-long-id]]
@@ -85,13 +84,12 @@
     [rt-ctx :- GraphQLRuntimeContext]
     (let [rfn (lift-realize-fn-with-context
                incident-default-realize rt-ctx)
-          e (-> (rfn new-obj id tempids ident-map prev-obj)
-                #_(assoc :timestamp
-                         (or (:timestamp new-obj)
-                             (:timestamp prev-obj)
-                             (clj-time/now))))]
-
-      e)))
+          now (time/internal-now)]
+      (-> (rfn new-obj id tempids ident-map prev-obj)
+          (assoc :timestamp
+                 (or (:timestamp new-obj)
+                     (:timestamp prev-obj)
+                     now))))))
 
 (s/defschema IncidentStatus
   (fs/->schema vocs/Status))
--- a/src/ctia/entity/incident.clj
+++ b/src/ctia/entity/incident.clj
@@ -1,17 +1,24 @@
 (ns ctia.entity.incident
   (:require
-   [clojure.string :as str]
-   [java-time.api :as jt]
    [clj-momo.lib.clj-time.core :as time]
+   [clj-time.core :as clj-time]
+   [clojure.string :as str]
    [ctia.domain.entities
     :refer [default-realize-fn un-store with-long-id]]
    [ctia.entity.feedback.graphql-schemas :as feedback]
    [ctia.entity.relationship.graphql-schemas :as relationship-graphql]
    [ctia.flows.crud :as flows]
+   [ctia.flows.schemas :refer [with-error]]
+   [ctia.graphql.delayed :as delayed]
    [ctia.http.routes.common :as routes.common]
    [ctia.http.routes.crud :as routes.crud]
    [ctia.lib.compojure.api.core :refer [POST routes]]
-   [ctia.schemas.core :refer [APIHandlerServices def-acl-schema def-stored-schema SortExtensionDefinitions]]
+   [ctia.schemas.core :refer [APIHandlerServices
+                              GraphQLRuntimeContext
+                              RealizeFnResult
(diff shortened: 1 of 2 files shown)


agzam:
@ereteog can you please take a look, is that close to what you had in mind?

ereteog (commented):
Yes it applies what I described.
Can you populate the QA section please?

agzam:
@ereteog Oh, It looks like I still have to address the part where you said:

> Then we need to have from and to applied to timestamp instead of [the default created](https://github.com/threatgrid/ctia/blob/9134c455ef0c7f7da150359b209c6583449e7d48/src/ctia/http/routes/crud.clj#LL180C10-L180C20) by adding :date-field :timestamp [here](https://github.com/threatgrid/ctia/blob/9134c455ef0c7f7da150359b209c6583449e7d48/src/ctia/entity/incident.clj#L416).

I'm still not very sure what do we have to change for that. Maybe we should do it in a separate PR?

ereteog:
> @ereteog Oh, It looks like I still have to address the part where you said:
> 
> > Then we need to have from and to applied to timestamp instead of [the default created](https://github.com/threatgrid/ctia/blob/9134c455ef0c7f7da150359b209c6583449e7d48/src/ctia/http/routes/crud.clj#LL180C10-L180C20) by adding :date-field :timestamp [here](https://github.com/threatgrid/ctia/blob/9134c455ef0c7f7da150359b209c6583449e7d48/src/ctia/entity/incident.clj#L416).
> 
> I'm still not very sure what do we have to change for that. Maybe we should do it in a separate PR?

Yes you can do it in a separate PR (I changed Close to Related)

ereteog (approved):
LGTM.
You should consider my remark to simplify the work of QA

jamoser-cisco:
I wish all QA sections were so detailed, thanks @agzam!

I verified patching a couple of incidents alone does not change the timestamp, however I am able to PATCH the timestamp value itself.  Is this intended??

jamoser-cisco:
@ereteog, maybe I should have tagged you to ask that?? After reading the linked tickets it seems agzam is no longer with Cisco??

jamoser-cisco:
and it seems you answered that question in the same comment I learned that in... =)

ereteog:
> I verified patching a couple of incidents alone does not change the timestamp, however I am able to PATCH the timestamp value itself. Is this intended??

Yes that's the right behaviour

PR #1379, opened 2023-07-21
Do not init disabled stores

frenchy64:
Close https://github.com/advthreat/iroh/issues/8042

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: Do not init disabled stores
```

### DIFF
--- a/src/ctia/graphql/delayed.clj
+++ b/src/ctia/graphql/delayed.clj
@@ -13,6 +13,7 @@
   [& body]
   `(DelayedGraphQLWrapper.
      (s/fn ~@body)))
+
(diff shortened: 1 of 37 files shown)


DeLaGuardo (approved):
amazing!

ereteog (commented):
that's nice.
Can you add a test that verifies that the store creation is actually disabled when [configured](https://github.com/advthreat/tenzin-config/blob/31bda6a59b043c323de1519ac242443cb3fa1662/ctia/INT/us-east-1/ctia-2.0.0.properties#L105) for that?

PR #1383, opened 2023-09-11
Patch existing entities in `POST /bundle/import`

frenchy64:
**Epic** https://github.com/advthreat/iroh/issues/7341
Related https://github.com/advthreat/iroh/issues/8207

Current behavior of bundle import on existing entities is to ignore them. This PR changes behavior to patch in this case, and loosens the schema for bundle import so then partial entities are allowed if patching.

<a name="qa">[§](#qa)</a> QA
============================

This PR adds the ability for `POST /bundle/import` to patch entities. The following steps test if this procedure works.

1. Create an existing entity (for example an Indicator) and note its id.
2. Patch this entity using `POST /bundle/import`. For example, patching the `:producer` field of the indicator:

```
POST /bundle/import?patch-existing=true
{:indicators [{:id "<id-from-step-1>" :producer "PATCHED"}]}
```

3. Verify that the result looks something like this (the `"updated"` part is the most important part):
```
{:results [{:result "updated" :id "<id-from-step-1>"}]}
```

4. Look up the indicator via `GET /ctia/indicator` to verify that the `producer` field was updated.

```
intern: Patch existing entities in POST /bundle/import
```

### DIFF
--- a/resources/ctia/public/doc/bulk-bundle.org
+++ b/resources/ctia/public/doc/bulk-bundle.org
@@ -671,6 +671,12 @@ When a bundle is submitted:
 
 If more than one entity is referenced by the same external ID, an error is reported.
 
+If provided query parameter `asset_properties-merge-strategy=merge-overriding-previous`, then when patching asset properties,
+existing asset properties will be retrieved and combined with the asset properties in the request bundle
+as if by concatenating existing and new properties together in a single list,
+removing properties to the left of a property with the same name, 
+then sorting the list lexicographically by name before using this list to patch the existing entity.
+
 Example for an incident along with its context, note the transient ids
 #+begin_src HTTP
 POST /ctia/bundle/import HTTP/1.1
--- a/resources/ctia/public/doc/bulk-bundle.org
+++ b/resources/ctia/public/doc/bulk-bundle.org
@@ -667,7 +667,7 @@ When a bundle is submitted:
 
 1. All entities that have already been imported with the external ID whose prefix has been configured with the `ctia.store.external-key-prefixes` property are searched.
 2. If they are identified by transient IDs, a mapping table between transient and stored IDs is built.
-3. New entities are created in the same way as the `/bulk` API endpoint with transient IDs resolutions. Existing entities are patched similarly.
+3. New entities are created in the same way as the `/bulk` API endpoint with transient IDs resolutions. If `upsert=true`, then existing entities are similarly patched with `result=updated`; otherwise, existing entities are not modified with `result=exists`.
 
 If more than one entity is referenced by the same external ID, an error is reported.
 
@@ -778,7 +778,7 @@ accept: application/json
 |----------------+----------------------------------------------------------|
 | `:id`          | The real ID                                              |
 | `:original_id` | Provided ID if different from real ID (ex: transient ID) |
-| `:result`      | `error`, `created` or `updated`                          |
+| `:result`      | `error`, `created`, `exists` or `updated`                |
 | `:external_id` | External ID used to identify the entity                  |
 | `:error`       | Error message                                            |
 
--- a/resources/ctia/public/doc/bulk-bundle.org
+++ b/resources/ctia/public/doc/bulk-bundle.org
@@ -667,7 +667,7 @@ When a bundle is submitted:
 
 1. All entities that have already been imported with the external ID whose prefix has been configured with the `ctia.store.external-key-prefixes` property are searched.
 2. If they are identified by transient IDs, a mapping table between transient and stored IDs is built.
-3. New entities are created in the same way as the `/bulk` API endpoint with transient IDs resolutions. If `upsert=true`, then existing entities are similarly patched with `result=updated`; otherwise, existing entities are not modified with `result=exists`.
+3. New entities are created in the same way as the `/bulk` API endpoint with transient IDs resolutions. If query parameter `patch-existing=true`, then existing entities are similarly patched with `result=updated`; otherwise, existing entities are not modified with `result=exists`.
 
 If more than one entity is referenced by the same external ID, an error is reported.
 
--- a/resources/ctia/public/doc/bulk-bundle.org
+++ b/resources/ctia/public/doc/bulk-bundle.org
@@ -667,7 +667,7 @@ When a bundle is submitted:
 
 1. All entities that have already been imported with the external ID whose prefix has been configured with the `ctia.store.external-key-prefixes` property are searched.
 2. If they are identified by transient IDs, a mapping table between transient and stored IDs is built.
-3. Only new entities are created in the same way as the `/bulk` API endpoint with transient IDs resolutions. Existing entities are not modified.
+3. New entities are created in the same way as the `/bulk` API endpoint with transient IDs resolutions. Existing entities are patched similarly.
 
 If more than one entity is referenced by the same external ID, an error is reported.
 
@@ -778,7 +778,7 @@ accept: application/json
 |----------------+----------------------------------------------------------|
 | `:id`          | The real ID                                              |
 | `:original_id` | Provided ID if different from real ID (ex: transient ID) |
-| `:result`      | `error`, `created` or `exists`                           |
+| `:result`      | `error`, `created` or `updated`                          |
 | `:external_id` | External ID used to identify the entity                  |
 | `:error`       | Error message                                            |
 
--- a/resources/ctia/public/doc/bulk-bundle.org
+++ b/resources/ctia/public/doc/bulk-bundle.org
@@ -677,6 +677,8 @@ as if by concatenating existing and new properties together in a single list,
 removing properties to the left of a property with the same name, 
 then sorting the list lexicographically by name before using this list to patch the existing entity.
 
+If `asset_properties-merge-strategy=ignore-existing`, then asset properties will be patched to their new values as they appear in the request bundle.
+
 Example for an incident along with its context, note the transient ids
 #+begin_src HTTP
 POST /ctia/bundle/import HTTP/1.1
--- a/resources/ctia/public/doc/bulk-bundle.org
+++ b/resources/ctia/public/doc/bulk-bundle.org
@@ -667,7 +667,7 @@ When a bundle is submitted:
 
 1. All entities that have already been imported with the external ID whose prefix has been configured with the `ctia.store.external-key-prefixes` property are searched.
 2. If they are identified by transient IDs, a mapping table between transient and stored IDs is built.
-3. New entities are created in the same way as the `/bulk` API endpoint with transient IDs resolutions. If query parameter `patch-existing=true`, then existing entities are similarly patched with `result=updated`; otherwise, existing entities are not modified with `result=exists`.
+3. New entities are created in the same way as the `/bulk` API endpoint with transient IDs and external ids resolutions. If query parameter `patch-existing=true`, then existing entities are similarly patched with `result=updated`; otherwise, existing entities are not modified with `result=exists`.
 
 If more than one entity is referenced by the same external ID, an error is reported.
 
--- a/src/ctia/bulk/core.clj
+++ b/src/ctia/bulk/core.clj
@@ -244,24 +244,28 @@
    auth-identity :- auth/AuthIdentity
    params
    services :- APIHandlerServices
-   {:keys [make-result]} :- {(s/optional-key :make-result) (s/maybe (s/=> (s/pred map?) flows/FlowMap))}]
+   {:keys [enveloped-result?]} :- {(s/optional-key :enveloped-result?) (s/maybe s/Bool)}]
   (when (seq patches)
     (let [get-fn #(read-entities %  entity-type auth-identity services)
           {:keys [realize-fn new-spec]} (get (all-entities) entity-type)]
-      (flows/patch-flow
-        :services services
-        :get-fn get-fn
-        :realize-fn realize-fn
-        :update-fn (update-fn entity-type auth-identity params services)
-        :long-id-fn #(with-long-id % services)
-        :entity-type entity-type
-        :identity auth-identity
-        :patch-operation :replace
-        :partial-entities patches
-        :tempids tempids
-        :spec new-spec
-        :make-result (or make-result make-bulk-result)
-        :get-success-entities (get-success-entities-fn :updated)))))
+      (cond-> (flows/patch-flow
+                :services services
+                :get-fn get-fn
+                :realize-fn realize-fn
+                :update-fn (update-fn entity-type auth-identity params services)
+                :long-id-fn #(with-long-id % services)
+                :entity-type entity-type
+                :identity auth-identity
+                :patch-operation :replace
+                :partial-entities patches
+                :tempids tempids
+                :spec new-spec
+                :make-result make-bulk-result
+                :enveloped-result? enveloped-result?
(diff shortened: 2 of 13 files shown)


ereteog (commented):
first pass.
Question: since we change the behaviour and the response, we might generate issues on the clients that rely on `exists` results. 
thoughts?

src/ctia/bundle/core.clj
```diff
@@ -75,30 +79,34 @@
...
-                                           (auth/ident->map auth-identity)
-                                           paging))
-          acc-entities (into entities results)
-          matched-ext-ids (into #{} (mapcat :external_ids results))
-          remaining-ext-ids (remove matched-ext-ids ext-ids)]
-      (if next-page
-        (recur remaining-ext-ids acc-entities)
-        acc-entities))))
+    (if (empty? ext-ids)
+      entities
```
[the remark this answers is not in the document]
frenchy64: Something about `all-pages` seems wrong based on my reading.

It checks `next-page` but doesn't use pagination and then recurs to the first page of a different query.

Say an external id is mapped to 2 entities with a page limit of 1. First, we miss the 2nd entity altogether since we recur with an empty `ext-ids`. And second, since `ext-ids` is empty, we make an empty query.

src/ctia/bulk/core.clj
```diff
@@ -113,7 +145,7 @@
                {action
                 (case action
                   :errors (format-bulk-flow-res res services)
-                  (map #(to-long-id % services) res))}))
+                  (map #(or (to-long-id % services) %) res))}))
```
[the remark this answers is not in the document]
frenchy64: I think this was from an earlier version where I didn't resolve transient ids early enough. I will try removing it.

src/ctia/bulk/core.clj
```diff
@@ -42,17 +42,47 @@
...
+  [(s/conditional
+     string? schemas/ID
+     :else {(s/optional-key :error) (s/conditional
+                                      string? s/Str
+                                      :else {:type (s/conditional
+                                                     string? s/Str
+                                                     :else s/Keyword)
+                                             :reason s/Str
+                                             (s/optional-key :index) s/Str
+                                             (s/optional-key :index_uuid) s/Str})
```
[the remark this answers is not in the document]
frenchy64: I'll investigate. I built this schema empirically via failing tests so I suspect there might be a leak.

src/ctia/bulk/core.clj
```diff
@@ -278,15 +332,20 @@
...
    :tempids TempIDs})
 
+(s/defschema BulkRefsAssocTempIDs
+  (st/assoc BulkRefs (s/optional-key :tempids) TempIDs))
+
 (s/defn import-bulks-with :- BulkRefs+TempIDs
   "Import each new-bulk in order while accumulating tempids."
-  [f :- (s/=> {s/Keyword {:data [s/Any]
+  [f :- (s/=> {s/Keyword {:data EnvelopedEntities+TempIDs
                           :tempids TempIDs}}
```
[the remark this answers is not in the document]
frenchy64: Yes, thanks, it's the same schema as the input of `merge-tempids`, which is actually checked so we know your suggestion is correct.

src/ctia/bulk/core.clj
```diff
@@ -196,24 +238,36 @@
...
+                :patch-operation :replace
+                :partial-entities patches
+                :tempids tempids
+                :spec new-spec
+                :make-result make-bulk-result
+                :enveloped-result? enveloped-result?
+                :get-success-entities (get-success-entities-fn :updated))
+        enveloped-result? (update :data #(mapv (fn [{:keys [error id] :as result}]
+                                                 (if error result id))
+                                               %))))))
```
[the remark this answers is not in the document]
frenchy64: > It seems that you tried several options here :-).

I introduced `enveloped-result?` in desperation :) I agree with your suggestion.

frenchy64:
@ereteog 

> Question: since we change the behaviour and the response, we might generate issues on the clients that rely on exists results.

Yes, also they might be relying on the the old behavior of not updating the entity. I can see this being an issue, a query param option for upsert seems like the obvious solution?

src/ctia/bundle/core.clj
```diff
@@ -162,14 +203,26 @@
...
+                          (when-not (schemas/transient-id? new-id)
+                            new-id))]
+    (if-some [old-entity (id->old-entity realized-id)]
+      (assoc entity-data :result "exists" :id (:id old-entity))
+      (cond-> entity-data
+        ;; if we want to support specifying ids on creation, start here
+        true ;;(not (auth/capable? auth-identity :specify-id))
+        (assoc :error {:type :unresolvable-id
+                       :reason (str "Long id must already correspond to an entity: " realized-id)}
+               :result "error")))
```
ereteog, line 223: you will have to combine this behaviour with the update entity option, right?
At some point we might need to split this function with the 2 branches of the `if-some` to keep it "easy" to read.

src/ctia/bundle/core.clj
```diff
@@ -200,28 +253,41 @@
...
   "Add existing entities to the import data map."
-  [import-data entity-type identity-map
-   services :- WithExistingEntitiesServices]
-  (let [entities-by-external-id
+  [import-data
+   entity-type
+   auth-identity :- auth/AuthIdentity
+   {{:keys [get-store]} :StoreService :as services} :- WithExistingEntitiesServices]
+  (let [{:keys [realized-entities unrealized-entities]} (group-by (fn [{{:keys [id]} :new-entity}]
+                                                                    (if (and id (not (schemas/transient-id? id)))
```
ereteog, line 261: there is a similar test in `with-existring-entity`, could we factorize it?

src/ctia/bundle/core.clj
```diff
@@ -230,42 +296,74 @@
...
-            (let [{submitted true
-                   not-submitted false} (group-by create? v)]
-              (concat
-               ;; Only submitted entities are processed
-               (map (fn [entity-import-data
-                         {:keys [error msg] :as entity-bulk-result}]
+   tempids :- TempIDs]
+  (reduce-kv (fn [acc k vs]
+               (reduce (fn [acc v]
+                         (let [op (when v
```
ereteog, line 323: In which case `v` is `nil`?

src/ctia/bundle/core.clj
```diff
@@ -288,27 +386,153 @@
...
+  [{{:keys [entity-enabled?]} :FeaturesService} :- APIHandlerServices]
+  (->> (entities/all-entities)
+       keys
+       (remove entity-enabled?)
+       (mapcat entity->bundle-keys)
+       (apply st/dissoc NewBundle)))
+
+(s/defschema AssetPropertiesProperties (st/get-in AssetProperties [:properties]))
+
+(s/defn merge-asset_properties-properties :- AssetPropertiesProperties
```
ereteog, line 409: that's a very custom interpretation of the model. I expected this merge to happen in IROH

test/ctia/bulk/core_test.clj
```diff
@@ -334,14 +336,14 @@
   (let [intermediate-tempids (atom [])
         incident1 (assoc incident-minimal :id "transientid1")
         incident2 (assoc incident-minimal :id "transientid2")]
-    (is (= {:bulk-refs {:incidents [incident1 incident2]}
+    (is (= {:bulk-refs {:incidents (mapv :id [incident1 incident2])}
```
ereteog, line 339: this a change of behaviour, did we have a bug?

test/ctia/bundle/core_asset_ref_entities_test.clj
```diff
@@ -46,34 +55,6 @@
       ;; in order to test association of Asset to AssetMappings/AssetProperties
       set-transient-asset-refs))
 
-(deftest bulk-for-asset-related-entities
```
ereteog, line 49: is it testing somewhere else now?

ereteog (commented):
second pass, I need more time for `ctia.bundle.core` (it became hard to read)

src/ctia/bundle/core.clj
frenchy64: Ok, I will add a query param for this behavior. It's too complicated and expensive to do this in IROH.

test/ctia/bundle/core_asset_ref_entities_test.clj
frenchy64: Yes, a more robust test that doesn't use implementation details was written for https://github.com/threatgrid/ctia/pull/1343

src/ctia/bundle/core.clj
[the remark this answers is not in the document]
frenchy64: Good to know, reverted.

test/ctia/bulk/core_test.clj
frenchy64: Mysterious, will circle back to this.

src/ctia/bundle/core.clj
frenchy64: I think this is old code, will remove.

src/ctia/bundle/core.clj
frenchy64: Added:
```
POST /bundle/import?asset_properties-merge-strategy=ignore-existing
- default behavior overwriting all properties
POST /bundle/import?asset_properties-merge-strategy=merge-overriding-previous
- latest property wins
```

test/ctia/bulk/core_test.clj
frenchy64: I changed the test a few lines down vvvv to satisfy a schema that was previously unenforced. This line ^^^^ also needed to change because the result changed.

src/ctia/bulk/core.clj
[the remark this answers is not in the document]
frenchy64: There's a test that intentionally closes an index, and these fields leak to the user. https://github.com/threatgrid/ctia/blob/59077597cabb4f6d982697eebec67fc68f057115/test/ctia/bundle/routes_test.clj#L391

```
{:type index_closed_exception, :reason closed, :index_uuid _nSHbVQcSFeU84cQWjyNig, :index ctia_indicator3c9866dd-bbb1-46fd-bcd9-9033dd947ff6-2023.10.04-000001}
```

src/ctia/bundle/core.clj
frenchy64: The update existing logic was moved to `prepare-bulk`, where I now return a separate result `:unsubmitted-result` containing both errors and existing (when appropriate).

src/ctia/bulk/core.clj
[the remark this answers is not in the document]
frenchy64: Opened https://github.com/advthreat/iroh/issues/8471

src/ctia/bulk/core.clj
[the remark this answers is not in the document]
frenchy64: I removed `enveloped-result?` from the flows namespace. It's still in the bundle/bulk namespace because it needs to change the result of `gen-bulk-from-fn`. I couldn't figure out a good way to completely remove it.

test/ctia/bundle/core_asset_ref_entities_test.clj
```diff
@@ -106,44 +87,61 @@
...
-              nil         ;; external-key-prefixes
-              login
-              services))))))))
+         (th/set-capabilities! app "foouser" ["foogroup"] "user" (all-capabilities))
+         (whoami-helpers/set-whoami-response app
+                                             "45c1f5e3f05d0"
+                                             "foouser"
+                                             "foogroup"
+                                             "user")
+         (let [create-response (th/POST app
```
ereteog, line 105: 🟠  why did you need to use the http layer (+ authorization) instead of the existing `bundle/import`?

test/ctia/bundle/routes_test.clj
```diff
@@ -133,10 +134,11 @@
                                   external_id)
                           :headers {"Authorization" "45c1f5e3f05d0"})
             [entity :as entities] (:parsed-body response)]
-        (is (= 1 (count entities))
-            "Only one entity is linked to the external ID")
-        (is (= id (:id entity))
-            "The submitted entity is linked to the external ID")))
+        (when (is (= 200 (:status response)))
```
ereteog, line 137: can we have other response code in this test? It seems that checking if the response is a 200 should be in an assert to validate that the test context is properly initialized.

resources/ctia/public/doc/bulk-bundle.org
```diff
@@ -667,10 +667,16 @@ When a bundle is submitted:
 
 1. All entities that have already been imported with the external ID whose prefix has been configured with the `ctia.store.external-key-prefixes` property are searched.
 2. If they are identified by transient IDs, a mapping table between transient and stored IDs is built.
-3. Only new entities are created in the same way as the `/bulk` API endpoint with transient IDs resolutions. Existing entities are not modified.
+3. New entities are created in the same way as the `/bulk` API endpoint with transient IDs resolutions. If query parameter `patch-existing=true`, then existing entities are similarly patched with `result=updated`; otherwise, existing entities are not modified with `result=exists`.
 
 If more than one entity is referenced by the same external ID, an error is reported.
 
+If provided query parameter `asset_properties-merge-strategy=merge-overriding-previous`, then when patching asset properties,
```
ereteog, line 674: can you also add a description for ` :ignore-existing`, please?

resources/ctia/public/doc/bulk-bundle.org
```diff
@@ -667,10 +667,16 @@ When a bundle is submitted:
...
+existing asset properties will be retrieved and combined with the asset properties in the request bundle
+as if by concatenating existing and new properties together in a single list,
+removing properties to the left of a property with the same name, 
+then sorting the list lexicographically by name before using this list to patch the existing entity.
```
ereteog, line 678:
```suggestion
then sorting the list lexicographically by name before using this list to patch the existing entity.
If `asset_properties-merge-strategy=ignore-existing`, then asset properties will be patched to their new values as they appear in the request bundle.
```

src/ctia/bundle/core.clj
```diff
@@ -109,14 +115,47 @@
...
 
+(s/defn find-by-asset_refs :- {s/Str (s/pred map?)}
+  "Returns a map from asset_ref to entity of entity-type that has
+  entry `:asset_ref asset_ref`."
+  [asset_refs :- #{s/Str}
+   entity-type :- (s/enum :asset-properties :asset-mapping)
+   auth-identity :- auth/AuthIdentity
+   {{:keys [get-store]} :StoreService} :- FindByExternalIdsServices]
+  (if (empty? asset_refs)
+    {}
```
ereteog, line 126: I guess that's now it's already included in the first test of the loop, right?:
``` clojure
           entities (loop [asset_refs asset_refs
                          entities {}]
                     (if (empty? asset_refs)
                       entities
```

ereteog (approved):
LGTM

That's a very sensitive PR, it would be nice to extend the QA section with the merge strategy for asset properties and dedicated tests to actively test the scenario behind this PR with score update.

test/ctia/bundle/core_asset_ref_entities_test.clj
frenchy64: I couldn't figure out how to make the test run on my machine.

src/ctia/bundle/core.clj
frenchy64: I just left it in to skip a useless log.

PR #1394, opened 2023-12-08
Eval routes and options given to `context` at initialization time

frenchy64:
Close https://github.com/advthreat/iroh/issues/6619

Compojure{-api}'s `context` wraps its body in a function. This causes anything inside the route to be evaluated on every call.

We have a very expensive call in the `POST /bundle/import` route that recreates the entire Bundle schema on every request. @devn found a request where this took 10% of the request time.

Additionally, since everything under "/ctia" is under a context, I believe many of the routes themselves are being recreated on every request.

To comprehensively prevent this issue, I've restricted our `context` macro to only allow options that don't bind variables. This way, we can evaluate the routes and the options separately and then close the `context` over the value of routes. This seems to substantially decrease the number of values reinitialized during a request.

CI seems three times faster after the change. Most of the speedup is from tests for individual entities, where typically each namespaces takes 2-3 minutes, and now takes 30 seconds. Compare these metrics, where slowest tests are first:

[Before (master)](https://github.com/threatgrid/ctia/actions/runs/7036030116/job/19148195186#step:9:17): 1 hour 28 minutes
[After](https://github.com/threatgrid/ctia/actions/runs/7173213423/job/19532288400#step:9:17): 40 minutes

I have not tested individual route performance.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: Stage initialization code under context
```

### DIFF
--- a/src/ctia/lib/compojure/api/core.clj
+++ b/src/ctia/lib/compojure/api/core.clj
@@ -77,8 +77,7 @@
                      options)]
        (core/context ~path ~arg
                      ~@(mapcat (fn [[k v]]
-                                 [k (or (option->g k)
-                                        v)])
+                                 [k (option->g k v)])
                                options)
                      routes#))))
 
--- a/src/ctia/lib/compojure/api/core.clj
+++ b/src/ctia/lib/compojure/api/core.clj
@@ -50,7 +50,8 @@
   [middleware & body]
   `(core/middleware ~middleware ~@body))
 
-(def ^:private allowed-context-options #{:tags :capabilities :description :return})
+(def ^:private allowed-context-options #{:tags :capabilities :description :return :summary})
+(def ^:private unevalated-options #{:tags})
 
 (defmacro context
   "Like compojure.api.core/context, except the binding vector must be empty and
@@ -66,13 +67,19 @@
             (throw (ex-info (str "Not allowed these options in `context`, push into HTTP verbs instead: "
                                  (pr-str (sort extra-keys)))
                             {})))
-        option->g (into {} (map (juxt identity (comp gensym name))) (keys options))]
+        option->g (into {} (comp (remove unevalated-options)
+                                 (map (juxt identity (comp gensym name))))
+                        (keys options))]
     `(let [routes# (core/routes ~@body)
            ~@(mapcat (fn [[k v]]
-                       [(option->g k) v])
+                       (when-some [g (option->g k)]
+                         [g v]))
                      options)]
        (core/context ~path ~arg
-                     ~@(mapcat (juxt identity option->g) (keys options))
+                     ~@(mapcat (fn [[k v]]
+                                 [k (or (option->g k)
+                                        v)])
+                               options)
                      routes#))))
 
 (defmacro GET     {:style/indent 2} [& args] `(core/GET ~@args))
--- a/src/ctia/lib/compojure/api/core.clj
+++ b/src/ctia/lib/compojure/api/core.clj
@@ -3,7 +3,9 @@
   
   Always use this namespace over compojure.api.{core,sweet}
   as it also loads the CTIA routing extensions."
-  (:require [ctia.http.middleware.auth :as magic]))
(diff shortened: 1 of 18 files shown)


brookeswanson (approved):
I'd warn UI folks before merging this, but this is a good find. Thank you @devn and @frenchy64

src/ctia/lib/compojure/api/core.clj
```diff
@@ -48,7 +50,36 @@
   [middleware & body]
   `(core/middleware ~middleware ~@body))
 
-(defmacro context {:style/indent 2} [& args] `(core/context ~@args))
+(def ^:private allowed-context-options #{:tags :capabilities :description :return :summary})
```
yogsototh, line 53: This is very difficult to really understand precisely what is going on. But I think the main problem comes from the auth middleware that set a `:lets`. By doing so, the `context` macro detects we have a `lets` and consider the `context` to be treated as `dynamic` (https://github.com/metosin/compojure-api/blob/master/src/compojure/api/meta.clj#L674) and this is AFAIU where the schema is potentially rebuilt.

Thus this put the `let` form just on top of the body and run for every request and not just at route init time.

So I am not sure this macro will be enough if you still allow the `:capabilities` param declaration at the context level.

yogsototh (commented):
I do not see anything wrong, and I might not be familiar enough with CTIA. 

That being said, from the problem description it looks like maybe the issue might be in `ctia.bundle.routes/bundle-routes`. 

The `(let [bundle-schema (prep-bundle-schema services) ,,,` is inside the form, I think this let should be put on top of route like:


```clojure
(let [bundle-schema (prep-bundle-schema services)]
  (routes 
    (context ,,,,
```

And I think this might generate the schema not for every request but only once during the route init.

frenchy64:
> That being said, from the problem description it looks like maybe the issue might be in ctia.bundle.routes/bundle-routes (here https://github.com/threatgrid/ctia/blob/master/src/ctia/bundle/routes.clj#L99):

@yogsototh I initially did that, then I realized that everything is under a context. The bundle routes are also under `/ctia`. The current compojure-api we use doesn't have the optimization you pointed out: https://github.com/metosin/compojure-api/blob/1.1.13/src/compojure/api/meta.clj#L310

src/ctia/lib/compojure/api/core.clj
frenchy64: > But I think the main problem comes from the auth middleware that set a :lets.

Yes, this approach is incompatible with any middleware that sets `:lets`. It throws a compile-time error. That's why things like `:auth-identity` and `:identity-map` are pushed into the HTTP verbs.

> By doing so, the context macro detects we have a lets and consider the context to be treated as dynamic (https://github.com/metosin/compojure-api/blob/master/src/compojure/api/meta.clj#L674) and this is AFAIU where the schema is potentially rebuilt.

Ah I see. We use compojure-api 1.1.13 which doesn't have this optimization. The difference between my macro and compojure-api master is that mine throws a compile-time error on a dynamic route. 

> So I am not sure this macro will be enough if you still allow the :capabilities param declaration at the context level.

FWIW here's the expansion:

```clojure
(macroexpand
  `(context "" []
     :capabilities #{:foo :bar}
     (GET "" req {:status 200})))
;=>
(let* [routes__207013__auto__ (compojure.api.core/routes (ctia.lib.compojure.api.core/GET "" ctia.lib.compojure.api.core/req {:status 200}))
       capabilities207088 #{:bar :foo}]
  (compojure.api.core/context
    "" []
    :capabilities capabilities207088
    routes__207013__auto__))
```

frenchy64: Perhaps even with the optimizations from compojure-api master, it would ok? This macro takes advantage of the fact that most arguments to `context` are expressions (as opposed to values like `:tags` and `:scopes`).

yogsototh (approved):
LGTM!

Thanks for the tests they are beautiful.

PR #1395, opened 2023-12-08
Never match existing asset-* entities when patch-existing=false

frenchy64:
This is the first of two fixes to this area of code.

There are two modes for `POST /bundle/import`: `patch-existing={true,false}`. If true, then existing entities are supposed to be detected and patched.

There is a bug where `asset_mappings` and `asset_properties` are erroneously mapped onto existing assets and essentially get merged when they should be created fresh.

This PR fixes this behavior for `patch-existing=false`, since no-one is using `patch-existing=true` yet and everyone is using `patch-existing=false`.

The merging code was made on the faulty assumption that there can only be one asset_mapping / asset_properties per asset.

There is a red test that is fixed by commit [Revert "Revert "green""](https://github.com/threatgrid/ctia/pull/1395/commits/79f4978455e8610795aaa675b2897f6b87b86872). The previous commit fails with the wrong number of results:

```
FAIL in ctia.bundle.routes-test/ (bundle-asset-relationships-test) (routes_test.clj:1504)
es-store relationships are created for asset mappings/properties
...
expected: (= 4 (count (filter (comp #{"created"} :result) create-results2)))
  actual: (not (= 4 2))

FAIL in ctia.bundle.routes-test/ (bundle-asset-relationships-test) (routes_test.clj:1506)
es-store relationships are created for asset mappings/properties
...
expected: (= 1 (count (filter (comp #{"exists"} :result) create-results2)))
  actual: (not (= 1 3))
```

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: Never match existing asset-* entities when patch-existing=false
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,4 +19,3 @@ data/**/*.edn
 .clj-kondo/cache
 .clj-kondo/rewrite-clj/
 tmp
-response_*.json
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,4 @@ data/**/*.edn
 .clj-kondo/cache
 .clj-kondo/rewrite-clj/
 tmp
+response_*.json
--- a/src/ctia/bundle/core.clj
+++ b/src/ctia/bundle/core.clj
@@ -115,10 +115,8 @@
              (all-pages entity-type external-ids auth-identity get-store))
       [])))
 
-(s/defschema ExistingWithAssetRef {s/Str [(s/pred map?)]})
-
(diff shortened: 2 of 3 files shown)


brookeswanson (approved):

marioaquino (approved):
Seems reasonable.  The `patch-existing` true case is covered by other tests.

frenchy64:
@marioaquino yes, but they were created with the wrong assumptions (see PR description). I will need to review them ASAP. This case is just more urgent since no-one is patching yet.

PR #1398, opened 2023-12-11
Use `st/merge` to merge schemas instead of `into`

frenchy64:
The crud routes namespace uses `into` to merge map schemas in the `DELETE /search` route.

Once fixed, the code becomes identical to the same logic in `ctia.entity.feed`, so an abstraction helps avoid code duplication.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: Use st/merge to merge schemas instead of into
```


ereteog (approved):
thank you (the feed is a lot of boiler plate)

PR #1406, opened 2023-12-21
Fix POST /bulk schema checking

frenchy64:
`:return` only checks 200 responses, so we miss validation on 201 and other responses. We have this problem in a lot of places, but in this particular case the was also schema was wrong, so I've extracted the fix into its own PR.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: Fix POST /bulk schema checking
```


ereteog (commented):
LGTM!

msprunck (approved):
LGTM!

ereteog (approved):

PR #1410, opened 2024-01-16
expose created and modified dates

ereteog:
> Related #advthreat/iroh/issues/8788

DO NOT MERGE, need to update ctia investigate module accordingly.

This PR expose created and modified date to (try to) stop the confusion on date filtering and date presentations in incidents.

The main change is to remove `un-store` fns from `src/ctia/domain/entities.clj` and references in routes that removed `created` and `modified` fields before returning documents. Then there is a test in the crud store helper to check that `modified` is properly updated on changes.
The other changes are fixing tests accordingly.

<a name="qa">[§](#qa)</a> QA
============================

1. Create an incident, check that the returned entity has `created` and `modified` fields set to the creation time. These dates should be different than the provided `timestamp` if any.
2. Retrieve this entity, check that the value of `created` and `modified` are available.
3. update the incident (patch the status for instance), and check that `created` did not change and `modified` was updated with the time of the change.
4. perform a search query that matches this incident and check that `created` and `modified` are returned in the result.
5. bundle export this incident and check that `created` and `modified` are returned

### DIFF
--- a/src/ctia/bulk/core.clj
+++ b/src/ctia/bulk/core.clj
@@ -5,7 +5,7 @@
    [clojure.tools.logging :as log]
    [ctia.auth :as auth]
    [ctia.bulk.schemas :refer [EntitiesResult]]
-   [ctia.domain.entities :as ent :refer [with-long-id short-id->long-id]]
+   [ctia.domain.entities :refer [with-long-id short-id->long-id]]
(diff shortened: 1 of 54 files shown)


frenchy64 (approved):
LGTM!

hisans:
QA Verified in Release 2.12.0 in TEST Environment.

PR #1417, opened 2024-03-25
bump ctim / remove status disposition

ereteog:
> Related #advthreat/iroh/issues/8181

It requires an ES surgery. The field was not indexed yet since everything was on hold
I will handle it myself.

<a name="qa">[§](#qa)</a> QA
============================

- check that you can use new statuses in `POST /ctia/incident/{incident-id}/status`:
New: Processing
New: Presented
Open: Investigating
Open: Reported
Open: Contained
Open: Recovered
Hold: Internal
Hold: External
Hold: Legal
Closed: Under Review
Closed: Confirmed Threat
Closed: Suspected
Closed: False Positive
Closed: Near-Miss
Closed: Other

- check that you cannot use the field "status_disposition" for incidents.


marioaquino (approved):
LGTM

PR #1419, opened 2024-04-05
remove ES5 support

ereteog:
ES5 EOL was reached a long time ago and all tests should happen on the last version ES7 (potentially ES8 in a close future)
I also want to try some ES7 features that would require to find a backward compatibility with ES5 that might be impossible (and not useful to maintain).

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.


test/ctia/task/migration/migrate_es_stores_test.clj
```diff
@@ -577,7 +577,13 @@
...
         {:keys [all-stores]} (helpers/get-service-map app :StoreService)
         services (app->MigrationStoreServices app)
-        conn (es-conn get-in-config)]
+        conn (es-conn get-in-config)
+        migrated-store-keys [:incident
+                             :asset
+                             :attack-pattern
+                             :judgement
+                             :sighting
+                             :malware]]
```
ereteog, line 586: migrate less indices (still validate the process), that's workaround for introduced shard limit on ES7
https://www.elastic.co/guide/en/elasticsearch/reference/7.17/breaking-changes-7.0.html#_cluster_wide_shard_soft_limit

frenchy64 (approved):
Congratulations! What a ride.

PR #1421, opened 2024-04-08
Properly filter Relationships to assemble a Feed View

gbuisson:
> **Epic** #
> Close #
> Related https://github.com/advthreat/iroh/issues/9111

Here are the steps to render a Feed in CTIA:

1. Fetch the feed document
2. Fetch the referenced indicator
3. Fetch all the relationships of the indicator
4. Fetch all the objects referenced by the `source_ref` field of the relationships
5. extract the observables from the Judgements and render the response  

The query on step 3 is too naive as it attempts to list any kind of relationship even though we're looking only for indicators with an "element-of" relationship_type, some indicators are linked to a very large number of Sightings, retrieving them consumes a huge amount of resources and ultimately leads to a failure.

This PR updates the filter so we only fetch the proper indicators relevant for a Feed.

<a name="qa">[§](#qa)</a> QA
============================

1. Create and Query a Feed
2. The functionality should still work as expected and there should be no regression

```
public: Feed performance improvements
```

### DIFF
--- a/src/ctia/entity/feed.clj
+++ b/src/ctia/entity/feed.clj
@@ -169,7 +169,8 @@
              (get-paginated-response
               (store/paginate
                relationship-store
-               #(store/list-records %1 {:all-of {:target_ref indicator_id}} feed-identity %2)
+               #(store/list-records %1 {:all-of {:relationship_type "element-of"
+                                                 :target_ref indicator_id}} feed-identity %2)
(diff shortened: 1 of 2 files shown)


msprunck (approved):
LGTM!

NeelSaini:
Verified on TEST in rel v2.17.0

PR #1439, opened 2024-07-25
Change regex pattern to find ambiguous index name

ryemelia:
we need to remove from match indices which ends on `000001` (can be any number in the same format)

because we have indices created by elastic ILM policy and rollover step which looks like this:
`partial-v2.0.0_ctia_identities-2024.07.11-000001`
and its valid index name 

current regex should match indices which ends on date but ignore if ends on `-6 dights`


ryemelia:
wrong PR 

should be like 

index starting with `partial-<index-name>`
index starting with `restored-<index-name>`

^ valid

PR #1440, opened 2024-08-21
Xdr 6078

ereteog:
> Related #XDR-6078

This PR modifies the es store init by replacing legacy templates by new templates.
It automatically init a basic index lifecycle using current rollover configuration.
It also introduces a migration fn that is triggered by update-index-state task when `migrate-to-ilm` is set to true.
it removes legacy tasks that are not used anymore (settings.clj and update-mapping.clj), since the introduction of update-index-state.

The tests anticipate the deplyment scenario:
- merge this PR and the application restarts on legacy indices: nothing should change
- we trigger the update index task, policies and index-templates are created, legacy templates are deleted, indices settings (lifecycle) and aliases are updated.
- lifecycle is operational (rollover is properly and automatically triggered)

<a name="qa">[§](#qa)</a> QA
============================

Check that crud and bundle requests still properly work.

<a name="qa">[§](#qa)</a> OPS
============================

Trigger the `update-index-state` task.


containers/dev/docker-compose.yml
```diff
@@ -8,6 +7,7 @@ services:
     image: elasticsearch:7.9.2
     environment:
       - cluster.name=elasticsearch7
+      - indices.lifecycle.poll_interval=1s
```
ereteog, line 10: for test purpose

msprunck (approved):
LGTM!

hisans:
QA verified in Release 2.28.0 in TEST Environment.

PR #1443, opened 2024-09-18
switch lazyseq to iteration

ereteog:
> **Epic** #
> Close #
> Related #

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
```

### DIFF
--- a/src/ctia/task/migration/migrate_es_stores.clj
+++ b/src/ctia/task/migration/migrate_es_stores.clj
@@ -124,12 +124,14 @@
                :documents data
                :search_after next-search-after)))))
 
-(s/defn read-source ;; WARNING: defining schema output breaks lazyness
+(s/defn read-source
   "returns a lazy-seq of batch from source store"
   [read-params :- (s/maybe BatchParams)]
-  (lazy-seq
-   (when-let [batch (read-source-batch read-params)]
-     (cons batch (read-source (dissoc batch :documents))))))
+  (iteration read-source-batch
+             :vf identity
+             :initk read-params
+             :kf #(dissoc % :documents)
+             :somef seq))
 
 (s/defn write-target :- s/Int
   "This function writes a batch of documents which are (1) modified with `migrations` functions,
--- a/src/ctia/task/migration/store.clj
+++ b/src/ctia/task/migration/store.clj
@@ -430,7 +430,8 @@ Rollover requires refresh so we cannot just call ES with condition since refresh
                       {"id" sort-order})
         params
         (merge
-         {:offset (or offset 0)
+         {:track_total_hits true
+          :offset (or offset 0)
           :limit batch-size}
          (when sort-order
            {:sort sort-by})


src/ctia/task/migration/migrate_es_stores.clj
```diff
@@ -124,12 +124,14 @@
...
   "returns a lazy-seq of batch from source store"
   [read-params :- (s/maybe BatchParams)]
-  (lazy-seq
-   (when-let [batch (read-source-batch read-params)]
-     (cons batch (read-source (dissoc batch :documents))))))
+  (iteration read-source-batch
+             :vf identity
+             :initk read-params
+             :kf #(dissoc % :documents)
+             :somef seq))
```
frenchy64, line 134: I think both `:vf` and `:somef` should be their default values. 

`:somef` => `some?` because it's testing a `(s/maybe BatchParams)`

`:vf` is already `identity`.

frenchy64 (commented):
Does this have any memory implications? This gets passed to `seque` which seems to behave the same based on my reading.

frenchy64:
I'll look into this, but I worry that `seque` recalculates elements of the seq. Since iteration is uncached, this might result in multiple calls.

The `& xs` here calls `next`, but then it's thrown away and maybe recalculated on the next fill.

https://github.com/clojure/clojure/blob/ce55092f2b2f5481d25cff6205470c1335760ef6/src/clj/clojure/core.clj#L5447

ereteog:
> Does this have any memory implications? This gets passed to `seque` which seems to behave the same based on my reading.

I tried to generate a memory problem and it didn't trigger any. 
However a very similar approach did generate a memory issue in IROH for reports and it actually appeared to have a memory leak issue during last migration.
I did not rework on it since last time since I could not reproduce the memory issue, but it still bother me :-).

frenchy64:
I think they are similarly lazy, I'm guessing `iteration` is slightly leaner since it doesn't need to cache its own seq.

I ended up finding a bug which might be relevant: seque forces n+2 elements of the producing seq.

https://ask.clojure.org/index.php/14178/seque-forces-n-2-items-ahead-of-consumer-instead-of-n


```clojure
(let [producer (fn [i]
                 (prn "producer" i)
                 (inc i))
      s (seque 1 (iteration get-page! :initk 0))]
  (Thread/sleep 1000))
;"id" 0
;"id" 1
;"id" 2
nil

(let [producer (fn step [i]
                 (lazy-seq
                   (prn "producer" i)
                   (cons i (step (inc i)))))
      s (seque 1 (producer 0))]
  (Thread/sleep 1000)
  nil)
;="producer" 0
;="producer" 1
;="producer" 2
nil
```

frenchy64:
In addition, `seque` has a memory leak where it will hold onto the the n+1 and n+2 elements even after the `seque` is garbage collected. This is because it sends off an agent which has a known memory leak since it coveys itself in an `*agent*` thread binding https://clojure.atlassian.net/browse/CLJ-2619

ereteog:
> In addition, `seque` has a memory leak where it will hold onto the the n+1 and n+2 elements even after the `seque` is garbage collected. 

That would explain a lot! 
I could not reproduce the issue with a similar test I did for another implementation that [I rewrote with iteration](https://github.com/advthreat/iroh/pull/8957). Now I think that I know what is the main difference, thx!
The `seque` is in [migrate-query](https://github.com/threatgrid/ctia/blob/592a2f55c55efec7548dbf255cf6c55b651360fa/src/ctia/task/migration/migrate_es_stores.clj#L175) that is called with a reduce [here](https://github.com/threatgrid/ctia/blob/592a2f55c55efec7548dbf255cf6c55b651360fa/src/ctia/task/migration/migrate_es_stores.clj#L236).
What you said would explain what happened last time for events store, which migration showed a memory leak and have to be restarted several time (hopefully the process can restart where it stopped :-) ). The migration is splitted per week because we need to sort and sorting on weeks is fast while sorting the whole index is slow (of course :-) ).
Since we migrated about 5 years of ~8 billion documents for events, it means that ~250 seque were not garbage collected :-)!

frenchy64 (approved):
Nice!

frenchy64:
Reported memory leak: https://ask.clojure.org/index.php/14185/memory-leak-in-seque-via-agents

ereteog:
we have 2 reduce
on each query (1 week of data) we have a seque on the lazyseq in migrate-query
```
 data-queue (seque buffer-size
                          (read-source read-params))
new-migrated-count (reduce #(write-target %1 %2 services)
                                   migrated-count
                                   data-queue)]
```   
and we reduce over queries:
```
(reduce #(migrate-query %1 %2 services)
                 base-params
                 queries)                      
```

In any case that's worth migrating to an idiomatic abstraction.

frenchy64:
> In any case that's worth migrating to an idiomatic abstraction, right?

Yes, I prefer reducing over an iteration. It will at least reduce the possible explanations if we continue to see memory issues.

PR #1451, opened 2024-10-25
Enable "partial-" and "restored-" prefixes

ereteog:
> Related https://cisco-sbg.atlassian.net/browse/XDR-13581

This PR exclude "restored-" and "partial-" prefixes from the ambiguous index names detection.

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

### DIFF
--- a/src/ctia/stores/es/init.clj
+++ b/src/ctia/stores/es/init.clj
@@ -155,15 +155,22 @@
                         :mappings mappings))
       (system-exit-error))))
 
+(defn ambiguous-indices
+  [existing index]
+  (let [existing-k (set (keys existing))
+        index-pattern (re-pattern (str "((partial|restored)-)?"
+                                       index
+                                       "(-\\d{4}.\\d{2}.\\d{2}.*)?"))
+        matching (filter #(re-matches index-pattern (name %))
+                         existing-k)
+        ambiguous (difference existing-k (set matching))]
+    ambiguous))
+
 (defn get-existing-indices
   [conn index]
   ;; retrieve existing indices using wildcard to identify ambiguous index names
   (let [existing (index/get conn (str index "*"))
-        existing-k (set (keys existing))
-        index-pattern (re-pattern (str index "(-\\d{4}.\\d{2}.\\d{2}.*)?"))
-        matching (filter #(re-matches index-pattern (name %))
-                         existing-k)
-        ambiguous (difference existing-k (set matching))]
+        ambiguous (ambiguous-indices existing index)]
     (if (seq ambiguous)
       (do (log/warn (format "Ambiguous index names. Index: %s, ambiguous: %s."
                             (pr-str index)
--- a/src/ctia/stores/es/init.clj
+++ b/src/ctia/stores/es/init.clj
@@ -157,14 +157,12 @@
 
 (defn ambiguous-indices
   [existing index]
-  (let [existing-k (set (keys existing))
-        index-pattern (re-pattern (str "((partial|restored)-)?"
-                                       index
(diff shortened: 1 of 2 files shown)


src/ctia/stores/es/init.clj
```diff
@@ -155,15 +155,22 @@
                         :mappings mappings))
       (system-exit-error))))
 
+(defn ambiguous-indices
+  [existing index]
+  (let [existing-k (set (keys existing))
+        index-pattern (re-pattern (str "((partial|restored)-)?"
+                                       index
```
frenchy64, line 162: Might want to quote this string: `(Pattern/quote index)`

https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html#quote-java.lang.String-

src/ctia/stores/es/init.clj
```diff
@@ -155,15 +155,22 @@
...
+                                       "(-\\d{4}.\\d{2}.\\d{2}.*)?"))
+        matching (filter #(re-matches index-pattern (name %))
+                         existing-k)
+        ambiguous (difference existing-k (set matching))]
```
frenchy64, line 166: Might be more direct to build it up if I'm reading it right.

```clojure
(into #{} (remove #(re-matches index-pattern (name %)))
      (keys existing))
```

test/ctia/stores/es/init_test.clj
```diff
@@ -175,6 +175,23 @@
...
+                     :existing {:v2.0.0_ctia_sighting-2022.07.07-000001 {}
+                                :partial-v2.0.0_ctia_sighting-2022.07.07-000001 {}
+                                :restored-v2.0.0_ctia_sighting-2022.07.07-000001 {}}
+                     :index "v2.0.0_ctia_sighting"
+                     :expected #{}}
+                    {:message "ambiguous names with unknown prefix should be detected"
+                     :existing {:v2.0.0_ctia_sighting-2022.07.07-000001 {}
+                                :prefix-v2.0.0_ctia_sighting-2022.07.07-000001 {}}
+                     :index "v2.0.0_ctia_sighting"
+                     :expected #{:prefix-v2.0.0_ctia_sighting-2022.07.07-000001}}]]
```
frenchy64, line 189: Can you add a test case that would fail if you don't regex-quote the index?

`.` is probably a regex character, maybe try this:

```clojure
                    {:message "ambiguous names with unknown prefix should be detected"
                     :existing {:v2x0x0_ctia_sighting-2022.07.07-000001 {}
                                :prefix-v2x0x0_ctia_sighting-2022.07.07-000001 {}}
                     :index "v2.0.0_ctia_sighting"
                     :expected #{:prefix-v2.0.0_ctia_sighting-2022.07.07-000001}}
```

frenchy64 (changes requested):
I like the tests, I think I found a potential issue.

src/ctia/stores/es/init.clj
```diff
@@ -155,15 +155,22 @@
                         :mappings mappings))
       (system-exit-error))))
 
+(defn ambiguous-indices
+  [existing index]
+  (let [existing-k (set (keys existing))
+        index-pattern (re-pattern (str "((partial|restored)-)?"
+                                       index
+                                       "(-\\d{4}.\\d{2}.\\d{2}.*)?"))
```
frenchy64, line 163: I think these `.` should be `[.]`.

ereteog:
@frenchy64 good catch!
I addressed your comment, wdyt?

frenchy64 (approved):
Satisfies my pedantry! Thanks!

PR #1468, opened 2025-02-10
XDR-19630 XDR-19632 Fix interval logic for new statuses.

samwagg:
This is an amendment to a previous change. In my previous PR I had missed that the logic for recording the new status changes in `:incident_time` was also needed (and necessary for interval computation).

See the main PR for details: https://github.com/threatgrid/ctia/pull/1467

I also took the opportunity to make a change suggested by @ereteog in the previous PR to use regex's instead of sets to recognize 'New', 'Open', and 'Closed' statuses.

### DIFF
--- a/src/ctia/entity/incident.clj
+++ b/src/ctia/entity/incident.clj
@@ -113,21 +113,31 @@
 (s/defschema IncidentStatusUpdate
   {:status IncidentStatus})
 
+(defn new-status? [status]
+   (some? (re-matches #"New(: .+)?" status)))
+
+(defn open-status? [status]
+   (some? (re-matches #"Open(: .+)?" status)))
+
+(defn closed-status? [status]
+  (some? (re-matches #"Closed(: .+)?" status)))
+
 (defn make-status-update
   [{:keys [status]}]
   (let [t (time/internal-now)
-        verb (case status
-               "New" nil
-               "Stalled" nil
-               "Hold" nil
-               ;; Note: GitHub syntax highlighting doesn't like lists with strings
-               "Containment Achieved" :remediated
-               "Restoration Achieved" :remediated
-               "Open" :opened
-               "Rejected" :rejected
-               "Closed" :closed
-               "Incident Reported" :reported
-               nil)]
+        verb (or (case status
+                   "New" nil
+                   "Stalled" nil
+                   "Hold" nil
+                   ;; Note: GitHub syntax highlighting doesn't like lists with strings
+                   "Containment Achieved" :remediated
+                   "Restoration Achieved" :remediated
+                   "Rejected" :rejected
+                   "Incident Reported" :reported
+                   nil)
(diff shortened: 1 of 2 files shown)


src/ctia/entity/incident.clj
```diff
@@ -113,21 +113,31 @@
...
+                   "Hold" nil
+                   ;; Note: GitHub syntax highlighting doesn't like lists with strings
+                   "Containment Achieved" :remediated
+                   "Restoration Achieved" :remediated
+                   "Rejected" :rejected
+                   "Incident Reported" :reported
+                   nil)
+                 (cond
+                   (open-status? status) :opened
+                   (closed-status? status) :closed))]
```
samwagg, line 140: Are any additional changes needed here to account for new incident statuses? I wasn't sure. I only made the changes necessary to support interval computation.

test/ctia/entity/incident_test.clj
```diff
@@ -779,102 +780,108 @@
...
                    ;; have value <expected-average> between time window <from> and <to> (latter is optional).
-                   (doseq [[field expected-count expected-average from to :as test-case]
-                           [["new_to_opened" 3 (avg first-new_to_opened second-new_to_opened third-new_to_opened) first-created]
-                            ["new_to_opened" 2 (avg second-new_to_opened third-new_to_opened) second-created]
-                            ["new_to_opened" 2 (avg first-new_to_opened second-new_to_opened) first-created (inc second-created)]
-                            ["new_to_opened" 1 first-new_to_opened first-created (inc first-created)]
-                            ["new_to_opened" 1 second-new_to_opened second-created (inc second-created)]
-                            ["new_to_opened" 1 third-new_to_opened third-created (inc third-created)]
-                            ["new_to_opened" 1 third-new_to_opened third-created]
-                            ["new_to_opened" 0 nil (inc third-created)]
```
frenchy64, line 860: Are the `(nil? expected-avg)` cases still covered?

frenchy64 (approved):

src/ctia/entity/incident.clj
frenchy64: I think this looks good.

test/ctia/entity/incident_test.clj
samwagg: Good catch. I think I got a little carried away with refactoring that test. I opted to revert and do a more minimal modification to exercise these new cases.

samwagg:
@frenchy64 Since the only modification was to test code I went ahead and merged so I can test in INT. But please review and let me know if you see any issues.

frenchy64:
@samwagg looks good!

src/ctia/entity/incident.clj
ereteog: that should be all in CTIA.
IROH changes may still be necessary, reach out to me offline please.

PR #1514, opened 2026-03-18
return 413 when entities exceed collection size limits

ereteog:
> **Epic** [XDR-46972](https://cisco-sbg.atlassian.net/browse/XDR-46972)
> Related [ctim#483](https://github.com/threatgrid/ctim/pull/483)
> Related [flanders#58](https://github.com/threatgrid/flanders/pull/58)

- update ctim to 1.3.30-SNAPSHOT, flanders to 1.1.1-SNAPSHOT
- check-spec detects collection limit violations via pred metadata
- add entity-too-large-error handler returning 413 with field details
- add check-spec-test covering valid, invalid, and limit scenarios

<a name="qa">[§](#qa)</a> QA
============================

No QA is needed.

```
intern: Return 413 when entities exceed collection size limits defined in CTIM.
```

[XDR-46972]: https://cisco-sbg.atlassian.net/browse/XDR-46972?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -89,8 +89,8 @@
                  ;; Schemas
                  [prismatic/schema "1.4.1"]
                  [metosin/schema-tools "0.13.1"]
-                 [threatgrid/flanders "1.1.0"]
-                 [threatgrid/ctim "1.3.30"
+                 [threatgrid/flanders "1.1.1-SNAPSHOT"]
+                 [threatgrid/ctim "1.3.30-SNAPSHOT"
                   :exclusions [com.cognitect/transit-java]] ;; ring-middleware-format takes precedence
                  [instaparse "1.4.10"] ;; com.gfredericks/test.chuck > threatgrid/ctim
                  [threatgrid/clj-momo "0.4.1"]
--- a/src/ctia/flows/crud.clj
+++ b/src/ctia/flows/crud.clj
@@ -99,17 +99,40 @@
       (id/str->short-id id)
       id)))
 
+(defn- size-limit-violation?
+  "Returns true if any of the spec problems are caused by a max-len predicate."
+  [explain-data]
+  (some (fn [{:keys [pred]}]
+          (when (fn? pred)
+            (:max-len (meta pred))))
+        (:clojure.spec.alpha/problems explain-data)))
+
+(defn- describe-size-violations
+  "Builds a human-readable message listing which fields exceeded their size limits."
+  [explain-data]
+  (->> (:clojure.spec.alpha/problems explain-data)
+       (keep (fn [{:keys [pred val in]}]
+               (when-let [max-len (and (fn? pred) (:max-len (meta pred)))]
+                 (format "field %s has %d elements (max %d)"
+                         (pr-str in) (count val) max-len))))
+       (str/join "; ")))
+
 (defn- check-spec [entity spec]
-  (if (and spec
-           (not (cs/valid? spec
-                           ;; the spec enforce long id for the API
-                           ;; while we store short ids
-                           (dissoc entity :id))))
-    {:msg (cs/explain-str spec entity)
-     :error "Entity validation Error"
-     :type :spec-validation-error
-     :entity entity}
-    entity))
+  (if-not spec
+    entity
+    (let [entity-no-id (dissoc entity :id)
+          explain-data (cs/explain-data spec entity-no-id)]
+      (if-not explain-data
+        entity
+        (if (size-limit-violation? explain-data)
+          {:msg (str "Entity too large: " (describe-size-violations explain-data))
+           :error "Entity too large"
+           :type :entity-too-large-error
+           :entity entity}
+          {:msg (cs/explain-str spec entity)
+           :error "Entity validation Error"
+           :type :spec-validation-error
+           :entity entity})))))
 
 (defn tlp-check
   [{:keys [tlp] :as entity} get-in-config]
--- a/src/ctia/http/exceptions.clj
+++ b/src/ctia/http/exceptions.clj
@@ -7,7 +7,8 @@
             [compojure.api.impl.logging :as logging]
             [ring.util.http-response :refer [internal-server-error
                                              bad-request
-                                             forbidden]]
+                                             forbidden
+                                             request-entity-too-large]]
             [clojure.data.json :as json]))
 
 (defn ex-message-and-data [^Exception e]
@@ -80,6 +81,18 @@
    (let [entity (:entity (ex-data e))]
      {:error "Spec Validation Client Error"
       :type "Invalid Entity Error"
+      :message (.getMessage e)
+      :entity entity
+      :class (.getName (class e))})))
+
+(defn entity-too-large-error-handler
+  "Handle entity too large error (413)"
+  [^Exception e _data _request]
+  (logging/log! :info e (ex-message-and-data e))
+  (request-entity-too-large
+   (let [entity (:entity (ex-data e))]
+     {:error "Entity Too Large"
+      :type "Entity Size Limit Error"
       :message (.getMessage e)
       :entity entity
       :class (.getName (class e))})))
--- a/test/ctia/flows/crud_test.clj
+++ b/test/ctia/flows/crud_test.clj
@@ -10,7 +10,9 @@
             [ctim.examples.sightings :refer [sighting-minimal]]
             [ctia.domain.entities :refer [short-id->long-id]]
             [java-time.api :as jt]
-            [puppetlabs.trapperkeeper.app :as app]))
+            [puppetlabs.trapperkeeper.app :as app]
+            [ctim.schemas.common :as ctim-common]
+            [ctim.schemas.sighting :as sighting-schema]))
 
 (deftest deep-merge-with-add-colls-test
   (let [fixture {:foo {:bar ["one" "two" "three"]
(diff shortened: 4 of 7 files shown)


yogsototh (commented):
**Verdict: NEEDS DISCUSSION** -- 0 CRITICAL, 0 HIGH (verified in-diff), 2 merge gates + 2 design questions.

The diff itself is small, focused, and idiomatic, and the dispatch wiring is correct end-to-end: the thrown `:type :entity-too-large-error` (`crud.clj:130`) matches the handler key (`handler.clj:116`) exactly, `request-entity-too-large` follows the same `ring.util.http-response` pattern as the working `bad-request`/`forbidden` referred alongside it, and `(.getMessage e)` carries the human-readable size message to the 413 body. No correctness/security/contract bug is provable from the code in this PR alone.

The blockers are not in the diff -- they are external dependencies and unverifiable behavior:

- **Merge gate -- SNAPSHOT dependencies on a master-targeting branch.** `project.clj`, `dependabot/pom.xml`, and `dependabot/dependency-tree.txt` move to `ctim 1.3.30-SNAPSHOT` / `flanders 1.1.1-SNAPSHOT` (mutable, unreleased). The entire feature depends on `:max-len` predicate metadata that exists only in those unreleased artifacts (Related: ctim#483, flanders#58). A master build is not reproducible until these are pinned to released versions. Flagged inline at `project.clj:93`.

- **Cannot verify the core detection mechanism here.** `size-limit-violation?` keys on `(fn? pred)` + `(:max-len (meta pred))` of a clojure.spec problem's `:pred`. In vanilla clojure.spec, `:pred` is the predicate *form* (not a fn object), so this only fires if the SNAPSHOT ctim/flanders register specs with a real fn-with-`:max-len`-metadata as the predicate. I could not confirm that against released artifacts; the new `check-spec-test` asserts it, so correctness hinges on that test passing in CI against the SNAPSHOT. Please confirm CI is green against the SNAPSHOT before merge. Detail inline at `crud.clj:105`.

### Design questions for the author (inline)

- If the SNAPSHOT attaches `:max-len` to the *shared* `max-len` predicate that ctim also uses for string-length bounds (ShortString/LongString/Markdown/Reference/URI), an over-long *string* field would be misclassified as 413 instead of 400, and the message would call characters "elements". Worth scoping detection to collection-count bounds. (`crud.clj:105`, `crud.clj:116`)
- `some` means an entity that is both oversized *and* otherwise invalid is reported only as 413, hiding the non-size error the old path surfaced. (`crud.clj:127`)

### PASS: Code Quality, Coding Style (minor nits inline only)

Also left 7 inline comments.

<details><summary>Non-blocking follow-ups</summary>

- No end-to-end test asserts an actual HTTP 413 status (only the private `check-spec` unit is tested); `entity-too-large-error-handler` is untested. The repo already has the pattern (`test_helpers/crud.clj` issues real requests asserting 400). (`crud_test.clj:310`)
- Bulk/bundle imports use `enveloped-result?` true, so oversized entities return inline per-entity errors inside a 201 rather than a 413 -- consistent with how bulk handles spec-validation today, but confirm that is intended.
- `:class (.getName (class e))` leaks the exception class to clients; matches existing sibling handlers, so out of scope here, but worth a future cleanup across all handlers.

</details>

project.clj
```diff
@@ -89,8 +89,8 @@
                  ;; Schemas
                  [prismatic/schema "1.4.1"]
                  [metosin/schema-tools "0.13.1"]
-                 [threatgrid/flanders "1.1.0"]
-                 [threatgrid/ctim "1.3.30"
+                 [threatgrid/flanders "1.1.1-SNAPSHOT"]
+                 [threatgrid/ctim "1.3.30-SNAPSHOT"
```
yogsototh, line 93: **Merge gate -- SNAPSHOT dependencies.** `ctim 1.3.30-SNAPSHOT` and `flanders 1.1.1-SNAPSHOT` (line 92) are mutable, unreleased artifacts. Merging them to master makes the build non-reproducible: the bytes resolved by a future build are not pinned and can change without any commit here. The feature also functionally depends on `:max-len` metadata that exists only in these unreleased versions (Related: ctim#483, flanders#58). Please pin to released `1.3.30` / `1.1.1` (or hold this PR until those release) before merge. Same change in `dependabot/pom.xml` and `dependabot/dependency-tree.txt`.

src/ctia/flows/crud.clj
```diff
@@ -99,17 +99,40 @@
       (id/str->short-id id)
       id)))
 
+(defn- size-limit-violation?
+  "Returns true if any of the spec problems are caused by a max-len predicate."
+  [explain-data]
+  (some (fn [{:keys [pred]}]
```
yogsototh, line 105: **Verify against the SNAPSHOT: does `:pred` carry `:max-len` metadata as a fn object?** This detection requires the clojure.spec problem's `:pred` to satisfy `(fn? pred)` *and* carry `:max-len` in its metadata. In vanilla clojure.spec, `:pred` in an `explain-data` problem is the predicate *form* (a list/symbol -- not `fn?`), so this branch would silently never fire and oversized entities would fall through to the 400 `:spec-validation-error` path. It only works if the SNAPSHOT ctim/flanders register specs with a real metadata-bearing fn as the predicate. The new `check-spec-test` asserts the 413 path, so please confirm CI passes against the SNAPSHOT -- the whole feature hinges on this single fact. Worth a code comment documenting the dependency contract so a future ctim/flanders refactor that changes the `:pred` shape doesn't silently disable this.

src/ctia/flows/crud.clj
```diff
@@ -99,17 +99,40 @@
...
+            (:max-len (meta pred))))
+        (:clojure.spec.alpha/problems explain-data)))
+
+(defn- describe-size-violations
+  "Builds a human-readable message listing which fields exceeded their size limits."
+  [explain-data]
+  (->> (:clojure.spec.alpha/problems explain-data)
+       (keep (fn [{:keys [pred val in]}]
+               (when-let [max-len (and (fn? pred) (:max-len (meta pred)))]
+                 (format "field %s has %d elements (max %d)"
```
yogsototh, line 116: **Message wording + field name.** Two issues: (1) If `:max-len` also tags ctim's string-length predicates (ShortString/LongString/Markdown/Reference/URI all use a shared `max-len`), an over-long *string* would format as `field ... has N elements (max M)` -- calling a character count "elements". (2) `(pr-str in)` emits the raw spec path vector (e.g. `[:observables]`, or `[:external_references 3 :url]`) as the client-facing field name. Consider `(str/join "." (map name in))` for a friendlier label, and either scope detection to collection bounds or generalize the "elements" wording.

src/ctia/flows/crud.clj
```diff
@@ -99,17 +99,40 @@
...
-     :type :spec-validation-error
-     :entity entity}
-    entity))
+  (if-not spec
+    entity
+    (let [entity-no-id (dissoc entity :id)
+          explain-data (cs/explain-data spec entity-no-id)]
+      (if-not explain-data
+        entity
+        (if (size-limit-violation? explain-data)
```
yogsototh, line 127: **`some` masks co-occurring non-size errors.** Because `size-limit-violation?` returns truthy if *any* problem is a max-len predicate, an entity that is both oversized *and* otherwise invalid (e.g. bad `:confidence`) is reported only as `entity-too-large` (413); the non-size error that the old `:spec-validation-error` path surfaced via `explain-str` is dropped. The client fixes the size, resubmits, then hits a fresh 400 for a field they were never told about. Consider reporting all problems, or document that size takes precedence.

src/ctia/flows/crud.clj
```diff
@@ -99,17 +99,40 @@
       (id/str->short-id id)
       id)))
 
+(defn- size-limit-violation?
+  "Returns true if any of the spec problems are caused by a max-len predicate."
```
yogsototh, line 103: **NIT -- docstring + duplication.** (1) Docstring says "Returns true" but the fn returns the `:max-len` value (truthy number) or `nil`, never a boolean -- reword or wrap in `boolean`. (2) The predicate `(and (fn? pred) (:max-len (meta pred)))` is duplicated here and in `describe-size-violations` (line 115), and `:clojure.spec.alpha/problems` is traversed twice. Factor a single `problem->max-len` helper and reuse it in both.

src/ctia/http/exceptions.clj
```diff
@@ -84,6 +85,18 @@
...
+(defn entity-too-large-error-handler
+  "Handle entity too large error (413)"
+  [^Exception e _data _request]
+  (logging/log! :info e (ex-message-and-data e))
+  (request-entity-too-large
+   (let [entity (:entity (ex-data e))]
+     {:error "Entity Too Large"
+      :type "Entity Size Limit Error"
+      :message (.getMessage e)
+      :entity entity
```
yogsototh, line 97: **MEDIUM -- the 413 body echoes the entire rejected entity.** The only way to reach this handler is an entity that exceeds a (large) collection-size limit, so `:entity` here is the full multi-megabyte payload the client just sent. Echoing it back means the server parses, spec-checks, then JSON-serializes and retransmits the whole oversized body -- a request->response amplification that defeats the point of cheaply rejecting oversized input with a 413. The `:msg` already names the offending fields and counts, which is enough for the client to act. Recommend dropping `:entity` from the 413 body (strip it in `throw-validation-error` for this `:type`, or omit it here). Note: the pre-existing `spec-validation-error-handler` does the same, but the size case is where echoing is most wasteful.

test/ctia/flows/crud_test.clj
```diff
@@ -305,6 +307,45 @@
          (delete-flow "delete flow deletes entities and creates events only for existing entities when some are not found"
                       {sighting-id-3 true missing-id-1 false missing-id-2 false}))))))
 
+(deftest check-spec-test
```
yogsototh, line 310: **Test coverage gaps (non-blocking).** This covers the private `check-spec` unit well, but: (1) nothing asserts an actual HTTP **413** end-to-end -- a wiring regression (renamed `:type` keyword on either side) would pass this test but break in production; the repo already has the pattern in `test/ctia/test_helpers/crud.clj` (real request asserting 400). (2) `entity-too-large-error-handler` is untested -- and its body strings (`"Entity Too Large"`) differ from `check-spec`'s (`"Entity too large"`), so this test cannot catch a handler defect. (3) The both-oversized-and-otherwise-invalid case and the multi-field `describe-size-violations` `"; "` join are untested. Consider one HTTP-level test POSTing an oversized sighting and asserting `413`.

PR #1516, opened 2026-04-09
Use wrap-routes for OTel http.route span attribute (XDR-50180)

msprunck:
## Summary

- Fix APM endpoint names showing only generic HTTP methods (POST, GET, PUT) instead of route paths like `/ctia/indicator/:id`
- Use `compojure.core/wrap-routes` to apply `trace-http/wrap-compojure-route` **after** route matching (mirroring the IROH approach from advthreat/iroh@e7c3454d)
- Bump Compojure from 1.6.1 to 1.7.2 to match IROH

## Context

The previous fix ([#1515](https://github.com/threatgrid/ctia/pull/1515), XDR-49687) placed `wrap-compojure-route` inline in the middleware chain, where it runs **before** route matching — so the route template is not yet populated and APM tools only see generic HTTP method names.

`wrap-routes` applies the middleware after the route has been matched, ensuring the `http.route` OTel span attribute is set to the actual route template.

This is the same approach used in IROH: advthreat/iroh@e7c3454d0c0ceab44e914d88f66566f6dd4bc05c

## Test plan

- [ ] Verify CTIA starts successfully
- [ ] Confirm APM traces in Datadog/Splunk show route paths (e.g., `GET /ctia/indicator/:id`) instead of generic method names
- [ ] Run existing test suite to check for regressions

### DIFF
--- a/dependabot/dependency-tree.txt
+++ b/dependabot/dependency-tree.txt
@@ -117,6 +117,9 @@ ctia:ctia:jar:1.1.1-SNAPSHOT
 |     +- org.eclipse.jetty.ee9.websocket:jetty-ee9-websocket-servlet:jar:12.1.0:compile
 |     |  \- org.eclipse.jetty.websocket:jetty-websocket-core-server:jar:12.1.0:compile
 |     \- org.eclipse.jetty.toolchain:jetty-jakarta-servlet-api:jar:5.0.2:compile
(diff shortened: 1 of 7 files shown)


msprunck:
Closing in favor of a new PR from the upstream branch (removes stale Ring/Trapperkeeper upgrade commits).

PR #1517, opened 2026-04-09
Fix OTel http.route for all responses including 401s (XDR-50180)

msprunck:
> Closes https://cisco-sbg.atlassian.net/browse/XDR-50180

## Summary

- Set the OTel `http.route` span attribute by matching request URIs against a pre-compiled route table **outside** the routing chain
- Works for ALL responses including 401s from `wrap-authenticated`, unlike `compojure.core/wrap-routes` which only fires at leaf route handlers
- New middleware in `ctia.http.middleware.otel` with try-catch isolation so instrumentation never interferes with request processing
- HEAD requests fall back to GET routes, matching Compojure's behavior
- Fix cache-control regression: restore the `cache-control-enabled?` config gate in `handler.clj` that was inadvertently removed by commit 56487d2

## Why this approach

APM tools (Datadog, Splunk) need the `http.route` span attribute (e.g., `GET /ctia/indicator/:id`) to group traces by endpoint. Without it, every trace shows up as a generic `GET /ctia/indicator/{random-id}`, making dashboards unusable.

The core tension: **Compojure only populates `:compojure/route` after a request reaches a leaf route handler, but auth middleware (`wrap-authenticated`) rejects unauthenticated requests with a 401 before they ever get there.** So any approach that depends on Compojure's routing to provide route info will miss all 401 responses — which in production are a significant share of traffic.

Two approaches that don't work:

1. **`wrap-compojure-route` before routing** (#1515, XDR-49687): Runs before Compojure has matched anything, so `:compojure/route` is always nil — APM sees no route info at all.
2. **`wrap-compojure-route` via `compojure.core/wrap-routes`** (the IROH approach): Only fires after leaf handlers, so it correctly captures routes for 200s but misses all 401s from auth middleware.

The solution is to **match routes independently** using the same `clout` library that Compojure uses internally. We extract the route table from the compiled `api` form at startup, compile it with `clout/route-compile`, and match each request before delegating to the handler chain. This runs outside and before the Compojure routing chain, so `http.route` is set regardless of what downstream middleware does with the request.

## Test plan

- [x] Unit tests for middleware logic (9 tests in `ctia.http.middleware.otel-test`)
- [x] Integration tests against real CTIA handler (2 tests in `ctia.http.server.otel-test`)
- [ ] Verify APM traces in Datadog/Splunk show route paths (e.g., `GET /ctia/indicator/:id`)
- [ ] Run full CI test suite

### DIFF
--- a/src/ctia/http/middleware/otel.clj
+++ b/src/ctia/http/middleware/otel.clj
@@ -6,6 +6,7 @@
   fires at leaf route handlers."
   (:require [clojure.tools.logging :as log]
             [clout.core :as clout]
+            [compojure.api.routes :as routes]
             [steffan-westcott.clj-otel.api.trace.http :as trace-http]))
 
 (defn compile-route-table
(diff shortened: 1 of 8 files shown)


src/ctia/http/middleware/otel.clj
```diff
@@ -0,0 +1,47 @@
...
+
+(defn find-route-template
+  "Return the route template matching `request`, or nil if none matches.
+  Falls back to GET routes for HEAD requests, matching Compojure's behavior."
+  [compiled-routes request]
+  (when-let [method (:request-method request)]
+    (let [method-name (name method)
+          match-route (fn [routes]
+                        (some (fn [[compiled-path template]]
+                                (when (clout/route-matches compiled-path request)
```
ereteog, line 28: The linear scan with `some` returns the first `clout/route-matches` hit — so correctness depends on `get-routes` returning specific paths (e.g. `/ctia/indicator/search`) before parameterized ones (`/ctia/indicator/:id`). That happens to hold today based on definition order, but nothing enforces it.

Worth either sorting compiled routes (exact before parameterized) within each method group, or adding an integration test that calls `get-routes` on the real handler and asserts the ordering invariant. Otherwise a future route addition could silently mis-attribute traces.

ereteog (approved):
Approach is solid — independent route matching with clout outside the routing chain is the right way to get `http.route` on 401s. Error isolation, HEAD fallback, and test coverage all look good.

One concern on route ordering (see inline) — would be nice to harden that before merge, but not a blocker.

src/ctia/http/middleware/otel.clj
msprunck: Good catch on the ordering dependency. We intentionally rely on definition order here rather than sorting, because the goal is to **mirror Compojure's own matching behavior** — which is itself first-match-wins in definition order.

The `Route` record's `childs` field drives both actual request dispatch and `get-routes` output (it walks the same tree). Re-sorting (e.g., exact before parameterized) would risk diverging from Compojure's actual matching order, which is more nuanced than just "exact vs parameterized" — consider paths like `/ctia/indicator/:id/sighting` vs `/ctia/indicator/:id`.

If `get-routes` ever returned routes in a different order than Compojure matches them, Compojure's own Swagger spec (which also uses `get-routes`) would also be wrong — the whole ecosystem implicitly depends on this ordering.

And if a future route addition broke the ordering, Compojure would misroute real requests — a far more visible failure than bad telemetry, caught immediately by existing tests.

Added a comment in the code documenting this assumption.

safi (approved):

jamoser-cisco:
@msprunck @ereteog  I'm not an expert in DataDog.  I don't see any 401s with missing route paths, but I am seeing 404s.  Should this have fixed those as well?

msprunck:
@jamoser-cisco The middleware works by matching the request URI against the compiled route table **before** dispatch, so it sets `http.route` regardless of the response status (401, 404, 200, etc.).

For 404s there are two cases:

1. **A known route returns 404** (e.g., `GET /ctia/indicator/:id` where the ID doesn't exist): the route template matches, so `http.route` is set to `/ctia/indicator/:id`. These should already be covered.
2. **No route matches at all** (unknown paths, scanners, typos): there's no route template to assign, so `http.route` is intentionally left empty. This is correct behavior per OTel semantic conventions.

Could you share a few example 404 paths you're seeing without `http.route`? That would help determine which case it is.

jamoser-cisco:
@msprunck no? =) I thought the point here was to fix the missing paths, so I can't tell what the paths for these are. =) Or maybe I just don't know how?

Screenshot 2026-04-23 at 6 56 59 AM

jamoser-cisco:
I think these are it? And I think they are no route matches:
Screenshot 2026-04-23 at 6 59 13 AM

PR #1519, opened 2026-04-16
Upgrade Trapperkeeper & Jetty dependencies

safi:
> XDR-50310

Upgrade Jetty(ring-adapter-jetty) and Trapperkeeper dependencies.

### DIFF
--- a/project.clj
+++ b/project.clj
@@ -7,7 +7,7 @@
 (def metrics-clojure-version "2.10.0")
 (def netty-version "4.1.125.Final")
 (def perforate-version "0.3.4")
-(def ring-version "1.9.5")
+(def ring-version "1.15.3")
 (def kitchensink-version "3.4.0")
 (def slf4j-version "2.0.17")
 (def schema-generators-version "0.1.5")
@@ -113,13 +113,13 @@
                  [ring/ring-devel ~ring-version]
                  [ring-cors "0.1.13"]
                  [commons-codec "1.17.1"] ;ring/ring* > threatgrid/ctim, threatgrid/clj-momo, clj-http
-                 [ring/ring-codec "1.1.3"]
+                 [ring/ring-codec "1.3.0"]
                  [threatgrid/clj-jwt "0.5.0"]
                  [threatgrid/ring-turnstile-middleware "0.1.1"]
                  [threatgrid/ring-jwt-middleware "1.1.7"]
                  [org.clojure/data.json "1.0.0"] ;; threatgrid/ring-jwt-middleware > threatgrid/ctim
                  [scopula "0.1.4"]
-                 [org.clojure/tools.reader "1.3.6"] ;; org.clojure/tools.namespace > ring-middleware-format
+                 [org.clojure/tools.reader "1.4.0"] ;; bump ring/ring-devel
 
                  ;; clients
                  [clj-http "3.13.1"]
@@ -143,7 +143,7 @@
 
                  ;; Docs
                  [markdown-clj "1.10.1"]
-                 [hiccup "2.0.0-alpha2"]
+                 [hiccup "2.0.0"] ;; bump ring/ring-devel
 
                  ;; Encryption
                  [lock-key "1.5.0"]
@@ -152,7 +152,7 @@
                  [threatgrid/redismq "0.1.1"]
 
                  [org.apache.zookeeper/zookeeper "3.8.4"] ; override zookeeper-clj, org.onyxplatform/onyx-kafka
-                 [commons-io "2.18.0"] ;; address CVE-2024-47554
+                 [commons-io "2.20.0"] ;; bump ring/ring-core
                  [args4j "2.33"] ;bump org.onyxplatform/onyx-kafka, threatgrid/ctim
                  [com.stuartsierra/component "1.1.0"] ;org.onyxplatform/onyx-kafka internal override
                  [org.onyxplatform/onyx-kafka "0.14.5.0"]
@@ -232,7 +232,7 @@
                                   [circleci/circleci.test "0.5.0"]
                                   [org.clojure/math.combinatorics "0.3.0"]
                                   [org.clojure/data.priority-map "1.1.0"]
-                                  [org.clojure/tools.namespace "1.2.0"]]
+                                  [org.clojure/tools.namespace "1.5.0"]] ;; bump ring/ring-devel
                    :pedantic? :warn
                    :source-paths ["dev"]}
              :ci {:pedantic? :abort
--- a/project.clj
+++ b/project.clj
@@ -84,7 +84,7 @@
                  ;; Trapperkeeper
                  [puppetlabs/trapperkeeper ~trapperkeeper-version]
                  [puppetlabs/kitchensink ~kitchensink-version]
-                 [prismatic/plumbing "0.5.5"] ;; upgrade puppetlabs/trapperkeeper
+                 [prismatic/plumbing "0.6.0"] ;; upgrade puppetlabs/trapperkeeper
                  [clj-commons/clj-yaml "1.0.26"] ;; upgrade snakeyaml dep
 
                  ;; Schemas
@@ -119,7 +119,7 @@
                  [threatgrid/ring-jwt-middleware "1.1.7"]
                  [org.clojure/data.json "1.0.0"] ;; threatgrid/ring-jwt-middleware > threatgrid/ctim
                  [scopula "0.1.4"]
-                 [org.clojure/tools.reader "1.4.0"] ;; bump ring/ring-devel
+                 [org.clojure/tools.reader "1.4.0"] ;; tools.namespace > ring-middleware-format
 
                  ;; clients
                  [clj-http "3.13.1"]
@@ -152,7 +152,7 @@
                  [threatgrid/redismq "0.1.1"]
 
                  [org.apache.zookeeper/zookeeper "3.8.4"] ; override zookeeper-clj, org.onyxplatform/onyx-kafka
-                 [commons-io "2.20.0"] ;; bump ring/ring-core
+                 [commons-io "2.20.0"] ;; address CVE-2024-47554; aligned with ring/ring-core 1.15.3
                  [args4j "2.33"] ;bump org.onyxplatform/onyx-kafka, threatgrid/ctim
                  [com.stuartsierra/component "1.1.0"] ;org.onyxplatform/onyx-kafka internal override
                  [org.onyxplatform/onyx-kafka "0.14.5.0"]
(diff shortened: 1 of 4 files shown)


msprunck (commented):
Review withdrawn; see [updated review](https://github.com/threatgrid/ctia/pull/1519#pullrequestreview-4170293703).

project.clj
```diff
@@ -75,14 +76,14 @@
...
+                 [org.clojure/tools.cli "1.0.206"]
                  [pandect "0.6.1"]
                  [org.clojure/math.combinatorics "0.3.0"]
                  [version-clj "2.0.1"]
 
                  ;; Trapperkeeper
                  [puppetlabs/trapperkeeper ~trapperkeeper-version]
-                 [puppetlabs/kitchensink ~trapperkeeper-version]
+                 [puppetlabs/kitchensink ~kitchensink-version]
                  [prismatic/plumbing "0.5.5"] ;; upgrade puppetlabs/trapperkeeper
```
msprunck: **Suggestion (non-blocking): bump to `0.6.0`.**

Trapperkeeper [#294](https://github.com/puppetlabs/trapperkeeper/issues/294) — `Method code too large!` at ~122+ services — is latent in `plumbing 0.5.5`; [fixed upstream](https://github.com/plumatic/plumbing/issues/138) in `0.6.0`. The trapperkeeper bump in this PR does not include the fix.

CTIA has ~17 `tk/defservice` calls today, so it's not an immediate problem — but bumping now is a one-line preventive change:

```clojure
[prismatic/plumbing "0.6.0"] ;; upgrade puppetlabs/trapperkeeper; also preempts #294 bytecode-size limit
```

project.clj
```diff
@@ -112,13 +113,13 @@
...
                  [commons-codec "1.17.1"] ;ring/ring* > threatgrid/ctim, threatgrid/clj-momo, clj-http
-                 [ring/ring-codec "1.1.3"]
+                 [ring/ring-codec "1.3.0"]
                  [threatgrid/clj-jwt "0.5.0"]
                  [threatgrid/ring-turnstile-middleware "0.1.1"]
                  [threatgrid/ring-jwt-middleware "1.1.7"]
                  [org.clojure/data.json "1.0.0"] ;; threatgrid/ring-jwt-middleware > threatgrid/ctim
                  [scopula "0.1.4"]
-                 [org.clojure/tools.reader "1.3.6"] ;; org.clojure/tools.namespace > ring-middleware-format
+                 [org.clojure/tools.reader "1.4.0"] ;; bump ring/ring-devel
```
msprunck: **The new comment is factually inaccurate.** Verified against the `ring-devel` POM and `lein deps :tree`:

- `ring-devel` declares a newer `tools.reader` version in its `<dependencyManagement>`, not 1.4.0.
- `tools.namespace 1.5.0` (dev dep, line 235) directly depends on `tools.reader 1.4.0`.
- `ring-middleware-format 0.7.4` depends on an older `tools.reader`.

So 1.4.0 is the effective conflict-resolution point between `tools.namespace` and `ring-middleware-format` — **not** driven by ring-devel. The previous comment (`;; org.clojure/tools.namespace > ring-middleware-format`) was actually correct.

Suggest:
```clojure
[org.clojure/tools.reader "1.4.0"] ;; tools.namespace > ring-middleware-format
```

msprunck (commented):
## Review

Verified the dependency changes against `lein deps :tree` and the project POMs. Main findings below; see inline comments for specifics.

### Important

1. **`UriCompliance` override is broader than the stated requirement.** The mode selected re-enables a set of path-parsing behaviors that Jetty's newer defaults tightened, where only one is actually needed per the existing comment. Consider scoping it down, or at minimum expand the inline comment so the full scope is documented. See inline comment on `src/ctia/http/server.clj`.

2. **CVE-2024-47554 rationale deleted from `commons-io` pin.** An intermediate commit on this branch had `;; address CVE-2024-47554, align with ring 1.15` before it was truncated to `;; bump ring/ring-core`. The artifact is still safe (2.20.0 contains the 2.14.0 fix), but the *reason* for the pin is what protects against a future dependabot downgrade. See inline comment on line 155.

3. **`tools.reader 1.4.0` and `tools.namespace 1.5.0` comments are inaccurate.** Verified via `lein deps :tree` and the POMs: the real drivers are different from what the comments say. See inline comments.

### Suggestion

4. **Consider bumping `prismatic/plumbing` to `0.6.0`.** Works around trapperkeeper [#294](https://github.com/puppetlabs/trapperkeeper/issues/294) — `plumbing 0.5.5` fails with `Method code too large!` at ~122+ services. CTIA has ~17 `tk/defservice` calls today, so not currently at risk, but bumping now is cheap insurance. The trapperkeeper bump in this PR does not include the fix. See inline comment on line 87.

### Notes

- **No silent failures in the configurator.** For standard HTTP/HTTPS setups the `instance?` guard matches at least one factory. Optional `log/info` on apply and `log/warn` on zero matches would be nice for audit but isn't required.
- **Test coverage is strong but implicit.** `test/ctia/test_helpers/crud.clj:265` URL-encodes fixture external_ids that contain `/`, so 20+ entity suites transitively exercise the relevant path through real Jetty. CI's 20 shards confirm the gate works. If a future refactor replaced slashy fixtures with plain strings the gate would silently disappear — worth a dedicated ~15-line explicit assertion in `test/ctia/http/server_test.clj` as insurance, but not blocking.
- **No `javax.servlet` usage** in `src/` or `test/` — jakarta migration has no code impact.

### Strengths

- `kitchensink-version` decoupling is the correct response to upstream splitting it from trapperkeeper.
- `hiccup` alpha → 2.0.0 stable is a quality improvement.
- Configurator correctly walks connectors and factories with an `instance?` guard.
- `dependabot/pom.xml` mirrors `project.clj` consistently.

*(Reposting this review with a clean slate; the prior review was withdrawn.)*

project.clj
```diff
@@ -142,7 +143,7 @@
 
                  ;; Docs
                  [markdown-clj "1.10.1"]
-                 [hiccup "2.0.0-alpha2"]
+                 [hiccup "2.0.0"] ;; bump ring/ring-devel
```
msprunck, line 146: Accurate — `ring-devel` directly declares `hiccup 2.0.0` in its POM. No action needed, just confirming I verified this one.

Optional tightening: `;; required by ring/ring-devel`.

project.clj
```diff
@@ -151,7 +152,7 @@
                  [threatgrid/redismq "0.1.1"]
 
                  [org.apache.zookeeper/zookeeper "3.8.4"] ; override zookeeper-clj, org.onyxplatform/onyx-kafka
-                 [commons-io "2.18.0"] ;; address CVE-2024-47554
+                 [commons-io "2.20.0"] ;; bump ring/ring-core
```
msprunck: **Losing the CVE rationale is comment rot.** An earlier commit on this branch read `;; address CVE-2024-47554, align with ring 1.15`; the final version dropped the CVE part.

The artifact is still safe (2.20.0 > 2.14.0 fix floor), but the *reason for pinning* is what protects against a future dependabot downgrade. `ring-core 1.15.3` now directly pins `commons-io 2.20.0` in its POM, so the override is at parity — but if ring ever goes to a `commons-fileupload2-core` that pulls a lower commons-io version, this pin is the safety net.

Suggest restoring both justifications:
```clojure
[commons-io "2.20.0"] ;; address CVE-2024-47554; aligned with ring/ring-core 1.15.3
```

project.clj
```diff
@@ -231,7 +232,7 @@
                                   [circleci/circleci.test "0.5.0"]
                                   [org.clojure/math.combinatorics "0.3.0"]
                                   [org.clojure/data.priority-map "1.1.0"]
-                                  [org.clojure/tools.namespace "1.2.0"]]
+                                  [org.clojure/tools.namespace "1.5.0"]] ;; bump ring/ring-devel
```
msprunck: **Misleading.** `ring-devel` does not depend on `tools.namespace` directly — it pulls `ns-tracker 1.0.0`, which declares `tools.namespace 1.5.0`. Older `ns-tracker` versions pinned `tools.namespace 0.2.11`.

Suggest:
```clojure
[org.clojure/tools.namespace "1.5.0"]] ;; bump via ring/ring-devel → ns-tracker
```

src/ctia/http/server.clj
```diff
@@ -294,6 +295,14 @@
         :min-threads min-threads
         :max-threads max-threads
         :join? false
-        :send-server-version? send-server-version})
+        :send-server-version? send-server-version
+        ;; Jetty 12 rejects %2F (encoded slashes) in URI paths by default.
+        ;; LEGACY restores Jetty 9.4 behavior, allowing encoded path separators.
```
msprunck, line 300: **Override is broader than the stated requirement.** The compliance mode selected re-enables a set of path-parsing behaviors that Jetty's newer defaults tightened. The actual need (based on the existing comment) is only the encoded path separator case.

Two options:

1. **Scope it down.** Jetty's `UriCompliance` supports `DEFAULT.with(...)` to re-enable only specific violations by name — recommended for a security-sensitive service. See the Jetty `UriCompliance` API docs for the builder pattern.

2. **Keep the current override** but expand the code comment so future maintainers know the full scope of what's being re-enabled, not just the originally motivating case.

Happy to discuss offline if useful.

src/ctia/http/server.clj
```diff
@@ -294,6 +295,14 @@
...
+        :configurator (fn [^Server server]
+                        (doseq [connector (.getConnectors server)]
+                          (doseq [factory (.getConnectionFactories connector)]
+                            (when (instance? HttpConnectionFactory factory)
+                              (doto (.getHttpConfiguration ^HttpConnectionFactory factory)
+                                (.setUriCompliance UriCompliance/LEGACY))))))})
```
msprunck, line 306: **Non-blocking: consider logging.** The configurator silently no-ops if no connector exposes an `HttpConnectionFactory` (e.g. exotic connector configs). For standard ring-jetty-adapter HTTP/HTTPS setups I verified this matches at least one factory.

Audit-trail optional improvement:

```clojure
:configurator (fn [^Server server]
                (let [n (atom 0)]
                  (doseq [connector (.getConnectors server)
                          factory (.getConnectionFactories connector)
                          :when (instance? HttpConnectionFactory factory)]
                    (.setUriCompliance (.getHttpConfiguration ^HttpConnectionFactory factory)
                                       UriCompliance/LEGACY)
                    (swap! n inc))
                  (if (pos? @n)
                    (log/infof "Applied URI compliance override to %d HttpConfiguration(s)" @n)
                    (log/warn "URI compliance configurator matched no HttpConnectionFactory"))))
```

On a quiet connector-config regression the `log/warn` would be the first signal.

msprunck (approved):
Thanks for the thorough follow-up. Verified `d68b14b` addresses the comment/pin findings:

- `prismatic/plumbing` → `0.6.0` ✅
- `commons-io` CVE rationale restored ✅
- `tools.reader` comment fixed ✅
- `tools.namespace` comment fixed ✅

On the `UriCompliance` override scope: your reasoning to keep the current setting to avoid breaking unknown code paths is a defensible call — accepting that.

Approving.

PR #1521, opened 2026-05-29
fix(incident): override stale upstream :opened on first New → Open (XDR-54438)

ereteog:
> Close https://cisco-sbg.atlassian.net/browse/XDR-54438
> Related https://cisco-sbg.atlassian.net/browse/XDR-42815

The Operational Insights "Mean Time To Engage" (MTTE) tile shows 0 mins for the team metric because `:new_to_opened` is never computed for promoted incidents.

**Root cause.** CTIM marks `:opened` as mandatory, so promoted incidents (e.g., from XDR Analytics, detection engines) always arrive with an upstream-stamped `:opened` that is unrelated to actual analyst engagement — and often earlier than CTIA's `:created`. When the user clicks New → Open in the UI, the PATCH flow's deep-merge carries that stale `:opened` forward. Then `update-interval :new_to_opened` requires `:created ≤ :opened`, silently bails on `:created > :opened`, and the interval is never stored. The MTTE tile averages a missing field → 0.

**Fix.** In `apply-status-update-logic`, on the first New → Open transition, override `:opened` with NOW when `:new_to_opened` has not yet been recorded. This mirrors `compute-intervals`' "don't clobber existing interval" semantics. Engagement time is the analyst's click time, not whatever the client passes.

**Tests.**
- Unit: override fires when `:new_to_opened` is absent.
- Unit (negative): inert when `:new_to_opened` already recorded.
- Unit (negative): inert on `New → New: Triaged`.
- Composed: `apply-status-update-logic` + `compute-intervals` produces `:new_to_opened`.
- HTTP integration: `PATCH /ctia/incident/:id` then `GET /ctia/incident/metric/average?aggregate-on=intervals.new_to_opened` — the exact path the MTTE tile reads — returns ~300s (5 min elapsed) instead of nil.
- **Full-lifecycle scenario**: walks one incident through `New → Open → Open: Contained → Closed: Confirmed Threat` and asserts `/metric/average` returns expected averages for `new_to_opened`, `new_to_contained`, and `opened_to_closed`. Guards all three Mean Time To X tile metrics against future regressions.

<a name="qa">[§](#qa)</a> QA
============================

1. Promote an incident in an environment where upstream sets `incident_time.opened`.
2. Move it from New → Open in the UI.
3. Open the Operational Insights dashboard and confirm Team MTTE reflects the actual elapsed time, not 0 mins.

\`\`\`
intern: fix Operational Insights Team MTTE tile showing 0 mins for promoted incidents (XDR-54438).
public: Operational Insights now reflects time-to-engage for incidents promoted from upstream sources.
\`\`\`

### DIFF
--- a/src/ctia/entity/incident.clj
+++ b/src/ctia/entity/incident.clj
@@ -173,13 +173,21 @@
           new-opened-date (when (and (new-status? old-status)
                                      (open-status? new-status))
                             (get-in status-update [:incident_time :opened]))
+          ;; On (New|Open) → Open: Contained, override :contained with NOW for the same reason as
+          ;; :opened: an upstream-stamped :contained earlier than :created would silently elide
+          ;; :new_to_contained via update-interval's (earlier ≤ later) guard, hiding MTTC.
+          ;; Trigger mirrors compute-intervals' :new_to_contained clause exactly.
+          new-contained-date (when (and (or (new-status? old-status) (open-status? old-status))
+                                        (contained-status? new-status))
+                               (get-in status-update [:incident_time :contained]))
           ;; prev-* and new-* dates for each verb are mutually exclusive (one requires the old
           ;; status to be in the category, the other requires it not to be).
           merged-incident-time (cond-> (merge status-stamped-incident-time client-incident-time)
                                  prev-closed-date (assoc :closed prev-closed-date)
                                  prev-opened-date (assoc :opened prev-opened-date)
                                  new-closed-date (assoc :closed new-closed-date)
-                                 new-opened-date (assoc :opened new-opened-date))]
+                                 new-opened-date (assoc :opened new-opened-date)
+                                 new-contained-date (assoc :contained new-contained-date))]
       (cond-> new-obj
         (seq merged-incident-time) (assoc :incident_time merged-incident-time)))
     ;; No status change or no previous object, return as-is
--- a/src/ctia/entity/incident.clj
+++ b/src/ctia/entity/incident.clj
@@ -147,10 +147,13 @@
     (let [old-status (:status prev-obj)
           new-status (:status new-obj)
           status-update (make-status-update {:status new-status})
-          ;; Merge the incident_time updates from status change logic
-          ;; Existing values take precedence (set-once semantics for opened, contained, etc.)
-          incident-time-updates (get status-update :incident_time {})
-          current-incident-time (get new-obj :incident_time {})
+          ;; Layered precedence for :incident_time on a status change:
+          ;;   1. status-stamped defaults (e.g. {:opened NOW} for any open status) seed the map
+          ;;   2. client-supplied values from the PATCH body override the defaults
+          ;;   3. cond-> overrides below trump both for category-transition fields the server owns
+          ;;      (e.g., :opened on New → Open, :closed on non-closed → closed)
+          status-stamped-incident-time (get status-update :incident_time {})
+          client-incident-time (get new-obj :incident_time {})
           ;; If transitioning from closed to closed, preserve the original closed date
           prev-closed-date (when (and (closed-status? old-status)
                                       (closed-status? new-status))
@@ -172,7 +175,7 @@
                             (get-in status-update [:incident_time :opened]))
           ;; prev-* and new-* dates for each verb are mutually exclusive (one requires the old
           ;; status to be in the category, the other requires it not to be).
-          merged-incident-time (cond-> (merge incident-time-updates current-incident-time)
+          merged-incident-time (cond-> (merge status-stamped-incident-time client-incident-time)
                                  prev-closed-date (assoc :closed prev-closed-date)
                                  prev-opened-date (assoc :opened prev-opened-date)
                                  new-closed-date (assoc :closed new-closed-date)
--- a/src/ctia/entity/incident.clj
+++ b/src/ctia/entity/incident.clj
@@ -100,10 +100,10 @@
   {:status IncidentStatus})
 
 (defn new-status? [status]
-   (some? (re-matches #"New(: .+)?" status)))
+  (some? (and status (re-matches #"New(: .+)?" status))))
 
 (defn hold-status? [status]
-   (some? (re-matches #"Hold(: .+)?" status)))
+  (some? (and status (re-matches #"Hold(: .+)?" status))))
 
 (defn open-status? [status]
   (some? (and status (re-matches #"Open(: .+)?" status))))
@@ -163,12 +163,22 @@
           new-closed-date (when (and (closed-status? new-status)
                                      (not (closed-status? old-status)))
                             (get-in status-update [:incident_time :closed]))
-          ;; prev-closed-date and new-closed-date are mutually exclusive:
-          ;; prev-closed-date requires old-status to be closed, new-closed-date requires it not to be.
+          ;; On first open (New → Open), set :opened to NOW if the engagement interval has not
+          ;; yet been recorded. This is the same invariant compute-intervals uses ("don't clobber
+          ;; existing interval"); applying it here ensures MTTE reflects the actual click time and
+          ;; not whatever upstream pre-stamped on a promoted incident (CTIM marks :opened as
+          ;; mandatory, so promoted incidents always carry it).
+          new-opened-date (when (and (new-status? old-status)
+                                     (open-status? new-status)
+                                     (not (get-in prev-obj [:intervals :new_to_opened])))
+                            (get-in status-update [:incident_time :opened]))
+          ;; prev-* and new-* dates for each verb are mutually exclusive (one requires the old
+          ;; status to be in the category, the other requires it not to be).
           merged-incident-time (cond-> (merge incident-time-updates current-incident-time)
                                  prev-closed-date (assoc :closed prev-closed-date)
                                  prev-opened-date (assoc :opened prev-opened-date)
-                                 new-closed-date (assoc :closed new-closed-date))]
+                                 new-closed-date (assoc :closed new-closed-date)
+                                 new-opened-date (assoc :opened new-opened-date))]
       (cond-> new-obj
         (seq merged-incident-time) (assoc :incident_time merged-incident-time)))
     ;; No status change or no previous object, return as-is
--- a/src/ctia/entity/incident.clj
+++ b/src/ctia/entity/incident.clj
@@ -163,14 +163,12 @@
           new-closed-date (when (and (closed-status? new-status)
                                      (not (closed-status? old-status)))
                             (get-in status-update [:incident_time :closed]))
-          ;; On first open (New → Open), set :opened to NOW if the engagement interval has not
-          ;; yet been recorded. This is the same invariant compute-intervals uses ("don't clobber
-          ;; existing interval"); applying it here ensures MTTE reflects the actual click time and
-          ;; not whatever upstream pre-stamped on a promoted incident (CTIM marks :opened as
-          ;; mandatory, so promoted incidents always carry it).
+          ;; On New → Open, override :opened with NOW. CTIM marks :opened as mandatory, so promoted
+          ;; incidents always carry an upstream-stamped value; MTTE is computed from :created →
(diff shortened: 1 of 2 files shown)


src/ctia/entity/incident.clj
```diff
@@ -163,12 +163,22 @@
...
-          ;; prev-closed-date and new-closed-date are mutually exclusive:
-          ;; prev-closed-date requires old-status to be closed, new-closed-date requires it not to be.
+          ;; On first open (New → Open), set :opened to NOW if the engagement interval has not
+          ;; yet been recorded. This is the same invariant compute-intervals uses ("don't clobber
+          ;; existing interval"); applying it here ensures MTTE reflects the actual click time and
+          ;; not whatever upstream pre-stamped on a promoted incident (CTIM marks :opened as
+          ;; mandatory, so promoted incidents always carry it).
+          new-opened-date (when (and (new-status? old-status)
+                                     (open-status? new-status)
+                                     (not (get-in prev-obj [:intervals :new_to_opened])))
```
yogsototh: **HIGH** [correctness/cross-cutting] -- **Dead guard fires on every New → Open, not just the first.**

`(get-in prev-obj [:intervals :new_to_opened])` is always `nil` in the real HTTP flow. `prev-obj` reaches this function via `read-records` → `handle-read-many`, which applies `es-partial-stored->partial-stored` = `un-store-incident` = `(dissoc doc :intervals)`. So `:intervals` is stripped before `prev-obj` ever gets here, the third `and` clause is permanently true, and the override fires unconditionally.

The MTTE feature still works (the write-side `update-interval` has its own independent don't-clobber guard), so this is not a feature regression. But because there is no status state machine (the generic PATCH route accepts any status), `Open → New → Open` is reachable: on the second New → Open, `:opened` is re-stamped to the later time while `:intervals.new_to_opened` keeps the original value, leaving the two fields inconsistent.

Suggestion: drop the `:intervals` clause (the override is intentionally unconditional on New → Open and harmless thanks to `update-interval`'s guard) and update the comment, OR gate on a signal that survives `un-store-incident`. As written the comment claims a 'don't clobber existing interval' invariant this code does not actually enforce.

test/ctia/entity/incident_test.clj
```diff
@@ -98,7 +98,55 @@
...
+    (testing "XDR-54438: New → Open preserves :opened when :new_to_opened interval already exists"
+      ;; If the engagement interval was already computed (e.g., previous transition cycle), do not
+      ;; overwrite. This preserves first-engagement semantics across re-open scenarios.
+      (helpers/fixture-with-fixed-time
+       t3
+       (fn []
+         (let [prev-obj {:status "New"
+                         :created t1
+                         :incident_time {:opened t1}
+                         :intervals {:new_to_opened 42}}
```
yogsototh: **MEDIUM** [tests] -- **This test validates a state unreachable in production.** It hand-builds `prev-obj` with `:intervals {:new_to_opened 42}`, but `un-store-incident` strips `:intervals` from every `prev-obj` before `apply-status-update-logic` runs in the real flow. The test passes only because it calls the function directly, so it gives false confidence that the guard is load-bearing. Either delete it, relabel it as unit-only intent (note the store strips `:intervals`), or convert to an E2E test that performs a real second New → Open and asserts the actual production behavior.

test/ctia/entity/incident_test.clj
```diff
@@ -589,6 +640,108 @@
...
+                                                         :query (format "title:\"%s\"" unique-title)})
+                                     :parsed-body
+                                     (get-in [:data :intervals (keyword field)])))
+                    near? (fn [expected actual]
+                            (and actual (<= (- expected 10) actual (+ expected 10))))]
+                (is (near? 300 (metric-avg "new_to_opened"))
+                    "MTTE tile metric must reflect ~5 min from creation to first Open")
+                (is (near? 600 (metric-avg "new_to_contained"))
+                    "MTTC tile metric must reflect ~10 min from creation to Open: Contained")
+                (is (= 600.0 (metric-avg "opened_to_closed"))
```
yogsototh: **LOW** [tests] -- **Exact float equality on an ES aggregation result is brittle.** `(= 600.0 ...)` compares against a raw ES `avg` double. The existing metric test in this same file defends against double fuzz with `(some-> ... Math/floor long)`. Use that proven idiom (or a numeric `match?` predicate) for consistency and robustness.

test/ctia/entity/incident_test.clj
```diff
@@ -589,6 +640,108 @@
...
+                                 (-> (GET app "ctia/incident/metric/average"
+                                          :headers {"Authorization" "45c1f5e3f05d0"}
+                                          :query-params {:aggregate-on (str "intervals." field)
+                                                         :from (str (jt/instant (t/minus fixed-now (t/seconds 1))))
+                                                         :query (format "title:\"%s\"" unique-title)})
+                                     :parsed-body
+                                     (get-in [:data :intervals (keyword field)])))
+                    near? (fn [expected actual]
+                            (and actual (<= (- expected 10) actual (+ expected 10))))]
+                (is (near? 300 (metric-avg "new_to_opened"))
```
yogsototh: **LOW** [tests] -- **±10s timing tolerance is a smell and can flake.** A `±10s` window on a 300s expected value can both mask a real off-by-one-step regression and flake under a GC pause / slow CI gap > 10s between the create call and the fixed-time block. More deterministic: capture the actual stored `:created` from the create response and compute the expected interval from it, rather than asserting a hard-coded 300/600 with a fuzz window.

yogsototh (changes requested):
**Verdict: CHANGES REQUESTED** -- 0 CRITICAL, 1 HIGH, 2 MEDIUM, several LOW

The MTTE fix is correct and genuinely well-tested at the E2E level: the full-lifecycle test walks `New → Open → Open: Contained → Closed` through the real HTTP path and proves `/metric/average` returns a non-zero `new_to_opened`. The nil-guard on `new-status?`/`hold-status?` is a real latent-NPE fix. The blocking items below are about a guard that is dead in the production flow, a comment that now misstates the behavior, and a test that validates an unreachable state -- all quick to fix and worth resolving before merge so the code doesn't mislead future maintainers.

### Findings

- **HIGH** [correctness] -- The `:intervals` guard in `apply-status-update-logic` is dead in production (`:intervals` is stripped by `un-store-incident` before `prev-obj` arrives), so the override fires on every New → Open. Harmless for MTTE itself, but `Open → New → Open` (reachable -- no status state machine) leaves `:opened` and `:intervals.new_to_opened` inconsistent. (inline at `incident.clj:173`)
- **MEDIUM** [quality] -- `incident.clj:151`: the comment now reads "Existing values take precedence (set-once semantics for opened…)", but for New → Open the new `new-opened-date` clause overrides the existing/client `:opened` with server NOW. Reword to note the New → Open exception. (Not anchorable inline -- the line sits just outside the diff hunk.)
- **MEDIUM** [tests] -- Unit test asserts an unreachable `:intervals`-bearing `prev-obj` shape. (inline at `incident_test.clj:129`)
- **LOW** [correctness] -- `:contained` has the same latent elision the PR fixes for `:opened`: a promoted incident arriving with an upstream `:contained` earlier than `:created` would still elide `new_to_contained`. The fix is asymmetric. Confirm whether promoted incidents can carry `:contained`; if so, a symmetric override is warranted (follow-up).
- **LOW** [quality] -- Layering: the realize-phase function reaches into `:intervals`, a storage-only key stripped before it runs. The correctly-placed sibling is the `opened_to_closed` dissoc, which lives in `compute-intervals` (the write layer) where `:intervals` actually exists.
- **LOW** [tests] -- No test pins the nil-guard (`(new-status? nil)` / `(hold-status? nil)` falsey); the cited `entity-realize-fn-test` short-circuits before reaching those predicates.
- **LOW** [tests] -- `run-reclose-scenario` (Closed → Open) never asserts `:opened` is unchanged, and Hold → Open is untested -- a future broadening of the clause would go uncaught.

### PASS: Security (net integrity hardening -- `:opened` becomes server-authoritative on first open), Coding Style

<details><summary>Non-blocking discussion</summary>

- **Migration**: existing New-status incidents get `:opened` overwritten with NOW on their first post-deploy Open. Intended, but it's a silent field rewrite -- worth a CHANGELOG note.
- **MTTE semantics**: post-override, `new_to_opened = NOW - created`. For incidents created long ago upstream, this measures age-since-creation, not time-in-analyst-queue. Confirm with product this is the intended MTTE definition.

</details>

Also left 4 inline comments.

yogsototh (approved):
**Verdict: APPROVED** — the blocking HIGH from the prior review is resolved.

`c897de899` drops the dead `:intervals` guard so the New → Open override is now unconditional, and rewrites the comment to describe the actual behavior (reset `:opened` on each New → Open). Traced the reachable `Open → New → Open` path: `update-interval`'s own don't-clobber guard keeps `:new_to_opened` (the MTTE metric) locked to the first engagement, so the residual `:opened` (latest-open) vs `:new_to_opened` (first-open) divergence is now an intentional, documented choice rather than a dead-guard accident. CI green (23/23).

**Resolved**: HIGH (dead guard, `incident.clj`), MEDIUM (unreachable `:intervals` unit test — replaced), LOW (layering — no longer touches `:intervals`).

One non-blocking nit remains (not anchorable inline — outside the diff hunk):

- `incident.clj:151`: the parenthetical "set-once semantics for opened, contained, etc." is now inaccurate for `:opened`, which the New → Open clause unconditionally overrides. Worth tightening so the comment does not mislead future maintainers.

<details><summary>Optional follow-ups (carried over from prior review, all non-blocking)</summary>

- `:contained` has the same latent elision this PR fixes for `:opened` (asymmetric fix); confirm whether promoted incidents can carry an upstream `:contained` earlier than `:created`.
- Lifecycle test: `(= 600.0 ...)` exact float equality on an ES `avg` is brittle — the existing metric test in this file uses `Math/floor long`.
- Lifecycle test: ±10s tolerance on a 300s value can flake under a slow CI gap; deriving the expected interval from the stored `:created` is more deterministic.
- No test pins `(new-status? nil)` / `(hold-status? nil)` falsey.
- `run-reclose-scenario` does not assert `:opened` is unchanged; Hold → Open is untested.

</details>

ereteog:
@yannesposito addressed the review feedback in 4 commits on top:

**HIGH — `:intervals` guard (c897de89)**
Removed the dead `(when-not (:intervals new-obj) ...)` guard at incident.clj:173. `un-store-incident` strips `:intervals` before this code runs, so the guard never triggered.

**MEDIUM — misleading variable names + nil-safety + gap tests (fa576413)**
- Renamed `incident-time-updates`/`current-incident-time` to `status-stamped-incident-time`/`client-incident-time` and added the layered-precedence comment so the merge order reads clearly.
- Added nil-safety on the 5 status predicates (`new-status?`, `hold-status?`, `open-status?`, `contained-status?`, `closed-status?`).
- Added 4 gap tests in `apply-status-update-logic-test`: Hold → Open inert, open → open `:opened` preservation, Hold → Open: Contained inert for `:contained`, plus the New → Open: Contained override.

**Sibling fix — MTTC parallel to MTTE (7dcf1da8)**
The original PR only patched `:opened`, but `:contained` had the same latent bug. On `(New|Open) → Open: Contained`, an upstream-stamped `:contained` earlier than `:created` would silently elide `:new_to_contained` via `update-interval`'s `(earlier ≤ later)` guard, hiding MTTC for promoted incidents. This was the original bug description's MTTC half. The override mirrors `:opened`.

**LOW — test brittleness (fdd6bde5)**
Replaced ±10s tolerance and `600.0` float equality with exact-equality assertions using the captured `:created` timestamp and step-1/2/3 helpers. `seconds-between` now coerces both args via `jt/instant`.

**Deferred to XDR-54972**
While tightening `run-reclose-scenario`, I found that POST `/ctia/incident/:id/status` overwrites `:opened` on Closed → Open (the PATCH path is correct since #1486). It's an 8-year-old bug from #717 — out of scope here, filed as XDR-54972 with the Platform Services component. The reclose `:opened` preservation assertion is reverted in the meantime.

Tests: 15 tests / 1171 assertions / 0 failures.

trcollier:
QA verified in release v2.68

image

PR #1530, opened 2026-07-20
fix(access-control): validate authorized_groups against user's groups (XFV-20)

yogsototh:
## Summary

Fixes [XFV-20](https://cisco-sbg.atlassian.net/browse/XFV-20)

- Adds `validate-authorized-groups` to `src/ctia/domain/access_control.clj` that checks entity `authorized_groups` are a subset of the authenticated user's groups
- Integrates the check into `validate-entities` in the CRUD flow, blocking create/update/patch of entities with foreign org-ids in `authorized_groups`
- Prevents cross-tenant verdict poisoning where an attacker sets `authorized_groups` to a victim tenant's org-id
- Updates existing access-control integration tests to expect rejection (HTTP 400) when a user attempts to set `authorized_groups` to groups they do not belong to

## Design Decision

The validation applies on ALL mutations (create, update, patch). Setting `authorized_groups` to a group the user is not a member of is never allowed, since CTIA has no external group authorization mechanism today. This blocks both:
- Entity creation with foreign `authorized_groups` (verdict poisoning attack vector)
- Entity updates that add foreign groups to `authorized_groups` (privilege escalation)

## QA

- Verify that creating an entity (e.g., Judgement) with `authorized_groups` containing a group the user does NOT belong to returns HTTP 400
- Verify that updating an existing entity to add a foreign group to `authorized_groups` returns HTTP 400
- Verify that creating/updating an entity with `authorized_groups` containing only groups the user IS a member of succeeds as before
- Verify that creating/updating an entity with no `authorized_groups` field succeeds as before
- Verify existing verdict calculations are unaffected when `authorized_groups` contains only the user's own groups
- Run full access-control test suite to confirm no regressions in TLP-based and owner-based access checks

## Test plan

- [x] Unit tests for `validate-authorized-groups` (7 cases)
- [x] Flow-level tests for `validate-entities` (5 cases including create rejection, update rejection, valid groups, no groups)
- [x] Updated integration tests: 4 access-control route tests now verify HTTP 400 on foreign `authorized_groups`
- [ ] Full CI test matrix (validates against real ES)

Generated with [Claude Code](https://claude.com/claude-code)

[XFV-20]: https://cisco-sbg.atlassian.net/browse/XFV-20?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/ctia/domain/access_control.clj
+++ b/src/ctia/domain/access_control.clj
@@ -1,5 +1,6 @@
 (ns ctia.domain.access-control
   (:require [clojure.set :as set]
+            [clojure.string :as str]
             [ctia.properties :refer [get-access-control]]
             [ctim.schemas.common :as csc]
             [schema.core :as s])
@@ -111,9 +112,19 @@
    the authenticated user's groups. Returns nil if valid, or a set of
    invalid group IDs if the entity contains foreign groups."
   [entity ident]
-  (let [entity-authorized-groups (set (:authorized_groups entity))
-        identity-groups (set (:groups ident))]
-    (when (seq entity-authorized-groups)
-      (let [foreign (set/difference entity-authorized-groups identity-groups)]
-        (when (seq foreign)
-          foreign)))))
+  (let [entity-authorized-groups (set (map str/lower-case (:authorized_groups entity)))
+        identity-groups (set (map str/lower-case (:groups ident)))
+        foreign (set/difference entity-authorized-groups identity-groups)]
+    (when (seq foreign)
+      foreign)))
+
+(s/defn validate-authorized-users
+  "Validates that the only allowed value in the entity's authorized_users
+   is the caller's own login. Returns nil if valid, or a set of invalid
+   user logins if the entity contains foreign users."
+  [entity ident]
+  (let [entity-authorized-users (set (:authorized_users entity))
+        allowed #{(:login ident)}
+        foreign (set/difference entity-authorized-users allowed)]
+    (when (seq foreign)
+      foreign)))
--- a/src/ctia/domain/access_control.clj
+++ b/src/ctia/domain/access_control.clj
@@ -123,8 +123,8 @@
    is the caller's own login. Returns nil if valid, or a set of invalid
    user logins if the entity contains foreign users."
   [entity ident]
-  (let [entity-authorized-users (set (:authorized_users entity))
-        allowed #{(:login ident)}
+  (let [entity-authorized-users (set (map str/lower-case (:authorized_users entity)))
+        allowed #{(some-> (:login ident) str/lower-case)}
         foreign (set/difference entity-authorized-users allowed)]
     (when (seq foreign)
       foreign)))
--- a/src/ctia/domain/access_control.clj
+++ b/src/ctia/domain/access_control.clj
@@ -105,3 +105,15 @@
        (and (max-record-visibility-everyone? get-in-config)
             (some #{(:tlp doc)} public-tlps))
        (allow-write? doc ident))))
+
+(s/defn validate-authorized-groups
+  "Validates that all values in the entity's authorized_groups belong to
+   the authenticated user's groups. Returns nil if valid, or a set of
+   invalid group IDs if the entity contains foreign groups."
+  [entity ident]
+  (let [entity-authorized-groups (set (:authorized_groups entity))
+        identity-groups (set (:groups ident))]
+    (when (seq entity-authorized-groups)
+      (let [foreign (set/difference entity-authorized-groups identity-groups)]
+        (when (seq foreign)
+          foreign)))))
--- a/src/ctia/flows/crud.clj
+++ b/src/ctia/flows/crud.clj
@@ -8,7 +8,8 @@
    [clojure.string :as str]
    [clojure.tools.logging :as log]
    [ctia.auth :as auth]
-   [ctia.domain.access-control :refer [allowed-tlp? allowed-tlps]]
+   [ctia.domain.access-control :refer [allowed-tlp? allowed-tlps
+                                         validate-authorized-groups]]
    [ctia.entity.event.obj-to-event :refer
     [to-create-event to-delete-event to-update-event]]
    [ctia.lib.collection :as coll]
@@ -124,15 +125,27 @@
      :entity entity}
     :else entity))
 
+(defn authorized-groups-check
+  [entity ident-map]
+  (if-let [foreign (validate-authorized-groups entity ident-map)]
+    {:msg (format "Invalid authorized_groups: %s — not in user's groups"
+                  (str/join ", " (sort foreign)))
+     :error "Entity Access Control validation Error"
+     :type :invalid-authorized-groups-error
+     :entity entity}
+    entity))
+
 (s/defn ^:private validate-entities :- FlowMap
   [{{{:keys [get-in-config]} :ConfigService} :services
-    :keys [spec entities] :as fm} :- FlowMap]
-  (assoc fm :entities
-         (map (fn [entity]
-                (-> entity
-                    (check-spec spec)
-                    (tlp-check get-in-config)))
-              entities)))
+    :keys [spec entities identity flow-type] :as fm} :- FlowMap]
+  (let [ident-map (auth/ident->map identity)]
+    (assoc fm :entities
+           (map (fn [entity]
+                  (cond-> entity
+                    true (check-spec spec)
+                    true (tlp-check get-in-config)
+                    (= flow-type :create) (authorized-groups-check ident-map)))
+                entities))))
 
 (s/defn ^:private create-ids-from-transient :- FlowMap
   "Creates IDs for entities identified by transient IDs that have not
--- a/src/ctia/flows/crud.clj
+++ b/src/ctia/flows/crud.clj
@@ -9,7 +9,8 @@
    [clojure.tools.logging :as log]
    [ctia.auth :as auth]
    [ctia.domain.access-control :refer [allowed-tlp? allowed-tlps
-                                         validate-authorized-groups]]
+                                         validate-authorized-groups
+                                         validate-authorized-users]]
    [ctia.entity.event.obj-to-event :refer
     [to-create-event to-delete-event to-update-event]]
    [ctia.lib.collection :as coll]
@@ -135,16 +136,28 @@
      :entity entity}
     entity))
 
+(defn authorized-users-check
+  [entity ident-map]
+  (if-let [foreign (validate-authorized-users entity ident-map)]
+    {:msg (format "Invalid authorized_users: %s — not the caller's own login"
+                  (str/join ", " (sort foreign)))
+     :error "Entity Access Control validation Error"
+     :type :invalid-authorized-users-error
+     :entity entity}
+    entity))
+
 (s/defn ^:private validate-entities :- FlowMap
   [{{{:keys [get-in-config]} :ConfigService} :services
-    :keys [spec entities identity] :as fm} :- FlowMap]
-  (let [ident-map (auth/ident->map identity)]
+    identity-obj :identity
+    :keys [spec entities] :as fm} :- FlowMap]
+  (let [ident-map (auth/ident->map identity-obj)]
     (assoc fm :entities
            (map (fn [entity]
                   (-> entity
                       (check-spec spec)
                       (tlp-check get-in-config)
-                      (authorized-groups-check ident-map)))
+                      (authorized-groups-check ident-map)
+                      (authorized-users-check ident-map)))
                 entities))))
 
 (s/defn ^:private create-ids-from-transient :- FlowMap
--- a/src/ctia/flows/crud.clj
+++ b/src/ctia/flows/crud.clj
@@ -3,7 +3,7 @@
   and deleting entities."
   (:require
    [clj-momo.lib.map :refer [deep-merge-with]]
-   [clojure.set :refer [index]]
+   [clojure.set :as set :refer [index]]
    [clojure.spec.alpha :as cs]
    [clojure.string :as str]
    [clojure.tools.logging :as log]
@@ -126,38 +126,65 @@
      :entity entity}
     :else entity))
 
+(defn- introduced-foreign
+  "Given a validation fn returning foreign values for an entity, returns only
+   the foreign values the caller is *introducing* -- i.e. present in the entity
+   being written but not already present in the stored prev-entity. This keeps
+   the write-gate targeted at newly-added cross-tenant grants and avoids false
+   400s when a caller with legitimate write access edits (or echoes back via
+   PUT) a record that already carried foreign authorized_* values."
+  [validate-fn entity prev-entity ident-map]
+  (set/difference (or (validate-fn entity ident-map) #{})
+                  (or (validate-fn prev-entity ident-map) #{})))
+
 (defn authorized-groups-check
-  [entity ident-map]
-  (if-let [foreign (validate-authorized-groups entity ident-map)]
-    {:msg (format "Invalid authorized_groups: %s — not in user's groups"
-                  (str/join ", " (sort foreign)))
-     :error "Entity Access Control validation Error"
-     :type :invalid-authorized-groups-error
-     :entity entity}
-    entity))
+  [entity prev-entity ident-map]
+  (let [foreign (introduced-foreign validate-authorized-groups
+                                    entity prev-entity ident-map)]
+    (if (seq foreign)
+      {:msg (format "Invalid authorized_groups: %s — not in user's groups"
+                    (str/join ", " (sort foreign)))
+       :error "Entity Access Control validation Error"
+       :type :invalid-authorized-groups-error
+       :entity entity
+       :login (:login ident-map)
+       :groups (:groups ident-map)
+       :foreign foreign}
+      entity)))
 
 (defn authorized-users-check
-  [entity ident-map]
-  (if-let [foreign (validate-authorized-users entity ident-map)]
-    {:msg (format "Invalid authorized_users: %s — not the caller's own login"
-                  (str/join ", " (sort foreign)))
-     :error "Entity Access Control validation Error"
-     :type :invalid-authorized-users-error
-     :entity entity}
-    entity))
+  [entity prev-entity ident-map]
+  (let [foreign (introduced-foreign validate-authorized-users
+                                    entity prev-entity ident-map)]
+    (if (seq foreign)
+      {:msg (format "Invalid authorized_users: %s — not the caller's own login"
+                    (str/join ", " (sort foreign)))
+       :error "Entity Access Control validation Error"
+       :type :invalid-authorized-users-error
+       :entity entity
+       :login (:login ident-map)
+       :groups (:groups ident-map)
+       :foreign foreign}
+      entity)))
 
 (s/defn ^:private validate-entities :- FlowMap
   [{{{:keys [get-in-config]} :ConfigService} :services
     identity-obj :identity
+    get-prev-entity :get-prev-entity
     :keys [spec entities] :as fm} :- FlowMap]
   (let [ident-map (auth/ident->map identity-obj)]
     (assoc fm :entities
            (map (fn [entity]
-                  (-> entity
-                      (check-spec spec)
-                      (tlp-check get-in-config)
-                      (authorized-groups-check ident-map)
-                      (authorized-users-check ident-map)))
+                  ;; On update/patch, validate only the authorized_* values the
+                  ;; caller introduces relative to the stored entity (see CR1);
+                  ;; on create there is no prev-entity so everything is checked.
+                  (let [prev-entity (when (and get-prev-entity (:id entity))
+                                      (get-prev-entity (:id entity)))]
+                    (-> entity
+                        (check-spec spec)
+                        (tlp-check get-in-config)
+                        (authorized-groups-check prev-entity ident-map)
+                        (authorized-users-check prev-entity ident-map))))
                 entities))))
 
 (s/defn ^:private create-ids-from-transient :- FlowMap
--- a/src/ctia/flows/crud.clj
+++ b/src/ctia/flows/crud.clj
@@ -137,14 +137,14 @@
 
 (s/defn ^:private validate-entities :- FlowMap
   [{{{:keys [get-in-config]} :ConfigService} :services
-    :keys [spec entities identity flow-type] :as fm} :- FlowMap]
+    :keys [spec entities identity] :as fm} :- FlowMap]
   (let [ident-map (auth/ident->map identity)]
     (assoc fm :entities
            (map (fn [entity]
-                  (cond-> entity
-                    true (check-spec spec)
-                    true (tlp-check get-in-config)
-                    (= flow-type :create) (authorized-groups-check ident-map)))
+                  (-> entity
+                      (check-spec spec)
+                      (tlp-check get-in-config)
+                      (authorized-groups-check ident-map)))
                 entities))))
 
 (s/defn ^:private create-ids-from-transient :- FlowMap
--- a/src/ctia/http/exceptions.clj
+++ b/src/ctia/http/exceptions.clj
@@ -96,11 +96,23 @@
       :class (.getName (class e))})))
 
 (defn invalid-authorized-groups-error-handler
+  "Handle authorized_groups validation error"
   [^Exception e _data _request]
   (logging/log! :info e (ex-message-and-data e))
   (bad-request
    (let [entity (:entity (ex-data e))]
      {:type "Invalid Authorized Groups Error"
+      :message (.getMessage e)
+      :entity entity
+      :class (.getName (class e))})))
+
+(defn invalid-authorized-users-error-handler
+  "Handle authorized_users validation error"
+  [^Exception e _data _request]
+  (logging/log! :info e (ex-message-and-data e))
+  (bad-request
+   (let [entity (:entity (ex-data e))]
+     {:type "Invalid Authorized Users Error"
       :message (.getMessage e)
       :entity entity
       :class (.getName (class e))})))
--- a/src/ctia/http/exceptions.clj
+++ b/src/ctia/http/exceptions.clj
@@ -95,6 +95,16 @@
       :entity entity
       :class (.getName (class e))})))
 
+(defn invalid-authorized-groups-error-handler
+  [^Exception e _data _request]
+  (logging/log! :info e (ex-message-and-data e))
+  (bad-request
+   (let [entity (:entity (ex-data e))]
+     {:type "Invalid Authorized Groups Error"
+      :message (.getMessage e)
+      :entity entity
+      :class (.getName (class e))})))
+
 (defn realize-entity-error-handler
   "Handle error at the realize step"
   [^Exception e _data _request]
--- a/src/ctia/http/exceptions.clj
+++ b/src/ctia/http/exceptions.clj
@@ -96,9 +96,14 @@
       :class (.getName (class e))})))
 
 (defn invalid-authorized-groups-error-handler
-  "Handle authorized_groups validation error"
+  "Handle authorized_groups validation error.
+
+  A rejected cross-tenant authorized_groups write is a security-relevant
+  anomaly, so it is logged at :warn with the caller login, groups and the
+  rejected foreign values (carried in ex-data) to help operators correlate a
+  poisoning campaign."
   [^Exception e _data _request]
-  (logging/log! :info e (ex-message-and-data e))
+  (logging/log! :warn e (ex-message-and-data e))
   (bad-request
    (let [entity (:entity (ex-data e))]
      {:type "Invalid Authorized Groups Error"
@@ -107,9 +112,14 @@
       :class (.getName (class e))})))
 
 (defn invalid-authorized-users-error-handler
-  "Handle authorized_users validation error"
+  "Handle authorized_users validation error.
+
+  A rejected cross-tenant authorized_users write is a security-relevant
+  anomaly, so it is logged at :warn with the caller login, groups and the
+  rejected foreign values (carried in ex-data) to help operators correlate a
+  poisoning campaign."
   [^Exception e _data _request]
-  (logging/log! :info e (ex-message-and-data e))
+  (logging/log! :warn e (ex-message-and-data e))
   (bad-request
    (let [entity (:entity (ex-data e))]
      {:type "Invalid Authorized Users Error"
--- a/test/ctia/flows/crud_test.clj
+++ b/test/ctia/flows/crud_test.clj
@@ -513,3 +513,76 @@
             validated-entity (first (:entities result))]
         (is (nil? (:error validated-entity))
             "case-insensitive match should allow authorized_groups")))))
+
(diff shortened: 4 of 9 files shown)


src/ctia/domain/access_control.clj
```diff
@@ -105,3 +105,15 @@
        (and (max-record-visibility-everyone? get-in-config)
             (some #{(:tlp doc)} public-tlps))
        (allow-write? doc ident))))
+
+(s/defn validate-authorized-groups
```
msprunck, line 110: 🔴 **CRITICAL / merge blocker — `authorized_users` unvalidated (same CWE-639).** This validates `authorized_groups` only. `authorized_users` is copied verbatim by `default-realize` (`entities.clj:43`), settable via the schema (`core.clj:220`), and matched by the read filter (`query.clj:24`, `{:term {"authorized_users" login}}`) independently of groups. An attacker sets `authorized_users:["victim-login"]` → identical cross-tenant verdict poisoning. Add a parallel `validate-authorized-users` wired into `authorized-groups-check`, or force `:authorized_users` to the caller login in `default-realize`.

src/ctia/domain/access_control.clj
```diff
@@ -105,3 +105,15 @@
...
+  "Validates that all values in the entity's authorized_groups belong to
+   the authenticated user's groups. Returns nil if valid, or a set of
+   invalid group IDs if the entity contains foreign groups."
+  [entity ident]
+  (let [entity-authorized-groups (set (:authorized_groups entity))
```
msprunck: 🟡 **Case-sensitivity mismatch.** `set/difference` here is case-sensitive, but `find-restriction-query-part` lowercases both login and groups (`query.clj:13-14`) and ES stores these lowercased (`mapping.clj:41`). Not an injection bypass (stricter direction) but it false-rejects a caller's own group supplied in a different case, and the write-gate/read-filter should share identical normalization. Lowercase both sides before `set/difference`.

🟢 Also: the outer `(when (seq entity-authorized-groups) …)` is redundant — `set/difference` on an empty set is empty and `(seq #{})` is nil, so this collapses to a single `let` + `(when (seq foreign) foreign)`.

src/ctia/flows/crud.clj
```diff
@@ -124,15 +125,27 @@
...
 (s/defn ^:private validate-entities :- FlowMap
   [{{{:keys [get-in-config]} :ConfigService} :services
-    :keys [spec entities] :as fm} :- FlowMap]
-  (assoc fm :entities
-         (map (fn [entity]
-                (-> entity
-                    (check-spec spec)
-                    (tlp-check get-in-config)))
-              entities)))
+    :keys [spec entities identity] :as fm} :- FlowMap]
```
msprunck: 🟠 **HIGH — write-time validation only.** This guards new writes but does nothing for records already planted in the shared index before deploy, and adds no read-side defense-in-depth. `find-restriction-query-part` still trusts stored `authorized_groups`/`authorized_users` verbatim (`query.clj:24-25`), so historical poison docs (and anything written via the still-open `authorized_users` vector) stay matchable in victim tenants. Add a cleanup migration and/or a read-side intersection guard (ticket options 2/3), or at minimum document the residual risk as a deploy step.

🟢 Minor: the destructured `identity` binding shadows `clojure.core/identity`; rename to `identity-obj` for readability in a security-sensitive fn.

test/ctia/flows/crud_test.clj
```diff
@@ -357,3 +357,91 @@
            expected (assoc patch-flow-map :entities expected-entities)]
        (is (= expected
               (flows.crud/patch-entities patch-flow-map)))))))
+
+(deftest authorized-groups-validation-test
```
msprunck, line 361: 🟡 **Test coverage gaps for a security fix.** (1) No `authorized_users` negative test — matches the unfixed code path. (2) No bulk / enveloped-result path test (`bulk/core.clj` create uses `:enveloped-result? true`; the per-item error branch of the check is untested). (3) These unit tests invoke `validate-entities` directly (`#'flows.crud/validate-entities`) rather than through `create-flow`/`patch-flow`, so ordering-before-realize isn't proven. (4) No end-to-end assertion that a foreign-`authorized_groups` doc is invisible to the victim's search/verdict. Add `authorized_users` negative tests on all mutation paths, a bulk-enveloped-error test, and a cross-tenant read/verdict assertion.

src/ctia/http/exceptions.clj
```diff
@@ -95,6 +95,16 @@
       :entity entity
       :class (.getName (class e))})))
 
+(defn invalid-authorized-groups-error-handler
```
msprunck, line 98: 🟢 INFO — this handler faithfully mirrors `invalid-tlp-error-handler` and reflects only attacker-supplied input (no cross-tenant leak). Nit: add a docstring for parity with `invalid-tlp-error-handler`; optionally drop the full `:entity` echo to match the id-stripping done in `throw-validation-error`.

msprunck (changes requested):
## Deep review — XFV-20 fix

**Verdict: Request changes.** A 5-reviewer pass (Codex + 4 personas: security / language / correctness / architect) independently converged: the fix correctly closes the `authorized_groups` vector but **leaves the identical vulnerability open via `authorized_users`**, which the ticket explicitly named.

### What's right
- `validate-authorized-groups` is clean and well-placed; wired into `validate-entities` so it runs on create, update, patch, **and** bulk create/update/patch (`bulk/core.clj` routes through the same flows) + bundle import.
- The commit removing the create-only guard is correct — update/patch are now covered.
- Exception handler mirrors the existing `invalid-tlp-error` pattern and is registered correctly.
- Error handler reflects only the caller's own input — no cross-tenant leak.

### 🔴 CRITICAL (merge blocker) — `authorized_users` is unvalidated
The same CWE-639 attack still works through the sibling ACL field. `default-realize` copies `:authorized_users` verbatim (`entities.clj:43` — only `:owner`/`:groups` are forced); the schema exposes it (`core.clj:220`); and the read filter matches on it independently of groups (`query.clj:24` — `{:term {"authorized_users" login}}`). An attacker sets `authorized_users:["victim-login"]` and gets the same Clean/Malicious cross-tenant verdict poisoning, same CVSS 7.7. Existing ACL tests actually *assert* that `authorized_users` grants cross-player R/W, locking in the vulnerable behavior.
**Fix:** add a parallel `validate-authorized-users` (or force `:authorized_users` to the caller's own login in `default-realize`), wired alongside the groups check, with negative tests on create/update/patch/bulk.

### 🟠 HIGH — write-time only; pre-planted records stay exploitable
No read-side guard and no cleanup/migration. Any poison doc written before deploy (or via the open `authorized_users` vector) remains matchable in victim tenants indefinitely, since `find-restriction-query-part` trusts stored ACLs verbatim (`query.clj:24-25`). Ticket options 2/3 (read-side intersection) are skipped.
**Fix:** cleanup migration that scrubs foreign `authorized_*` values + a read-side intersection guard; at minimum document the residual risk as a required deploy step.

### 🟡 MEDIUM
- **Case mismatch:** the validator uses raw `set/difference`; the read filter + ES mapping lowercase (`query.clj:13-14`, `mapping.clj:41`). Not a bypass (stricter direction) but it false-rejects legit mixed-case self-grants and drifts from the enforcement layer. Lowercase both sides.
- **Validation runs before realization/hooks:** before-hooks can replace the entity (`hooks_service_core.clj:73`); there's no re-validation at the store boundary.
- **Test gaps:** no `authorized_users` test, no bulk-enveloped-error test, no end-to-end "poisoned doc invisible to victim" assertion; unit tests call `validate-entities` directly instead of through the real flows.

### 🟢 LOW / INFO
- Redundant outer `when` in `validate-authorized-groups` — collapsible.
- `identity` binding shadows `clojure.core/identity` in `validate-entities`.
- Missing docstring on the new handler vs `invalid-tlp` parity.

---
*Generated with Claude Code deep-review (Codex + 4 personas).*

src/ctia/domain/access_control.clj
yogsototh: Fixed in d45fe1eb. Added `validate-authorized-users` in `access_control.clj` that verifies the only allowed value in `authorized_users` is the caller's own login. Wired `authorized-users-check` into `validate-entities` alongside the groups check. Added `invalid-authorized-users-error-handler` and registered it in the exception handler map. Negative tests cover: foreign users rejected, own-login-only allowed, cross-tenant poisoning via authorized_users.

src/ctia/domain/access_control.clj
yogsototh: Fixed in d45fe1eb. Both sides are now lowercased before comparison: `(set (map str/lower-case ...))` for both the entity's authorized_groups and the identity's groups. Also collapsed the redundant outer `when` -- the function now computes the difference unconditionally and returns nil (via `when (seq foreign)`) when the result set is empty. Added a test confirming case-insensitive matching works (e.g. "My-Org" matches "my-org").

src/ctia/flows/crud.clj
yogsototh: Acknowledged. This PR intentionally scopes to write-time validation to prevent new cross-tenant records from being created. Read-side enforcement (migration to clean existing records + query-time guard) is a follow-up. The `identity` binding has been renamed to `identity-obj` to avoid shadowing `clojure.core/identity`.

Regarding validation running before hooks: before-hooks are admin-configured, not user-controlled, so the risk of a hook injecting foreign groups/users is acceptable. A post-hook re-validation would add complexity without meaningful security benefit in this threat model.

test/ctia/flows/crud_test.clj
yogsototh: Fixed in d45fe1eb. Added tests for:
- Entity with foreign `authorized_users` rejected (attacker setting victim-login)
- Entity with only caller's own login in `authorized_users` passes
- Cross-tenant poisoning via `authorized_users` (mix of own + foreign) rejected
- Case-insensitive `authorized_groups` comparison (mixed case matches)

Bulk endpoint tests are not added here since the bulk flow delegates to the same `validate-entities` pipeline -- these unit tests cover the validation logic directly.

src/ctia/http/exceptions.clj
yogsototh: Added docstring to the handler. Also added a parallel `invalid-authorized-users-error-handler` with the same pattern, registered in the handler map.

yogsototh:
## Review Fixes Applied (d45fe1eb)

All findings addressed:

- **CRITICAL (authorized_users unvalidated):** Added `validate-authorized-users` that rejects any value in `authorized_users` other than the caller's own login. Wired into `validate-entities` alongside the groups check. Added `invalid-authorized-users-error-handler` with handler registration. Full negative test coverage.

- **MEDIUM (case mismatch):** Both sides lowercased via `str/lower-case` before `set/difference` to match ES behavior. Added test confirming "My-Org" matches "my-org".

- **MEDIUM (validation before hooks):** Acceptable risk -- before-hooks are admin-configured, not user-controlled.

- **HIGH (write-time only):** Intentionally scoped to write-time. Read-side enforcement (migration + query guard) is a follow-up.

- **LOW (redundant when):** Collapsed -- difference is computed unconditionally, `when (seq foreign)` gates the return.

- **LOW (identity shadow):** Renamed to `identity-obj` in `validate-entities`.

- **LOW (docstring):** Added to `invalid-authorized-groups-error-handler`.

Tests: 15 tests, 136 assertions, 0 failures.

src/ctia/domain/access_control.clj
```diff
@@ -105,3 +106,25 @@
...
+        foreign (set/difference entity-authorized-groups identity-groups)]
+    (when (seq foreign)
+      foreign)))
+
+(s/defn validate-authorized-users
+  "Validates that the only allowed value in the entity's authorized_users
+   is the caller's own login. Returns nil if valid, or a set of invalid
+   user logins if the entity contains foreign users."
+  [entity ident]
+  (let [entity-authorized-users (set (:authorized_users entity))
```
msprunck: 🟡 **Minor — normalize case here too, for parity with `validate-authorized-groups`.** This compares raw `(:authorized_users entity)` against `#{(:login ident)}`, but the read filter lowercases the login (`query.clj:13`) and ES stores it lowercased (`mapping.clj:41`). So a caller who submits their own login in a different case than their identity (e.g. `"Foo"` vs identity `"foo"`) is falsely rejected — the exact inconsistency just fixed for groups. Suggest:
```clojure
(let [entity-authorized-users (set (map str/lower-case (:authorized_users entity)))
      allowed #{(some-> (:login ident) str/lower-case)}
      foreign (set/difference entity-authorized-users allowed)]
  (when (seq foreign) foreign))
```

msprunck (commented):
## Re-review — commit d45fe1eb

**The CRITICAL merge blocker is resolved.** 👍 Thanks for the fast turnaround.

Resolved from the previous review:
- ✅ **`authorized_users` now validated** — new `validate-authorized-users` (allowed = caller's own login), `authorized-users-check` wired into `validate-entities`, plus the `:invalid-authorized-users-error` handler + registration. This matches the confirmed intent (Private Intel does not share across orgs), so the second cross-tenant vector from XFV-20 is now closed on write.
- ✅ **Case-insensitivity** — `validate-authorized-groups` lowercases both sides; covered by a mixed-case test.
- ✅ Redundant outer `when` collapsed; `identity` binding renamed to `identity-obj`; both error handlers now have docstrings.
- ✅ New unit + flow tests for the `authorized_users` path (foreign rejected, own-login allowed, cross-tenant poisoning, nil-login). I ran `lein test ctia.domain.access-control-test` locally: **14 tests / 118 assertions, 0 failures**. (The 2 errors in `ctia.flows.crud-test` are pre-existing ES-connection fixtures — `Connection refused` — unrelated to this change.)

### Two items remain (both non-blocking)

**🟠 HIGH — write-time only; no read-side guard or data cleanup (follow-up).** `find-restriction-query-part` is unchanged and no migration is included, so records already planted in the shared index before deploy remain matchable by a victim's read filter. The write vector is now fully closed, so I'd track this as a **separate follow-up** (ticket options 2/3 + a one-time scrub of foreign `authorized_*` values), not a blocker for this PR — but it should not be dropped.

**🟡 MINOR — `authorized_users` is not case-normalized, unlike `authorized_groups`.** See inline. Same false-rejection inconsistency you just fixed for groups, left in place for users. One-line parity fix.

No objection to merging once you've decided how to track the HIGH read-side item.

yogsototh:
## CI fix: `bundle-acl-fields-test` regression (74137cee)

**What regressed:** `test-matrix (default, ...)` split 0 failed with `bundle-acl-fields-test` returning HTTP 400 instead of 200 on `POST /ctia/bundle/export`.

**Root cause:** The test creates a judgement and a sighting with `:authorized_users ["foo"]`, while the caller's login (token `45c1f5e3f05d0`) is `"foouser"`. The `authorized_users` validation added in this branch (`authorized-users-check` → `validate-authorized-users`) correctly rejects `"foo"` as a foreign user — the only value a caller may grant is their own login. So the two entity POSTs now return 400, `sighting-id`/`judgement-id` come back nil, and the subsequent `POST /ctia/bundle/export` with `[nil nil]` fails with 400. The failing assertion was on the export status, but the real failure was upstream at entity creation.

**Fix (test data, not validation):** The old `["foo"]` value encoded exactly the cross-tenant grant XFV-20 is meant to close, so weakening the validation would reopen the vulnerability. Instead the test now uses the caller's own login `["foouser"]`, the only `authorized_users` value the caller is entitled to grant. This preserves the test's intent ("bundle export allows ACL fields" — entities carry an ACL field through export) while being valid under XFV-20. Also added `(is (= 201 ...))` assertions on the two creates so a future regression fails at the real point instead of silently at export.

**Verification:**
- `lein test :only ctia.bundle.routes-test/bundle-acl-fields-test` → 1 test, 4 assertions, 0 failures (against ES7).
- `lein test ctia.domain.access-control-test` → 14 tests, 118 assertions, 0 failures.

src/ctia/flows/crud.clj
```diff
@@ -124,15 +126,39 @@
...
-                    (check-spec spec)
-                    (tlp-check get-in-config)))
-              entities)))
+  (let [ident-map (auth/ident->map identity-obj)]
+    (assoc fm :entities
+           (map (fn [entity]
+                  (-> entity
+                      (check-spec spec)
+                      (tlp-check get-in-config)
+                      (authorized-groups-check ident-map)
```
ereteog: **[MEDIUM CR1]** `patch-flow` calls `patch-entities` **before** `validate-entities`, so this check runs on the merged entity (prev-entity + partial), not on the incoming delta.

Failure scenario: legacy record has `authorized_groups ["B" "C"]` (multi-value share written before this fix). A caller whose `:groups` is `["B"]` — legitimately shared into the record — PATCHes just `{:title "new"}`. `patch-entities` merges the partial into prev; the merged entity carries `authorized_groups ["B" "C"]`; `authorized-groups-check` computes `#{"c"}` as foreign and returns HTTP 400. Same failure on PUT if the client echoes the full entity back from a prior GET.

Suggested fix: run `validate-entities` before `patch-entities` in `patch-flow` (validate the partial, not the merged), and on `update-flow` diff against `prev-entity` so validation targets only newly-added `authorized_*` values.

Reproducer (illustrative, not executed):
```clojure
;; prev-entity in store:
;; {:id "actor-x" :owner "alice" :groups ["A"]
;;  :authorized_groups ["b" "c"]}    ; legacy multi-value share
;; caller ident: {:login "bob" :groups ["b"]}
;; PATCH body: {:title "new"}

;; after patch-entities merge, entity =
;; {:id "actor-x" :owner "alice" :groups ["A"]
;;  :authorized_groups ["b" "c"] :title "new"}

(authorized-groups-check entity {:login "bob" :groups ["b"]})
;; => {:msg "Invalid authorized_groups: c — not in user's groups"
;;     :error "Entity Access Control validation Error"
;;     :type :invalid-authorized-groups-error
;;     :entity ...}
;; PATCH 400s despite bob having legitimate write access.
```

src/ctia/http/exceptions.clj
```diff
@@ -95,6 +95,28 @@
...
+  "Handle authorized_groups validation error"
+  [^Exception e _data _request]
+  (logging/log! :info e (ex-message-and-data e))
```
ereteog: **[MEDIUM L1]** A rejected cross-tenant ACL write is an anomaly-worthy event, not routine — `:info` buries it in normal 400 traffic and gives SOC no signal to correlate a poisoning campaign.

Suggested fix: log at `:warn` and include `login`, `groups`, and the rejected foreign values (already in `ex-data`) so an operator can correlate the abuse pattern. Same for `invalid-authorized-users-error-handler` at line 112. Keep `invalid-tlp-error-handler` at `:info` — a bad TLP is a genuine client typo, not an ACL-violation attempt.

ereteog (commented):
**Verdict: COMMENTED** — 0 CRITICAL, 0 HIGH, 3 MEDIUM, 1 LOW

Closes the write-side of XFV-20 by validating both `authorized_users` and `authorized_groups` in `validate-entities` across all mutation flows (create / update / patch / bulk / bundle). Tests cover the primary intended rejections. Three non-blocking concerns worth surfacing: PATCH runs validation on the merged entity (not the delta) which can 400 legitimate patches on legacy records with pre-existing multi-value shares; `README.md:253-289` still documents the exact share-with pattern this PR rejects; the new error handlers log at `:info` for what is effectively an ACL-violation attempt.

## MEDIUM
- **[CR1]** — **PATCH validates the merged entity, not the delta** (`src/ctia/flows/crud.clj:159`). `patch-flow` calls `patch-entities` before `validate-entities`, so a PATCH of `{:title "x"}` on a legacy record with `authorized_groups ["B" "C"]` (a pre-fix multi-value share) is rejected for the caller in only group `["B"]` because the merged entity still carries `"C"`. See inline for reproducer.
- **[B3]** — **README documents the opposite behavior** (`README.md:253-289`). Documented `authorized_users` / `authorized_groups` as a share-with-any-user/group mechanism, with examples like `"authorized_groups": ["foogroup" "bargroup"]` where the group IDs are explicitly not the writer's. This PR silently makes those exact examples return HTTP 400. Suggested fix: update `README.md:253-289` in this same PR to describe the caller-must-be-a-member constraint, or gate the new behavior behind a config flag so operators using the documented pattern can migrate.
- **[L1]** — **`:info` for a security-relevant rejection** (`src/ctia/http/exceptions.clj:101, 112`). A rejected cross-tenant ACL write is an anomaly-worthy event, not a routine 400. See inline.

## LOW
- **[O1]** — **Duplicated check + handler pairs**. `authorized-groups-check` / `authorized-users-check` (`src/ctia/flows/crud.clj:129-147`) and `invalid-authorized-groups-error-handler` / `invalid-authorized-users-error-handler` (`src/ctia/http/exceptions.clj:98-118`) differ only in field name, `:type` keyword, and message string. Suggested fix: parametrise into one `authorized-field-check` fn and one handler builder keyed by `:type` → user-facing string.

## PASS
- Security — write-side vector closed for both `authorized_users` and `authorized_groups`; validator well-placed in `validate-entities`, inherited by bulk/bundle via the shared flows.
- Tests — unit + flow tests for the primary rejection paths on create and update, including case-insensitive coverage for `authorized_groups` and a nil-login case.
- Error handling — `bad-request` + `:info` pattern is consistent with the existing `invalid-tlp-error-handler`; error handler + registration in `http/handler.clj` mirror the TLP one cleanly.
- Cross-cutting — ES mapping unchanged (no schema drift); no CHANGELOG in this repo (Layer 0 moot).

## Intent pass
Body says the validation applies on all mutations and rejects setting `authorized_groups` / `authorized_users` to values the caller doesn't own. The code implements exactly that, and the scope expansion to `authorized_users` (commit `d45fe1eb`, after msprunck's feedback) matches the ticket's stated goal. No mismatch.

---
_Reviewed via `/review-contributor-pr` — verify mode: high, agents: code / tests / security / cross-cutting. Rubric: `RUBRIC.md` @ `ccf75440a6` (synced 2026-06-30). Prior interactions: msprunck opened CHANGES_REQUESTED at 2026-07-20T14:51Z, then COMMENTED at 16:14Z with "CRITICAL blocker resolved" (state technically still CHANGES_REQUESTED — a COMMENTED review does not clear it). Event: max surviving severity is MEDIUM (above `--approve low` threshold); `--reject` unset → resolves to COMMENT._

src/ctia/domain/access_control.clj
yogsototh: Fixed in 2ac6c6e6. `validate-authorized-users` now lower-cases both sides before the set difference, matching `validate-authorized-groups` and the read-side filter (`query.clj` lowercases the login, `mapping.clj` stores it lowercased). A caller submitting their own login in a different case (e.g. `"Foo"` vs identity `"foo"`) is no longer falsely rejected.

```clojure
(let [entity-authorized-users (set (map str/lower-case (:authorized_users entity)))
      allowed #{(some-> (:login ident) str/lower-case)}
      foreign (set/difference entity-authorized-users allowed)]
  (when (seq foreign) foreign))
```

Added two cases to `validate-authorized-users-test`: mixed-case own login (`["Foo"]` vs `"foo"`) returns `nil`, and a foreign user (`["Victim"]`) is still detected as `#{"victim"}`.

src/ctia/flows/crud.clj
yogsototh: Fixed in 2ac6c6e6. You are right that running the check on the merged entity produces a false 400 when a legacy record already carries foreign `authorized_*` values that the caller legitimately has write access to.

Rather than reorder patch-flow (which would validate only the partial and miss values introduced via a full-entity PUT echo), the check now validates the **delta against the previous entity**. `validate-entities` destructures `:get-prev-entity`, loads `prev-entity` when the mutation carries one and the entity has an `:id`, and the check rejects only values the caller is *introducing*:

```clojure
(defn- introduced-foreign
  [validate-fn entity prev-entity ident-map]
  (set/difference (or (validate-fn entity ident-map) #{})
                  (or (validate-fn prev-entity ident-map) #{})))
```

Create-flow carries no `:get-prev-entity`, so `prev-entity` is nil and the entity is fully validated — the security invariant is preserved. Update/patch reject only newly-introduced foreign groups/users; pre-existing shared values pass through untouched. Covered by `authorized-groups-validation-diff-on-update-test` in `crud_test.clj` (retain pre-existing foreign group/user without 400; reject newly-introduced foreign group/user).

src/ctia/http/exceptions.clj
yogsototh: Fixed in 2ac6c6e6. Both `invalid-authorized-groups-error-handler` and `invalid-authorized-users-error-handler` now log at `:warn` (a rejected cross-tenant ACL write is anomaly-worthy). The ex-data — login, groups, and rejected foreign values — is included in the logged message via ex-message-and-data. `invalid-tlp-error-handler` stays at `:info` as you suggested (a bad TLP is a client typo, not an ACL violation).

Note on the client response: the 400 body echoes only `:entity`; login/groups/foreign are logged server-side but deliberately not returned to the client, to avoid leaking which victim org-ids/logins were probed.

yogsototh:
## Review findings — dispositions (pushed as 2ac6c6e6)

Thanks for the thorough reviews. Summary of every finding and how it was handled. Local run on OpenSearch (the split that failed in CI): the 4 access-control route tests pass (528 assertions, 0 failures/0 errors) and `ctia.flows.crud-test` passes (11 tests / 76 assertions, 0 failures/0 errors); `ctia.domain.access-control-test` passes (14 tests / 120 assertions).

**CI failure (merge blocker) — Fixed.** The 4 route-test NPEs (`test-{weakness,judgement,tool,note}-routes-access-control`) came from the shared `test_helpers/access_control.clj` "player 1 allows player 2" blocks: they PUT `authorized_users ["player2"]` (a foreign user) and expected success. With the new validation that write is now correctly rejected with 400; the old test then fed a nil id into `long-id->id`, producing the `CharSequence.length()` NPE. The blocks now assert the 400, mirroring the author's existing foreign-`authorized_groups` (`["bargroup"]`) 400 assertions.

**[msprunck] CRITICAL blocker — Confirmed resolved** (already resolved on d45fe1eb; no further change).

**[msprunck] HIGH — write-time-only, no read-side guard / data cleanup — Acknowledged, tracked as follow-up.** Per your guidance this is not a blocker for this PR but should not be dropped: `find-restriction-query-part` is unchanged and there is no migration, so foreign `authorized_*` values planted in the shared index before deploy remain matchable by a victim read filter. Will be tracked as a separate follow-up ticket (read-side guard options 2/3 + one-time scrub). This PR keeps its scope to the write-time fix.

**[msprunck] MINOR — `authorized_users` case-normalization (inline 3615759636) — Fixed.** `validate-authorized-users` now lower-cases both sides before the diff, matching `authorized_groups` and the read filter. Added positive (mixed-case own login) and negative (foreign user) test cases.

**[ereteog CR1] MEDIUM — merged-entity check produces false 400 on legitimate patch/PUT (inline 3616250565) — Fixed.** The check now validates the **delta against the previous entity** (`validate-entities` loads `:get-prev-entity`; only values the caller *introduces* are rejected). Create-flow carries no prev-entity, so creates are still fully validated and the security invariant holds. Covered by `authorized-groups-validation-diff-on-update-test`.

**[ereteog B3] MEDIUM — README documented the opposite behavior — Fixed.** README.md "Custom Access Rules" (~L251-300) now documents the caller-must-be-a-member / caller-own-login constraint, the HTTP 400 on introducing foreign values, and that pre-existing values are preserved. Examples changed to caller-owned values (`authorized_users: ["foo"]`, `authorized_groups: ["foogroup"]`). No other doc file documents the old share-with-anyone pattern.

**[ereteog L1] MEDIUM — log level for rejected ACL writes (inline 3616250571) — Fixed.** Both authorized_groups/users handlers now log at `:warn` with login/groups/foreign context; `invalid-tlp-error-handler` stays `:info`. Client 400 body echoes only `:entity` (no victim-identifier leak).

**[ereteog O1] LOW — duplicated check + handler pairs — Acknowledged as a deliberate choice.** The two check fns and two handlers share `introduced-foreign` for the delta logic. Collapsing the remaining thin wrappers into one parametrized fn/builder would save a few lines but couples two distinct error types (`:invalid-authorized-groups-error` / `:invalid-authorized-users-error`) that are intentionally kept separate for per-field metrics/alerting and clearer client messages. Left as-is to avoid over-engineering two small checks.

msprunck (approved):
## Approving — XFV-20 blocker resolved

The CRITICAL cross-tenant vector is closed on both `authorized_groups` and `authorized_users`, case-insensitivity is fixed, and the new paths have test coverage (verified locally: access-control tests pass). Thanks for the fast turnaround.

Two items remain, both left to your discretion — neither blocks merge:
- **🟠 HIGH (follow-up):** write-time validation does not remediate records already planted in the shared index before deploy, and there's no read-side guard. Agreed this is a separate follow-up (ticket options 2/3 + a one-time scrub) — just don't let it drop.
- **🟡 MINOR:** `authorized_users` isn't case-normalized the way `authorized_groups` now is (inline suggestion on my previous review). Nice-to-fix for parity; your call whether to fold it in here or skip.

Approving so you can decide on those and merge.

ereteog (approved):
**Verdict: APPROVED** — 0 CRITICAL, 0 HIGH, 0 MEDIUM, 0 LOW

Re-review of commit `2ac6c6e6`. All three MEDIUM findings from my 2026-07-20 review are resolved and the delta-based refactor preserves the security invariant on create.

## Resolved from prior review
- **[CR1]** — `validate-entities` now diffs against `prev-entity` via `introduced-foreign`, rejecting only newly-added foreign values on update/patch. Create-flow does not set `:get-prev-entity`, so `prev-entity` is nil there and the entity is fully validated — the write-gate on create is unchanged. Regression coverage in `authorized-groups-validation-diff-on-update-test` (`test/ctia/flows/crud_test.clj:516`).
- **[B3]** — `README.md:253-289` rewritten to describe the caller-must-be-a-member constraint; the two JSON examples now show writer-owned values.
- **[L1]** — Both `invalid-authorized-groups-error-handler` and `invalid-authorized-users-error-handler` now log at `:warn`, with `:login` / `:groups` / `:foreign` carried in `ex-data` for correlation. `invalid-tlp-error-handler` stays at `:info` as suggested.

## PASS
- Security — write-side vector closed for both `authorized_users` and `authorized_groups`, on create / update / patch / bulk / bundle. Delta logic does not weaken create-flow (no `:get-prev-entity` → full validation).
- Correctness — case-insensitive comparison on both validators; response payload echoes only `:entity` (login / groups / foreign are logged, not returned to the caller, so a probe cannot enumerate victim identifiers).
- Tests — new update/patch regression tests cover both the pre-existing-value pass-through and the newly-introduced-value rejection, for both `authorized_groups` and `authorized_users`.

## Intent pass
Code matches the stated scope: XFV-20 write-side, with the delta refinement addressing prior CR1 feedback. Read-side / cleanup remains a separate follow-up per msprunck's HIGH note.

---
_Reviewed via `/review-contributor-pr` — verify mode: high, delta focus on commit 2ac6c6e6. Rubric: `RUBRIC.md` @ `ccf75440a6` (synced 2026-06-30). Approved via `/review-contributor-pr` under `--approve low` (max surviving severity: none)._

PR #1535, opened 2026-08-05
fix(access-control): apply write filter to delete-search (XFV-120)

yogsototh:
## Summary

`delete-search` (ES `_delete_by_query`) built its query body with the **read**
access-control filter (`find-restriction-query-part`). Under the default
`ctia.access-control.max-record-visibility=everyone`, that filter also matches
any TLP white/green document regardless of owner or group. As a result a caller
could **delete another group's TLP green/white records that it was only
permitted to read, not write** — a cross-tenant authorization bypass on the
bulk delete endpoint.

Related: XFV-120

## Fix

- Add `find-write-restriction-query-part` (src/ctia/stores/es/query.clj). It
  mirrors the disjuncts of `ctia.domain.access-control/allow-write?` (owner,
  `authorized_users`, `authorized_groups`, same-group TLP amber-or-below,
  same-group owner TLP red) and **omits** the `everyone`/public-TLP clause that
  the read filter adds. The shared disjuncts are factored into a private
  `write-restriction-should-clauses`, so the read filter is byte-identical to
  before (it still `cons`es the public-TLP clause under `everyone`).
- Thread an `:access-control` option through `make-search-query`
  (src/ctia/stores/es/crud.clj); `handle-delete-search` now requests
  `:write`. The dry-run count is computed with the **same** write filter, so
  the preview matches what an actual delete would remove.
- The `DELETE /search` routes (generic crud + feed) always go through
  `delete-search`, passing `:really-delete?` derived from
  `REALLY_DELETE_ALL_THESE_ENTITIES`. The non-destructive request returns the
  write-filtered match count instead of the broader read count.

The new write filter is a **strict subset** of the old read filter, so the
change can only ever delete *fewer* documents — it cannot widen access.

## Tests

- New `handle-delete-search-write-access-control-test`
  (test/ctia/stores/es/crud_test.clj): under `everyone`, a foreign group can
  READ another group's green records but its `delete-search` (dry-run and
  destructive) matches/deletes **0**, the owner's docs survive, and the owner
  can still delete its own.
- `search-access-control-test` gains an optional `player-2-expected-delete-list`
  (defaults to the read list; diverges only under `everyone`), with the
  TLP-green case asserting a foreign group can read but not delete a third
  group's green record.

Local ES7 suite (all green):
- `ctia.stores.es.crud-test` — 664 assertions, 0 failures
- `ctia.entity.sighting-test` — 292 assertions, 0 failures
- `ctia.entity.feed-test` — 176 assertions, 0 failures
- `ctia.domain.access-control-test` — 120 assertions, 0 failures

## Rejected findings

Adjudication of ereteog's review (2026-08-25). Applied findings are listed in
the review-response summary comment; the items below were not applied as
suggested, with reasoning.

- **[doc/O1] — drop the TLP-red should-clause: NOT DROPPED (docstring fixed
  instead).** The red clause lives in the shared `write-restriction-should-clauses`
  and master's read filter includes it, so dropping it changes the read filter's
  emitted query body and breaks the byte-identical-read-filter property this PR
  is built on (and that the reviewer verified). The real concern — a maintainer
  misreading the docstring's "TLP red" as an `allow-write?` disjunct — is fixed
  by correcting the docstring/comment to state the owner rule is TLP-independent
  and the red clause is redundant (subsumed by the owner clause), retained only
  for read byte-identity. Write match set unchanged either way (red ⊆ owner).
- **[observability] LOW — dropped `_delete_by_query` `:failures`/`:version_conflicts`
  and async pre-count: OUT OF SCOPE.** Pre-existing behavior not introduced by
  this PR; tracked as a separate follow-up rather than expanding this PR's scope.
- **[tests] LOW — remaining coverage follow-ups: PARTIALLY DEFERRED.** The main
  gap (T1, dry-run write-filter count unguarded end-to-end) is now closed with a
  foreign-caller dry-run assertion. The remaining suggestions (feed DELETE /search
  route currently `:delete-search-tests? false`, a direct unit test for
  `find-write-restriction-query-part`) are noted as optional follow-ups; the
  write-vs-read divergence is already regression-tested at the route level
  (green/everyone) and the store level (`handle-delete-search-write-access-control-test`).

### Second-review adjudication (ereteog, 2026-09-03; addressed in 4f244c79)

- **[S1] MEDIUM — fail-closed default: APPLIED.** Dropped `:or {access-control
  :read}` in `make-search-query`; selection is now mandatory. The `case` has no
  default, so an omitted/typo'd key throws rather than silently selecting the
  read filter. Read callers pass `:access-control :read` explicitly; the comment
  is rewritten to describe this accurately.
- **[coupling] MEDIUM — direct write-clause unit test: APPLIED (no longer
  optional).** `find-write-restriction-query-part-test` pins the exact emitted
  `:should` vector and asserts no blanket public-TLP disjunct on the write path,
  so a read-side widening of the shared vector that leaked onto the delete path
  now breaks a unit test. This supersedes the "optional follow-up" note above.
- **[doc] LOW ×2 — APPLIED.** Write-filter docstring narrowed (blanket-visibility
  vs `authorized_users`/`authorized_groups` grants); `delete-search` docstring
  now says "a truthy `:really-delete?`".
- **[style] LOW — `normalize-ident` map/`:unset`: NOT APPLICABLE.** Current
  `normalize-ident` returns `{:login .. :groups ..}` with no `:unset` sentinel;
  both callers destructure the same two keys, so the map is the shared shape,
  not redundant.
- **[tests] LOW — feed DELETE /search route; `purge-incidents!` count: DEFERRED**
  (consistent with the first-review adjudication above).

### DIFF
--- a/src/ctia/stores/es/crud.clj
+++ b/src/ctia/stores/es/crud.clj
@@ -597,17 +597,23 @@ It returns the documents with full hits meta data including the real index in wh
   ([es-conn-state :- ESConnState
     search-query :- SearchQuery
     ident
-    {:keys [extra-filters]}]
+    {:keys [extra-filters access-control]
+     :or {access-control :read}}]
    (let [{:keys [services]} es-conn-state
          {{:keys [get-in-config]} :ConfigService} services
          {:keys [filter-map range full-text]} search-query
          range-query (when range
                        {:range range})
+         ;; Delete operations must use the write access-control filter so a
+         ;; caller can only remove documents it is allowed to write (XFV-120).
+         restriction-query-part (case access-control
+                                  :write (es.query/find-write-restriction-query-part ident)
+                                  (es.query/find-restriction-query-part ident get-in-config))
          filter-terms (-> (ensure-document-id-in-map filter-map)
                           q/prepare-terms)]
      {:bool
       {:filter
-       (cond-> [(es.query/find-restriction-query-part ident get-in-config)]
+       (cond-> [restriction-query-part]
          true (into extra-filters)
          (seq filter-map) (into filter-terms)
          (seq range)      (conj range-query)
@@ -646,24 +652,36 @@ It returns the documents with full hits meta data including the real index in wh
                                       get-in-config)))))))
 
 (s/defn handle-delete-search
-  "ES delete by query handler"
+  "ES delete by query handler.
+
+   Builds the `_delete_by_query` body with the write access-control filter so a
+   caller can only delete documents it is allowed to write (XFV-120), rather
+   than every document it can read.
+
+   When `es-params` does not carry `:really-delete? true` no deletion is
+   performed: the number of matching (deletable) documents is returned as a
+   dry-run preview, computed with the same write filter so the preview matches
+   what an actual delete would remove."
   [{:keys [conn index] :as es-conn-state} :- ESConnState
    search-query :- SearchQuery
    ident
    es-params]
-  (let [query (make-search-query es-conn-state search-query ident)
+  (let [query (make-search-query es-conn-state search-query ident {:access-control :write})
         opts (select-keys es-params [:wait_for_completion :refresh])
-        nb-deleted (ductile.doc/count-docs conn index query)]
-    (log/info (format "deleting %s documents in %s (%s)"
-                      nb-deleted
-                      index
-                      (pr-str query)))
-    (:deleted
-     (ductile.doc/delete-by-query conn
-                                  [index]
-                                  query
-                                  opts)
-     nb-deleted)))
+        nb-matched (ductile.doc/count-docs conn index query)]
+    (if (:really-delete? es-params)
+      (do
+        (log/info (format "deleting %s documents in %s (%s)"
+                          nb-matched
+                          index
+                          (pr-str query)))
+        (:deleted
+         (ductile.doc/delete-by-query conn
+                                      [index]
+                                      query
+                                      opts)
+         nb-matched))
+      nb-matched)))
 
 (s/defn handle-query-string-count :- (s/pred nat-int?)
   "ES count handler"
--- a/src/ctia/stores/es/crud.clj
+++ b/src/ctia/stores/es/crud.clj
@@ -593,22 +593,25 @@ It returns the documents with full hits meta data including the real index in wh
   ([es-conn-state :- ESConnState
     search-query :- SearchQuery
     ident]
-   (make-search-query es-conn-state search-query ident {}))
+   (make-search-query es-conn-state search-query ident {:access-control :read}))
   ([es-conn-state :- ESConnState
     search-query :- SearchQuery
     ident
-    {:keys [extra-filters access-control]
-     :or {access-control :read}}]
+    {:keys [extra-filters access-control]}]
    (let [{:keys [services]} es-conn-state
          {{:keys [get-in-config]} :ConfigService} services
          {:keys [filter-map range full-text]} search-query
          range-query (when range
                        {:range range})
-         ;; Delete operations must use the write access-control filter so a
-         ;; caller can only remove documents it is allowed to write (XFV-120).
-         ;; Fail closed: only the explicit :read/:write values are accepted.
-         ;; An unknown value (typo, wrong type, nil) throws rather than
-         ;; silently widening a delete to the broader read filter (XFV-120).
+         ;; Access-control filter selection is mandatory and fail-closed: the
+         ;; caller MUST pass an explicit :read or :write. There is no default --
+         ;; an omitted key leaves `access-control` nil and the `case` below
+         ;; throws, so a future destructive path that forgets :access-control
+         ;; fails loudly instead of silently widening a delete to the broader
+         ;; read filter. An unknown value (typo, wrong type) throws for the same
+         ;; reason. Delete operations pass :write so a caller can only remove
+         ;; documents it is allowed to write; read/count/aggregate pass :read
+         ;; (XFV-120).
          restriction-query-part (case access-control
                                   :write (es.query/find-write-restriction-query-part ident)
                                   :read  (es.query/find-restriction-query-part ident get-in-config))
@@ -766,7 +769,8 @@ It returns the documents with full hits meta data including the real index in wh
    {:keys [agg-type] :as agg-query} :- AggQuery
    ident]
   (let [agg-query (assoc agg-query :agg-key :metric)
-        query (make-search-query es-conn-state search-query ident {:extra-filters (aggregation-filters agg-query)})
+        query (make-search-query es-conn-state search-query ident {:access-control :read
+                                                                    :extra-filters (aggregation-filters agg-query)})
         agg (make-aggregation agg-query)
         es-res (ductile.doc/query conn
                                   index
--- a/src/ctia/stores/es/crud.clj
+++ b/src/ctia/stores/es/crud.clj
@@ -606,9 +606,12 @@ It returns the documents with full hits meta data including the real index in wh
                        {:range range})
          ;; Delete operations must use the write access-control filter so a
          ;; caller can only remove documents it is allowed to write (XFV-120).
+         ;; Fail closed: only the explicit :read/:write values are accepted.
+         ;; An unknown value (typo, wrong type, nil) throws rather than
+         ;; silently widening a delete to the broader read filter (XFV-120).
          restriction-query-part (case access-control
                                   :write (es.query/find-write-restriction-query-part ident)
-                                  (es.query/find-restriction-query-part ident get-in-config))
+                                  :read  (es.query/find-restriction-query-part ident get-in-config))
          filter-terms (-> (ensure-document-id-in-map filter-map)
                           q/prepare-terms)]
      {:bool
@@ -651,7 +654,7 @@ It returns the documents with full hits meta data including the real index in wh
                                       ident
                                       get-in-config)))))))
 
-(s/defn handle-delete-search
+(s/defn handle-delete-search :- (s/pred nat-int?)
   "ES delete by query handler.
 
    Builds the `_delete_by_query` body with the write access-control filter so a
--- a/src/ctia/stores/es/query.clj
+++ b/src/ctia/stores/es/query.clj
@@ -64,7 +64,11 @@
   "Access-control filter for write/delete operations. Mirrors the disjuncts of
    `ctia.domain.access-control/allow-write?`. Unlike `find-restriction-query-part`
    it never adds the `max-record-visibility=everyone` public-TLP clause, so a
-   caller cannot match another group's TLP white/green records. See XFV-120."
+   caller cannot reach another group's TLP white/green records by blanket
+   visibility alone. It can still match a foreign group's record when that
+   record explicitly grants the caller via `authorized_users`/`authorized_groups`
+   — that is intended, since `allow-write?` grants write on the same basis.
+   See XFV-120."
   [ident]
   (let [{:keys [login groups]} (normalize-ident ident)]
     {:bool
--- a/src/ctia/stores/es/query.clj
+++ b/src/ctia/stores/es/query.clj
@@ -7,35 +7,59 @@
    [schema-tools.core :as st]
    [schema.core :as s]))
 
-(defn find-restriction-query-part
-  [{:keys [login groups]} get-in-config]
+(defn- normalize-ident
   ;; TODO do we really want to discard case on that?
-  (let [login (str/lower-case login)
-        groups (map str/lower-case groups)]
+  [{:keys [login groups]}]
+  {:login (str/lower-case login)
+   :groups (map str/lower-case groups)})
+
+(defn- write-restriction-should-clauses
+  "The should-clauses shared by the read and write access-control filters.
+   These mirror the disjuncts of `ctia.domain.access-control/allow-write?`:
+   document owner, `authorized_users`, `authorized_groups`, same-group records
+   at TLP amber or below, and same-group owner records at TLP red."
+  [login groups]
+  [;; Document Owner
+   {:bool {:filter [{:term {"owner" login}}
+                    {:terms {"groups" groups}}]}}
+
+   ;; or if user is listed in authorized_users or authorized_groups field
+   {:term {"authorized_users" login}}
+   {:terms {"authorized_groups" groups}}
+
+   ;; CTIM records with TLP equal or below amber that are owned by org BAR
+   {:bool {:must [{:terms {"tlp" (conj ac/public-tlps "amber")}}
+                  {:terms {"groups" groups}}]}}
+
+   ;; CTIM records with TLP red that is owned by user FOO
+   {:bool {:must [{:term {"tlp" "red"}}
+                  {:term {"owner" login}}
+                  {:terms {"groups" groups}}]}}])
+
+(defn find-restriction-query-part
+  "Access-control filter for read operations. In addition to the write
+   disjuncts, when `ctia.access-control.max-record-visibility` is `everyone`
+   it also matches any TLP white/green document regardless of owner/groups."
+  [ident get-in-config]
+  (let [{:keys [login groups]} (normalize-ident ident)]
     {:bool
      {:minimum_should_match 1
       :should
-      (cond->>
-       [;; Document Owner
-        {:bool {:filter [{:term {"owner" login}}
-                         {:terms {"groups" groups}}]}}
-
-           ;; or if user is listed in authorized_users or authorized_groups field
-        {:term {"authorized_users" login}}
-        {:terms {"authorized_groups" groups}}
-
-           ;; CTIM records with TLP equal or below amber that are owned by org BAR
-        {:bool {:must [{:terms {"tlp" (conj ac/public-tlps "amber")}}
-                       {:terms {"groups" groups}}]}}
-
-           ;; CTIM records with TLP red that is owned by user FOO
-        {:bool {:must [{:term {"tlp" "red"}}
-                       {:term {"owner" login}}
-                       {:terms {"groups" groups}}]}}]
-
+      (cond->> (write-restriction-should-clauses login groups)
         ;; Any Green/White TLP if max-visibility is set to `everyone`
         (ac/max-record-visibility-everyone? get-in-config)
         (cons {:terms {"tlp" ac/public-tlps}}))}}))
+
+(defn find-write-restriction-query-part
+  "Access-control filter for write/delete operations. Mirrors the disjuncts of
+   `ctia.domain.access-control/allow-write?`. Unlike `find-restriction-query-part`
+   it never adds the `max-record-visibility=everyone` public-TLP clause, so a
+   caller cannot match another group's TLP white/green records. See XFV-120."
+  [ident]
+  (let [{:keys [login groups]} (normalize-ident ident)]
+    {:bool
+     {:minimum_should_match 1
+      :should (write-restriction-should-clauses login groups)}}))
 
 
 (s/defn make-date-range-query :- search-schemas/RangeQuery
--- a/src/ctia/stores/es/query.clj
+++ b/src/ctia/stores/es/query.clj
@@ -16,8 +16,15 @@
 (defn- write-restriction-should-clauses
   "The should-clauses shared by the read and write access-control filters.
    These mirror the disjuncts of `ctia.domain.access-control/allow-write?`:
-   document owner, `authorized_users`, `authorized_groups`, same-group records
-   at TLP amber or below, and same-group owner records at TLP red."
+   document owner (TLP-independent), `authorized_users`, `authorized_groups`,
+   and same-group records at TLP amber or below.
+
+   The trailing TLP-red owner clause is redundant: `allow-write?` has no red
+   branch, and its owner rule matches at any TLP, so the owner should-clause
+   above already covers same-group owner records at TLP red. It is kept
+   unchanged only so `find-restriction-query-part` (read) emits the exact same
+   query body as before these clauses were factored out — the read filter must
+   stay byte-identical (see XFV-120)."
   [login groups]
   [;; Document Owner
    {:bool {:filter [{:term {"owner" login}}
@@ -31,7 +38,10 @@
    {:bool {:must [{:terms {"tlp" (conj ac/public-tlps "amber")}}
                   {:terms {"groups" groups}}]}}
 
-   ;; CTIM records with TLP red that is owned by user FOO
+   ;; CTIM records with TLP red owned by user FOO. Redundant with the owner
+   ;; clause above (which is TLP-independent, matching owner+group at any TLP);
+   ;; retained only to keep the read filter byte-identical to its pre-refactor
+   ;; form. See the fn docstring and XFV-120.
    {:bool {:must [{:term {"tlp" "red"}}
                   {:term {"owner" login}}
                   {:terms {"groups" groups}}]}}])
--- a/src/ctia/entity/feed.clj
+++ b/src/ctia/entity/feed.clj
@@ -396,16 +396,16 @@
                  (if (empty? query)
                    (forbidden {:error "you must provide at least one of from, to, query or any field filter."})
                    (ok
-                    (if (:REALLY_DELETE_ALL_THESE_ENTITIES params)
-                      (-> (get-store :feed)
-                          (delete-search
-                           query
-                           identity-map
-                           {:wait_for_completion (boolean (:wait_for params))}))
-                      (-> (get-store :feed)
-                          (query-string-count
-                           query
-                           identity-map))))))))
+                    ;; Both the dry-run count and the destructive delete go
+                    ;; through delete-search so the reported count reflects the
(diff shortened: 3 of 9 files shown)


ereteog (commented):
**Verdict: COMMENTED** — 0 CRITICAL, 0 HIGH, 3 MEDIUM, 4 LOW

Write-access-control filter for ES `delete-by-query` so a caller can only delete what it may write, not merely read (XFV-120). Reviewed description-blind across six specialty passes (code, tests, silent-failure, types, comments, simplify) with every finding re-verified against the PR head. The core fix is correct and well-built: I traced `find-write-restriction-query-part` against `ctia.domain.access-control/allow-write?` disjunct-for-disjunct, confirmed the write filter is a strict subset of the read filter (so the change can only ever delete *fewer* documents), confirmed the read filter output is byte-identical to before, and confirmed the regression test fails if the write filter is reverted to the read filter. Nothing blocks. The three MEDIUMs are hardening/coverage, not defects: (1) the `:access-control` dispatch fails *open* to the read filter on an unknown value; (2) the store `delete-search` default silently became a no-op; (3) the dry-run write-filter count is unguarded end-to-end.

## MEDIUM
- **[S1]** — **Access-control dispatch fails open to the read filter**. `make-search-query`'s `case` sends any non-`:write` value (typo, wrong type, `nil`) to the broader read filter (`src/ctia/stores/es/crud.clj:609`), so a future delete-path caller that mistypes the option would silently delete against read visibility, re-opening XFV-120. No current caller triggers it — `handle-delete-search` passes the literal `:write` — so this is fail-open hardening, not a live bug (inline fix on the line).
- **[B3]** — **Store `delete-search` default silently became a no-op**. Deletion now happens only when the caller passes `:really-delete? true`; otherwise `handle-delete-search` returns `nb-matched` (`src/ctia/stores/es/crud.clj:672`), so a downstream `IQueryStringSearchableStore/delete-search` caller relying on the previous always-deletes contract gets a plausible non-zero count with nothing deleted. All in-repo callers are updated, so this is latent. Suggested fix: document the `:really-delete?` contract on the protocol method (`src/ctia/store.clj:35`).
- **[T1]** — **Dry-run write-filter count is unguarded end-to-end**. The route change's stated purpose is that the `REALLY_DELETE_ALL_THESE_ENTITIES=false` count now uses the write filter, but the green `search-access-control-test` only issues the destructive delete and `test-delete-search` uses a single owner where read==write (`test/ctia/test_helpers/access_control.clj:557`); reverting the dry-run route to `query-string-count` would pass every test. Suggested fix: add a foreign-caller dry-run DELETE asserting the body equals the write-filtered count.

<details>
<summary>LOW / non-blocking follow-ups (4)</summary>

- **[doc/O1]** — the TLP-red `should` clause is fully subsumed by the owner clause (owner+groups already matches at any TLP), and the docstrings list "same-group owner records at TLP red" as an `allow-write?` disjunct, which has no red branch — its owner rule is TLP-agnostic (`src/ctia/stores/es/query.clj:20`, `:35`).
- **[type]** — `handle-delete-search` now serves both the count and the delete and is `ok`'d by both routes but has no return schema, unlike its sibling `handle-query-string-count :- (s/pred nat-int?)` (`src/ctia/stores/es/crud.clj:654`).
- **[observability]** — pre-existing: the destructive branch reads only `:deleted`, dropping `_delete_by_query` `:failures`/`:version_conflicts`, and in async mode (route default `wait_for_completion=false`) returns the pre-count as the deleted count (`src/ctia/stores/es/crud.clj:678`); the touched expression is a good place to log partial failures and document the async semantics.
- **[tests]** — coverage follow-ups: no owner (non-zero) store-level dry-run assertion; the feed DELETE /search route is untested (`:delete-search-tests? false`); no direct unit test for `find-write-restriction-query-part` (its `authorized_users`/`authorized_groups` disjuncts are never exercised on the write path); confirm no other entity test relies on cross-identity purge now that the write filter is enforced.
</details>

## PASS
- **Security (core)** — `find-write-restriction-query-part` mirrors `ctia.domain.access-control/allow-write?` disjunct-for-disjunct (owner, `authorized_users`, `authorized_groups`, `tlp ∈ {white,green,amber}`+group; the red+owner clause is subsumed by owner), and is a strict subset of the read filter, so the change can only ever delete fewer documents. The read filter output is byte-identical to before the refactor.
- **Correctness** — the dry-run preview and the destructive delete run through one write-filtered query, so they cannot drift; every non-delete `make-search-query` caller (`handle-query-string-search`, `handle-query-string-count`, aggregation) keeps the `:read` default and is unaffected.
- **Tests** — the core regression is well-guarded: negative control (foreign group → 0, owner's docs survive) and positive control (owner → full count), and the read-first assertion (foreign reads all 60 under `everyone`) makes the zero-delete non-vacuous.
- **Cross-cutting** — `purge-incidents!` correctly migrated to a store-level owner-identity delete so the test suite keeps purging under the tighter filter.

## Intent pass
Code matches the stated scope: the write filter is a strict subset of the read filter (verified), only `delete-search` opts into `:write`, and read/count/aggregate paths are untouched — no scope creep, and the "can only ever delete fewer documents" claim holds. The one inherent consequence is that a dry-run now returns the write-*deletable* count, so a caller who can read but not delete sees `0` (see the LOW observability note); that is the intended security tradeoff, not a mismatch.

---
_Reviewed via `/review-contributor-pr` — verify mode: high; agents: code-reviewer, silent-failure-hunter, pr-test-analyzer, type-design-analyzer, comment-analyzer, code-simplifier. Layer-0 findings dropped: 0. Short-PR heuristic: off. Rubric: RUBRIC.md @ ccf75440 (no upstream criteria file in ctia — drift check skipped). Not approved: `--approve low` requires max severity ≤ LOW, but 3 MEDIUM findings survived, so the event resolved to COMMENT._

src/ctia/stores/es/crud.clj
```diff
@@ -597,17 +597,23 @@ It returns the documents with full hits meta data including the real index in wh
...
+    {:keys [extra-filters access-control]
+     :or {access-control :read}}]
    (let [{:keys [services]} es-conn-state
          {{:keys [get-in-config]} :ConfigService} services
          {:keys [filter-map range full-text]} search-query
          range-query (when range
                        {:range range})
+         ;; Delete operations must use the write access-control filter so a
+         ;; caller can only remove documents it is allowed to write (XFV-120).
+         restriction-query-part (case access-control
```
ereteog, line 615: **[S1]** The `case` selects the broader **read** filter for any `:access-control` value other than the literal `:write` (a typo, a string, or `nil`), so a future delete-path caller that mistypes the option silently deletes against read visibility, re-opening the XFV-120 cross-group delete this PR closes. No current caller triggers it — `handle-delete-search` passes `:write` — so this is fail-open hardening, not a live bug. Suggested fix: make the dispatch fail closed so an unknown value throws instead of widening:
```clojure
(case access-control
  :write (es.query/find-write-restriction-query-part ident)
  :read  (es.query/find-restriction-query-part ident get-in-config))
```

src/ctia/stores/es/query.clj
```diff
@@ -7,36 +7,60 @@
...
+   ;; or if user is listed in authorized_users or authorized_groups field
+   {:term {"authorized_users" login}}
+   {:terms {"authorized_groups" groups}}
+
+   ;; CTIM records with TLP equal or below amber that are owned by org BAR
+   {:bool {:must [{:terms {"tlp" (conj ac/public-tlps "amber")}}
+                  {:terms {"groups" groups}}]}}
+
+   ;; CTIM records with TLP red that is owned by user FOO
+   {:bool {:must [{:term {"tlp" "red"}}
```
ereteog, line 45: **[doc]** This TLP-red `should` clause is fully subsumed by the owner clause above — owner+groups already matches the document at any TLP, including red — and the shared docstring lists "same-group owner records at TLP red" as a disjunct of `allow-write?`, which has no red branch (its owner rule is TLP-agnostic). Harmless: the effective match set still equals `allow-write?`. But a maintainer checking parity will look for a red branch and not find one. Suggested fix: drop the red clause and describe the owner rule as TLP-independent in the docstring.

src/ctia/stores/es/crud.clj
```diff
@@ -646,24 +652,36 @@ It returns the documents with full hits meta data including the real index in wh
                                       get-in-config)))))))
 
 (s/defn handle-delete-search
```
ereteog: **[type]** `handle-delete-search` now serves both the dry-run count and the destructive delete and is `ok`'d by both routes, but unlike its sibling `handle-query-string-count :- (s/pred nat-int?)` just below it, it has no return schema. Suggested fix: annotate `:- (s/pred nat-int?)` to pin the "always a nat-int count" contract.

src/ctia/stores/es/crud.clj
yogsototh: Fixed in acf340ff. `make-search-query`'s `case` now has explicit `:read`/`:write` branches with no fallthrough default, so an unknown `:access-control` value throws `IllegalArgumentException` (fail closed) instead of widening to the read filter. All current callers still resolve to `:read` (default) or `:write` (delete-search only), so the change is behavior-preserving for existing paths.

src/ctia/stores/es/query.clj
yogsototh: Docstring fixed in acf340ff; the red clause is kept rather than dropped. Dropping it would change the read filter's emitted query body — the clause lives in the shared `write-restriction-should-clauses`, and master's `find-restriction-query-part` includes it, so removing it breaks the byte-identical-read-filter property this PR relies on (and which you verified). Instead the fn docstring and the inline comment now state that the owner rule is TLP-independent and that the red clause is redundant (subsumed by the owner clause, `owner+groups` matching at any TLP), retained only to keep the read filter byte-identical to its pre-refactor form. The effective write match set is unchanged either way (red ⊆ owner). Recorded in the PR body under Rejected findings.

src/ctia/stores/es/crud.clj
yogsototh: Fixed in acf340ff. `handle-delete-search` is now annotated `:- (s/pred nat-int?)`, matching its sibling `handle-query-string-count`. The contract holds on every path: the dry run returns the `count-docs` result, the synchronous delete returns `:deleted`, and async mode (no `:deleted` key) falls back to `nb-matched` — all nat-ints.

yogsototh:
Thanks for the thorough review. Adjudication of all findings below; all three inline threads are answered and resolved. Pushed as acf340ff (CI green).

### MEDIUM
- **[S1] Access-control dispatch fails open — FIXED (acf340ff).** `make-search-query`'s `case` now has explicit `:read`/`:write` branches with no fallthrough default; an unknown value throws `IllegalArgumentException` (fail closed). Existing callers still resolve to `:read`/`:write`, so behavior is unchanged.
- **[B3] Store `delete-search` default silently became a no-op — FIXED (acf340ff).** Documented the `:really-delete?` dry-run contract on the `IQueryStringSearchableStore/delete-search` protocol method (`src/ctia/store.clj`).
- **[T1] Dry-run write-filter count unguarded end-to-end — FIXED (acf340ff).** Added a foreign-caller dry-run DELETE assertion to `search-access-control-test`: under `everyone`, player-2 reads 3 green records but the dry-run preview count is 2 (write-filtered). Reverting the route's dry run to the read-filtered count now fails this assertion.

### LOW
- **[doc/O1] TLP-red clause subsumed — docstring FIXED, clause KEPT (acf340ff).** Dropping the shared red clause would change the read filter's emitted body and break byte-identity. Fixed the docstring/comment to describe the owner rule as TLP-independent and the red clause as redundant-but-retained-for-byte-identity instead. Recorded under Rejected findings in the PR body.
- **[type] `handle-delete-search` no return schema — FIXED (acf340ff).** Annotated `:- (s/pred nat-int?)`, matching `handle-query-string-count`.
- **[observability] dropped `:failures`/`:version_conflicts`, async pre-count — NOT APPLIED (out of scope).** Pre-existing, not introduced here; left as a separate follow-up.
- **[tests] remaining coverage follow-ups — PARTIALLY DEFERRED.** T1 closes the main gap; the feed DELETE /search route (`:delete-search-tests? false`) and a direct unit test for `find-write-restriction-query-part` are noted as optional follow-ups. The write-vs-read divergence is already regression-tested at the route (green/everyone) and store (`handle-delete-search-write-access-control-test`) levels.

Read access-control filter output remains byte-identical (only comments/docstrings changed in `query.clj`; the should-clause vector is untouched). The write filter is still a strict subset of the read filter.

src/ctia/stores/es/crud.clj
```diff
@@ -597,17 +597,26 @@ It returns the documents with full hits meta data including the real index in wh
   ([es-conn-state :- ESConnState
     search-query :- SearchQuery
     ident
-    {:keys [extra-filters]}]
+    {:keys [extra-filters access-control]
+     :or {access-control :read}}]
```
ereteog: **[S1]** Still open: the `case` now throws on an unknown *value*, but `:or {access-control :read}` still selects the broader **read** filter whenever the key is *absent* — including when a future caller mistypes the key itself (`:acess-control`), which is the realistic mistake rather than a bad value. The comment three lines down reads as a guarantee, so a maintainer adding a second destructive path omits the option and reintroduces the cross-group delete with no exception and a green suite.

Suggested fix: narrow the comment to what the code does, or drop the default and pass `:access-control` explicitly at all four call sites:

```clojure
{:keys [extra-filters access-control]}]
;; ... :read at handle-query-string-search / handle-query-string-count / aggregate,
;; :write at handle-delete-search — no default, so a missing key throws.
```

src/ctia/stores/es/query.clj
```diff
@@ -7,36 +7,70 @@
...
-  [{:keys [login groups]} get-in-config]
+(defn- normalize-ident
   ;; TODO do we really want to discard case on that?
-  (let [login (str/lower-case login)
-        groups (map str/lower-case groups)]
+  [{:keys [login groups]}]
+  {:login (str/lower-case login)
+   :groups (map str/lower-case groups)})
+
+(defn- write-restriction-should-clauses
```
ereteog, line 16: **[coupling]** This vector is now the sole source for both the read filter and the delete filter, and nothing in the code marks which disjuncts belong to which policy, so a future read-side widening lands silently on the delete path. `handle-delete-search-write-access-control-test` catches the specific public-TLP case, not an arbitrary new disjunct.

Suggested fix: pin the write filter's emitted clause set in a unit test, so any edit here has to be acknowledged on the write side:

```clojure
(deftest find-write-restriction-query-part-test
  (is (= {:bool {:minimum_should_match 1
                 :should [{:bool {:filter [{:term {"owner" "johndoe"}}
                                           {:terms {"groups" ["group1"]}}]}}
                          {:term {"authorized_users" "johndoe"}}
                          {:terms {"authorized_groups" ["group1"]}}
                          {:bool {:must [{:terms {"tlp" ["white" "green" "amber"]}}
                                         {:terms {"groups" ["group1"]}}]}}
                          {:bool {:must [{:term {"tlp" "red"}}
                                         {:term {"owner" "johndoe"}}
                                         {:terms {"groups" ["group1"]}}]}}]}}
         (sut/find-write-restriction-query-part {:login "JohnDoe" :groups ["Group1"]})))
  ;; and no public-TLP disjunct, whatever max-record-visibility says
  )
```

ereteog (commented):
**Verdict: COMMENTED** — 0 CRITICAL, 0 HIGH, 2 MEDIUM, 6 LOW

Second pass over the write access-control filter for ES `delete-by-query` (XFV-120), reviewed description-blind across six specialty passes with every finding re-verified against `acf340ff`. The three MEDIUMs from the first review are addressed and I confirmed each fix independently: the protocol contract is documented, and the route-level dry-run is now genuinely discriminating (reverting the dry run to a read-filtered count makes the green/`everyone` case assert `"2"` against an actual `"3"` at `test/ctia/test_helpers/access_control.clj:146`). I re-traced the security core against `ctia.domain.access-control/allow-write?` clause by clause and against the base ref: the write filter is a strict subset of the read filter, the read filter's emitted body is unchanged, and the dry-run count is literally the same query value the delete uses. Nothing blocks. The two MEDIUMs are the residue of the fail-closed fix and a new coupling introduced by the refactor.

## MEDIUM
- **[S1]** — **Still open: the "fail closed" comment does not describe the default**. The `case` now throws on an unknown *value*, but `:or {access-control :read}` still selects the broader **read** filter whenever the key is *absent* — including when a future caller mistypes the key itself (`:acess-control`), which is the realistic mistake (`src/ctia/stores/es/crud.clj:601`). A maintainer adding a second destructive path reads the comment at `:607` as a guarantee, omits the option, and reintroduces the cross-group delete with no exception and a green suite. No current caller is affected: `handle-delete-search` passes `:write`, and every read/count/aggregate caller wants the `:read` default. Suggested fix: keep the default but narrow the comment to what it does — unknown *values* throw, an omitted key means read — or drop `:or` and pass `:access-control` explicitly at all four call sites.
- **[coupling]** — **One clause vector now defines two security policies**. `write-restriction-should-clauses` (`src/ctia/stores/es/query.clj:16`) is the sole source for both the read filter and the delete filter, and nothing in the code distinguishes which disjuncts belong to which policy, so a future read-side widening lands silently on the delete path. The store-level regression test catches the specific public-TLP case, not an arbitrary new disjunct. Suggested fix: add a unit test pinning `find-write-restriction-query-part`'s emitted clause set, so any edit to the shared vector has to be acknowledged on the write side.

<details>
<summary>LOW / non-blocking follow-ups (6)</summary>

- **[doc]** — `find-write-restriction-query-part`'s "a caller cannot match another group's TLP white/green records" is absolute, but the shared `authorized_users` / `authorized_groups` disjuncts do let a foreign group delete a green record it was granted (`src/ctia/stores/es/query.clj:66`); a maintainer debugging that will read correct delegation as a regression.
- **[doc]** — the protocol docstring says deletion happens when `params` carries `:really-delete? true`, while the code tests truthiness (`src/ctia/store.clj:40` vs `src/ctia/stores/es/crud.clj:675`); harmless today because both routes coerce with `boolean`, but a future caller forwarding a raw param would delete on `"false"`.
- **[tests]** — the feed `DELETE /search` route changed to the always-`delete-search` shape with no test at all (`src/ctia/entity/feed.clj:401`; `test/ctia/entity/feed_test.clj` keeps `:delete-search-tests? false`), so dropping `:really-delete?` there is undetectable.
- **[tests]** — `purge-incidents!` discards the delete count (`test/ctia/entity/incident_test.clj:912`), so a future test that creates incidents under a non-`foouser` identity fails as an opaque `(assert (< tries 10))` about 55s later rather than "purge deleted 0".
- **[style]** — `normalize-ident` builds a map both callers immediately destructure back into two positional args (`src/ctia/stores/es/query.clj:10`), and the `:unset` sentinel is equivalent to `or` at every real call site (`test/ctia/test_helpers/access_control.clj:71`).
- **[observability]** — noted as accepted follow-up, not re-raised: the rewritten `(:deleted … nb-matched)` still drops `:failures` / `:version_conflicts` and reports the pre-count as deleted in async mode (`src/ctia/stores/es/crud.clj:681`).
</details>

## PASS
- **Security (core)** — re-traced against the PR base: `write-restriction-should-clauses` mirrors `allow-write?` disjunct for disjunct (owner is TLP-independent in both; the single `tlp ∈ {white,green,amber}` + groups clause covers `allow-write?`'s two separate branches; the red clause is a strict conjunctive refinement of the owner clause, so it is inert), and the write filter is a strict subset of the read filter — the change can only ever delete fewer documents.
- **Read path unchanged** — `find-restriction-query-part` emits the same clauses in the same order with the same `cons` under `everyone` as the base ref; the refactor is behaviour-preserving on the read side.
- **Correctness** — `query` is bound once and passed to both `count-docs` and `delete-by-query`, so the preview cannot drift from the delete. Every non-delete caller keeps the `:read` default. The single protocol implementation (`src/ctia/stores/es/store.clj:100`) threads `params` verbatim, and all four in-repo call sites pass the flag — no store silently stopped deleting.
- **Tests** — the store-level regression is discriminating (foreign group reads 60, deletes 0, owner still deletes 60) and pins the `max-record-visibility=everyone` precondition explicitly; the route-level dry-run assertion closes the first review's T1 gap, and dry-run purity is guarded by the destructive delete that follows returning the full deletable count.

## Intent pass
Code matches the stated scope, including the claims the description makes about its own invariants: I verified the strict-subset property, the unchanged read filter, and the shared query value behind the dry-run count rather than taking them from the summary. No performance claim is made, so no benchmark is owed. The one inherent consequence stands as intended: a caller who can read but not delete now sees `0` from a dry run, which is the security tradeoff, not a mismatch.

---
_Reviewed via `/review-contributor-pr` — verify mode: high; agents: code-reviewer, silent-failure-hunter, pr-test-analyzer, type-design-analyzer, comment-analyzer, code-simplifier. Findings dropped in verify: 6 (including one agent claim of a large test deletion that came from a stale local base, not this PR). Layer-0 findings dropped: 0. Short-PR heuristic: off. Rubric: RUBRIC.md @ ccf75440 (no upstream criteria file in ctia — drift check skipped). Not approved: `--approve low` requires max severity ≤ LOW, and 2 MEDIUM findings survived, so the event resolved to COMMENT._

src/ctia/stores/es/crud.clj
yogsototh: Addressed in 4f244c79. I took the stronger of your two options: dropped `:or {access-control :read}` so filter selection is now mandatory and fail-closed. The `case` has no default, so an omitted key leaves `access-control` nil and throws (`No matching clause: nil`) rather than silently selecting the read filter. The read call sites now pass `:access-control :read` explicitly (3-arg arity and `handle-aggregate`); `handle-delete-search` still passes `:write`. This means the exact realistic mistake you describe -- a maintainer adding a destructive 4-arg call and omitting the option -- now fails loudly at runtime instead of green. The comment at ~:607 is rewritten to describe this accurately: mandatory selection, no default, unknown/omitted values throw.

src/ctia/stores/es/query.clj
yogsototh: Addressed in 4f244c79. Added `find-write-restriction-query-part-test` in `test/ctia/stores/es/query_test.clj`, a pure unit test that pins the exact `:should` clause vector emitted by `find-write-restriction-query-part`, plus a second assertion that the write filter contains no blanket `{:terms {"tlp" public-tlps}}` disjunct. Because `write-restriction-should-clauses` is the shared source for both policies, any read-side widening of that vector that landed on the delete path now breaks this test on the write side (the exact-`=` assertion fails on any added/reordered disjunct). Test passes locally (2 assertions, 0 failures).

yogsototh:
## Second-review follow-ups addressed (4f244c79)

Thanks for the second pass. Dispositions for the 2 MEDIUM + 6 LOW:

**MEDIUM**
- **[S1] fail-closed default** -- Fixed in `4f244c79`. Dropped `:or {access-control :read}`; selection is now mandatory. The `case` has no default, so an omitted/typo'd key throws instead of silently selecting the read filter. Read callers (3-arg arity, `handle-aggregate`) pass `:access-control :read` explicitly; `handle-delete-search` passes `:write`. Comment rewritten to match. The realistic footgun (new destructive path omits the option) now fails loudly.
- **[coupling] shared clause vector** -- Fixed in `4f244c79`. Added `find-write-restriction-query-part-test` pinning the exact emitted `:should` vector and asserting no blanket public-TLP disjunct on the write path. Any read-side widening of the shared vector that lands on the delete path now breaks this unit test.

**LOW**
- **[doc] write-filter docstring absolute claim** -- Fixed in `4f244c79`. Narrowed to "by blanket visibility alone" and now notes `authorized_users`/`authorized_groups` grants can match a foreign group's record (intended, matches `allow-write?`).
- **[doc] `:really-delete? true` vs truthiness** -- Fixed in `4f244c79`. Docstring now says "a truthy `:really-delete?`".
- **[tests] feed DELETE /search route** -- Deferred (as noted in the Rejected-findings ledger for the first review). `:delete-search-tests?` remains false; the store-level regression test covers the write-ACL behavior.
- **[tests] `purge-incidents!` discards delete count** -- Deferred. Test-only ergonomics; no behavior change.
- **[style] `normalize-ident` map/`:unset`** -- Not applicable to current code: `normalize-ident` returns `{:login .. :groups ..}` with no `:unset` sentinel; both callers destructure the same two keys, so the map is the shared shape, not redundant.
- **[observability] `_delete_by_query` :failures/:version_conflicts + async pre-count** -- Deferred (accepted follow-up, pre-existing, out of scope for XFV-120).

Local targeted unit test green (2 assertions, 0 failures); the ES-integration `make-search-query-test` is unchanged in behavior (all callers pass explicit `:read`/`:write`) and is covered by CI.

yogsototh:
Re-requesting review. Since the last round (`acf340ffce`), `4f244c793f` addresses the two remaining findings:

- **[S1] fail-open dispatch** — `make-search-query` now selects the filter with a mandatory, fail-closed rule rather than a `case` that fell through to the broader read filter for any non-`:write` value.
- **[T1] dry-run write-filter count** — added a test that pins the `REALLY_DELETE_ALL_THESE_ENTITIES=false` count to the write filter (so reverting the dry-run route to the read-visibility count fails).

All inline threads are resolved and CI is green.

